PR #15, opened 2017-06-22
Update installing.md

daftaupe:
Removed the version in the repo (unmaintainable in the long term, the page will always be obsolete). 
Changed the order of the repos, most up-to-date first.


digitalcraftsman:
Hello @daftaupe,

thank you for submitting this fix. Removing the version was the best from a long term perspective

Cheers,
Digitalcraftsman

PR #16, opened 2017-06-22
Add Docker Hugo to installation page.

felicianotech:


digitalcraftsman:
Hi @felicianotech,

thanks for providing a third-party docker image :+1:

Issue #1, opened 2017-06-14
Add contribution guidelines


digitalcraftsman:
It's basically of the contribution guidelines in the main repo. Should also explain the purpose of the three branches.

bep:
Also, even if the commit message prefixing does not make much sense, we should keep some kind of standard here as well. A sloppy commit log looks like a sloppy project.

rdwatters:
I agree. I can propose a handful of new doc-specific prefixes, @bep @digitalcraftsman

bep:
I was making a point of getting rid of the prefix, so no new artificial prefixes. It does not make sense in a small doc project as this is. It made sense when it was part of the Hugo codebase.

rdwatters:
Okay, then perhaps I can help with editorial guidelines or some sort of commit standards if it helps.

rdwatters:
And please note...

https://hugodocs.info/contribute/documentation/

digitalcraftsman:
> Also, even if the commit message prefixing with "docs:"does not make much sense with one project, so we should get rid of that here

That makes sense - but old habbits. Release notes always had a 'docs' section. Now, the commits would have to be retrieved from the Git log of the `docs` submodule. Do you plan to keep this section in the docs

PR #19, opened 2017-06-22
docs: Add horeaporutiu.github.io to the showcase

horeaporutiu:


horeaporutiu:
Add horeaporutiu.github.io to the showcase.

digitalcraftsman:
Hello @horeaporutiu,

thank you for adding your website to the showcase of Hugo. I've merged your contribution as 309adec5.
Hugo's showcase section should be updated within the next minutes.

Cheers,
Digitalcraftsman

horeaporutiu:
Hello @digitalcraftsman,

I have checked the https://gohugo.io/showcase/ site this morning and I have yet to see my contribution. I was wondering if there is some problem with my submission (I have heard my .gitignore file should have my name and email address associated with Github)?

Best,
Horea Porutiu

digitalcraftsman:
Hello @horeaporutiu,

your commit was fine. So no problem on your end. When pushing a new commit to this repo a rebuild of the website should be triggered. However, I merged you commit into master. But the rebuild is only triggered for new commits on the stable branch.

@bep according to my proposal in #1 the stable branch is intended for the commits until the latest release. I.e. the stable branch will be referenced in the docs submodule when releasing a new version of Hugo.

The master branch on other hand can be updated between two releases and should be considered as the source for www.gohugo.io. I.e. it gets rebuild with a new commit on this branch.

I just want to clarify why we use `stable` as the source for www.gohugo.io?

bep:
>I just want to clarify why we use stable as the source for www.gohugo.io?

Because `gohugo.io` represents a tagged version with potential additions that is correct for that tag. So when I released `v0.24.1` I reset the stable branch to that tag. So for any commits into master that is also relevant/correct for `v0.24.1` should be merged into `stable`.

It would make very little sense to have a docs site marked with `v0.24.1` and dump lots of 0.25 changes into it. Which is probably not what you described in your proposal, but `master` is by definition not stable, so I suggest we use the `dev` branch for bigger changes only (think: @rdwatters )

PR #57, opened 2017-07-18
Correct movie title

GamingCoder:
As a Bruce Willis fan I just had to correct this


digitalcraftsman:
Good catch @GamingCoder :+1: Thanks for submitting this fix.

PR #59, opened 2017-07-19
Revise homepage feature dynamic content claim

rdwatters:
See the following from a Hugo user for more background: https://discourse.gohugo.io/t/dynamic-api-driven-content/7499/8

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -10,7 +10,7 @@ features:
   - heading: Robust Content Management
     image_path: /images/icon-content-management.svg
     tagline: Flexibility rules. Hugo is a content strategist's dream.
-    copy: Hugo supports unlimited content types, taxonomies, menus, dynamic API-driven content, and more, all without plugins.
+    copy: Hugo supports unlimited content types, taxonomies, menus, flexible data-driven content, and more, all without plugins.
 
   - heading: Shortcodes
     image_path: /images/icon-shortcodes.svg


budparr:
Personally, I like the way it was and I'm not so sure this change doesn't present more questions than it answers. That said, if the original statement must go (that seemed to be @bep's direction), then perhaps just delete it (repeal and replace later <--bad joke).

bep:
@budparr please don't push your PR branches to gohugoio; that will get messy very fast.

bep:
Reopen on fork.

budparr:
Okay @bep

budparr:
@bep as this isn't my PR, I assume you're talking about my PR on the themes site: https://github.com/gohugoio/gohugoioTheme/pull/53 which I'm happy to redo as a fork.

bep:
I'm talking about branches in this repo, which I get when I do a `git pull`... So, in general, we do branching from forks via PRs. So we have a limited set of branches in this repo. I deviated from this rule with my big "next" branch, but that was special. By pushing feature branches to this repo you create maintenance work, and you also push branches on people who doesn't really care about some "marketing copy" thing.

budparr:
Thank you. I understand.

rdwatters:
> (repeal and replace later <--bad joke)

#dadjokes (which I love, for the record) 😆 

But seriously, I'm okay with either option. Maybe adding a form of the word "flexibility" was a bit redundant too.

bep:
Better than "capability", which I remember seeing more than once. :-)

bep:
That said, I'm from Norway, but Isn't the "ity" form considered a little heavy, chancellery language. As in: "Flexible" would be better in most cases.

rdwatters:
> That said, I'm from Norway

I've never been able to tell in any of our conversations. Your English is _fantastic_. 

Not sure what you mean by "chancellery" in this context&mdash;perhaps "uppity" or "highfalutin"&mdash;but the redundancy was between "Flexibility" rules and "flexible, data-driven content." 

If you ever find you're having difficulty falling asleep @bep, feel free to read the following: 

"Flexibility rules" is a complete sentence; i.e., there is a subject (the noun "flexibility") and a predicate (the present tense conjugation of the infinitive "to rule"). 

"Flexible rules" is an adjectival phrase and incomplete sentence (dependent clause); i.e., there is a pre-positive modifier (the adjective "flexible") of a plural noun ("rules"). 

Both are equally "correct," but for this case, the first makes more sense since it includes terminal punctuation; the "period" or "full stop." 

All said, marketing copy is very forgiving when it comes to prescriptive usage, and English is quirky as sh%#.

jhabdas:
When I originally set out to create the site which I built After Dark for I'd intended to use hugo do dynamically drive the site based on a FaaS (e.g. Lambda). This is not a stretch to me as hugo is fast enough it could, for all intents and purposes, be considered dynamic IMHO.

http://bezdelev.com/post/hugo-aws-lambda-static-website/

bep:
@jhabdas the problem isn't "how true it is", it is how the common man reads it ... wrong, and then we get questions on the forum, which want to avoid.

PR #62, opened 2017-07-19
removed redundant text

gaetawoo:


rdwatters:
Thanks @gaetawoo! Nice catch :smile:

PR #70, opened 2017-07-21
Reordered to match the following content order

gaetawoo:


rdwatters:
@gaetawoo Thanks for catching this inconsistency. It helped me realize that this is the only page that's really organized in this way. Do you think the entire unordered list is unnecessary in that it just reflects the toc? I can't remember my reasoning for the internal links for this page specifically...

bep:
@gaetawoo I know I probably enabled the CLA check after you submitted this PR (we had in the hugo repo, but then we split them and forgot ...). You only have to sign (aka push the button) once for each Hugo repository. It is just a formality, and makes little sense for small adjustment edits like these, but we might as well be pedantic about it.

gaetawoo:
@bep I'm sorry but I don't understand what you mean with CLA, or signing once.. As far as I know, I forked the Hugo Docs repo and have made changes that way.. Is that not the case? I've been doing this through the quick and convenient website GUI, not the command line, so I haven't been pulling down the current source before making the change and pushing. Is there a best way you would prefer for how changes that are small like this be made? Can it be done through the web gui?

@rdwatters It's brief and it introduces the upcoming topics. I don't think it's bad. Just that it should match the order. I would keep it in there. It's nice for readers to know what to expect before scrolling through the wall of text they might be unfamiliar with.

digitalcraftsman:
@gaetawoo usually the CLA bot will post a comment with a link to the CLA. The CLA was added as a requirement after your opened this pull request. It's likely that the bot doesn't touch such commits.

Nonetheless, in the "Some checks haven’t completed yet" box at the very bottom click on the black **logo** of CLA bot. Sign in with your GitHub account, click on accept and you're done :wink:

bep:
Let's delay that CLA until your next PR; there is not much point being pedantic over spelling corrections. Thanks for your contribution!

gaetawoo:
@bep oh i see, well I signed the CLA anyway just now. So, in the future, I can make editorial changes (spelling, grammar, etc.) just by doing it through the gui the same way I have been? that's no problem?

PR #72, opened 2017-07-21
Complete "content" spelling under theme components

chesedo:


rdwatters:
Thanks @chesedo Good catch :smile:

PR #79, opened 2017-07-23
Remove inline icons from installation fixes #78

### LINKED ISSUE #78: Remove unused "<i class="icon-linux"></i>" and similar
It is not used.

And for the future: Do not add stuff that we may or may not use some time. It adds maintainance (this issue is an example of that).

rdwatters:
Per request from @bep in #78, this removes the inline icons for the various operating systems that were a holdover from the previous docs content.


bep:
Could you also fix the commit message + rebase against master so the build goes green.

PR #81, opened 2017-07-23
Remove inline icons from installation guide Fixes #78

### LINKED ISSUE #78: Remove unused "<i class="icon-linux"></i>" and similar

rdwatters:
Remove inline icons from `/getting-started/installing` per request from @bep in #78. These icons were carried over from content from the previous site.

### DIFF
--- a/content/getting-started/installing.md
+++ b/content/getting-started/installing.md
@@ -28,10 +28,10 @@ Hugo is written in [Go](https://golang.org/) with support for multiple platforms
 
 Hugo currently provides pre-built binaries for the following:
 
-* <i class="icon-apple"></i> macOS (Darwin) for x64, i386, and ARM architectures
-* <i class="icon-windows"></i> Windows
-* <i class="icon-linux"></i> Linux
-* <i class="icon-freebsd"></i> FreeBSD
+* macOS (Darwin) for x64, i386, and ARM architectures
+* Windows
+* Linux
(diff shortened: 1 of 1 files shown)


bep:
I fixed your commit message to be in line with: https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines

rdwatters:
Just so I am clear, the mistake I made in my commit was that I put `Fixes #78` in the commit summary and not the commit description?

bep:
Yes, or as I called it in the contribution guideline, the commit message title/subject.

rdwatters:
I see. Thank you. I've always seen it referred to as `git summary` and `git description`, so it wasn't entirely clear to me when we used "git commit message body" in the guidelines.

I won't make the same mistake next time :smile:

bep:
>git summary and git description

I have never seen those terms used in the wild, but that does not matter ...

rdwatters:
I think they're just baked into my head from when I first tried git a couple years ago and started with [the desktop app (which I don't use anymore)](https://stackoverflow.com/questions/36369602/what-is-the-difference-between-summary-and-description-in-github-desktop), but you're right that it doesn't matter :smile: Thanks!

PR #83, opened 2017-07-24
Load CSS and JS via HTTP/2 server push

bep:
See https://www.netlify.com/blog/2017/07/18/http/2-server-push-on-netlify/

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -25,7 +25,7 @@ pygmentsCodeFences = true
 pygmentsStyle = "friendly"
 
 [outputs]
-home = [ "HTML", "RSS", "REDIR" ]
+home = [ "HTML", "RSS", "REDIR", "HEADERS" ]
 section = [ "HTML", "RSS"]
 
 [mediaTypes]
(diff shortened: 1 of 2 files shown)


bep:
Note that this is more gimmicky/educational than useful.

layouts/index.headers
```diff
@@ -0,0 +1,11 @@
+ /*
+  X-Frame-Options: DENY
+  X-XSS-Protection: 1; mode=block
+  X-Content-Type-Options: nosniff
+  Referrer-Policy: origin-when-cross-origin
+
+
+# HTTP/2 server push config, Netlify Link headers https://www.netlify.com/blog/2017/07/18/http/2-server-push-on-netlify/
+/dist/*
```
moorereason: You're requesting that H2 Push only happen when a page is requested from the `/dist/` path.  I don't think that's what you're going for.

If you move these under the `/*` section, will Netlify send `Link` headers with non-HTML content?  We don't want to push extra content if the user is requesting an image or XML file, for example.

We may just want to scope it to the homepage (`/`) for now.

bep: You are commenting on a pretty old commit.

PR #105, opened 2017-07-27
WIP: Add namespaces for functions

digitalcraftsman:
This pull request is still a work in progress. But feel free to comment.

**What has been done so far?**

- all function files have been moved into their own subsection / namespace
- all functions have been removed from the side menu
- all subsections / namespaces have an `_index.md` that shows all functions in a namespace. They appear as links in the side menu instead

**What needs to be done?**

- [ ] the prev/next buttons are broken due to the reorganization (cc @budparr)
- [ ] handle template functions that are in none of the namespaces (see https://github.com/gohugoio/hugo/issues/3743)
- [x] list all namespaces  / subsections at `www.gohugo.io/functions/`
- [ ] put each function in it's own file (see base64.md etc)
- [x] make function names, link titles and file names  consistent re. capitalization
- [ ] set up aliases for redirects to prevent broken links

**Questions:**

@budparr it's not quite clear for me how I can list all subsections instead of the functions itself at `www.gohugo.io/functions/`

@bep how is it possible to output the current subsection in a template? For the file content/functions/math/add.md I would like to get the subsection math.

You can see the current status via Netlify's deploy preview below.


digitalcraftsman:
I wasn't happy with the link titles either and waited for alternatives :wink: I'll change them accordingly.

digitalcraftsman:
Is there any particular order you prefer for the namespaces in the menu?

moorereason:
Alphabetic order, please.

moorereason:
> use namespaces in all code examples in the docs (perhaps out of the scope of this issue)

Yes, this is out of scope, and I'm not sure I agree.  Most people will be using the aliases anyway.  The search feature should make it easy to find the docs for a given alias if you don't know which namespace it lives in.

digitalcraftsman:
> Alphabetic order, please.

Hugo orders menu items alphabetically per default when no weight is set. So there's nothing to do here.

>  Yes, this is out of scope, and I'm not sure I agree. Most people will be using the aliases anyway. The search feature should make it easy to find the docs for a given alias if you don't know which namespace it lives in.

Then I'll remove this bullet point.

bep:
This is me being not totally sure, but I don't like the "collections namespace" title on the section listing either. We call it a namespace in the code, but that is just a ... term that the end users don't relate too much with. I think I would prefer just to have one title: "Collections". And then, if we need to lookup in a JSON map, we could do a `.Title | lower` or something.

digitalcraftsman:
Shall I remove "namespace" from all page titles as well so that "collections namespace" becomes "Collections"?

bep:
I suggest you remove the linktitle, i.e. moving the linkttitle value into title and remove it.

bep:
As to the `.GetPage` methods, I suggest we go the most precise and correct way and create a new headline/top level section called "Methods", and add the relevant structs, i.e. `Page`, `Menu` etc.

moorereason:
I don't think most non-programmers know the difference between methods and functions.  I'd suggest choosing a different name, such as "Objects."

bep:
>I don't think most non-programmers know the difference between methods and functions. 

That doesn't matter, then they can learn something. `Method` is correct -- object doesn't mean anything.

bep:
It should maybe be consolidated with Variables, which is strongly related.

digitalcraftsman:
> As to the .GetPage methods, I suggest we go the most precise and correct way and create a new headline/top level section called "Methods", and add the relevant structs, i.e. Page, Menu etc.

With top level you mean on the same level as "Variables" and "Functions", right?

digitalcraftsman:
Regarding consistency: all template functions like `namespace.funcname` or `namespace.funcName`, where the function name **always** starts lowercase. `lang.NumFmt` is the only exception. Shall we create an alias called `numFmt`?

rdwatters:
> That doesn't matter, then they can learn something.

This feels unnecessarily antagonistic towards end users, @bep. I don't have quantitative _data_ to back this up, but I can say empirically that people don't come to the Hugo docs to learn object-oriented programming terminology. Further, I don't see how structuring web pages this way teaches them the difference between these terms; i.e. if our goal is to teach them.  

"Functions namespaces" makes more assumptions about end users than does "Quick Reference" and removes the convenience of the latter by creating more drill down. I'm not saying that more assumptions is a bad thing, but I can tell you (again, empirically) that conversations like "This won't make sense to our target end user" will go on ad infinitum if we don't define said user.

bep:
>This feels unnecessarily antagonistic towards end users, @bep. I don't have quantitative data to back this up, but I can say empirically that people don't come to the Hugo docs to learn object-oriented programming terminology. 

I don't know what "antagonistic" means, which is ironic. I don't see how "object" fits into "object oriented terminology", but that does not matter. Maybe we don't have a good term for this and can fall back on "object". But it then belongs in the next section.

So I suggest this:

```
Functions
  ...
Objects (old value: Variables)
   "Lists objects available to the templates with its methods and fields (variables)."
```

And then rearrange the content files below Varables to also hold ".GetPage" etc.

rdwatters:
I am not saying that I think we should use "object." I don't really have an opinion of "object" be "method." 

I'm talking about teaching users the differences between "function" and "method" and the approach of saying that the responsibility is on them (the users) to learn the difference rather for us to teach them or, more specifically, that it's up to them to know the difference coming in. 

The restructuring of this PR gets rid of the Quick Reference and now requires more drill down. When I said it makes "more assumptions," I mean the following: 

1. The previous single QR list was alphabetical and newbs with only front-end dev can figure out what titles like "safeCSS" are related to. 
2. We now assume the user knows to look in these namespaces (and that they know what is meant by "namespaces") as a means to aid in navigation. So now, a user needs to know what we mean by "collections" and "inflect." 

Again, I'm not saying it's bad to organize it this way. It has nothing to do with correctness. With respect to #2, if we think this fits the sort of baseline knowledge of our target users, then that's great and this is a smart improvement. I also have no problem with automating documentation whenever we can, and if you think this is a step in that direction, I'm all for it. 

Also keep in mind that we can accomplish the same thing by treating namespace as a taxonomy or just FM field and sorting accordingly in the quick reference list: 

1. It creates less drilldown
2. It treats `namespace` as an attribute of an archetype, which it is. "Section" is an attribute of site architecture. This is a conflation of content versus context. 
3. The attribute would then travel with the instance of the content/archetype. 
4. This allows the functions to surface in more than one list. You *could* get fancy with a client-side filter too if you want, but I'm not really talking about design suggestions :smile:

bep:
Let us close the discussion with this:

```
Functions
  ...
Objects (renamed Variables)
   "Lists objects available to the templates with its methods and fields (variables)."
```

So, all misplaced methods in Functions must be moved to its logical place in `Objects` (renamed from `Variables`).

digitalcraftsman:
Did you, @bep, @moorereason  and @rdwatters, found a consensus about how we should name and structure things?

rdwatters:
Here are my final thoughts: 

1. In my experience, conflating context and content never works in the long run, but that said...
2. @bep is the leader of this project and therefore has the D on content, editorial, and UX decisions; I'm okay with deferring to whatever he wants and will support his decision unequivocally :smile:

moorereason:
I came across some Microsoft docs that I thought were clean and well organized.  It may help us think through how we could make these namespace docs look.

[Convert-MsolDomainToFederated](https://docs.microsoft.com/en-us/powershell/module/msonline/convert-msoldomaintofederated?view=azureadps-1.0) is a Powershell cmdlet in the [MSOnline](https://docs.microsoft.com/en-us/powershell/module/msonline/?view=azureadps-1.0) module.  The module is analogous to our namespace.  I like the "module" terminology and wouldn't be opposed to use using that term instead of "namespace."  Something to think about.

bep:
Agree that `module` may be easier to understand for the "common man".

rdwatters:
@moorereason These docs looks really nice. I like the menu as well in terms of nested dropdowns:

http://www.webhook.com/docs/template-rules-and-filters/

Note that the "subsections" are really just attributes/metadata written to the page to group semantically related entries and don't require the extra drilldown in terms of URLs...

BTW, these were designed by Dave Snider, the guy responsible for the ever-popular [Sphinx RTD Theme](https://github.com/rtfd/sphinx_rtd_theme/graphs/contributors). 

@budparr How difficult would it be to implement nested dropdowns in the main navigation like the MS docs or the simpler WH doc?

bep:
BUT: I don't see how module vs namespaces or nested menus helps @digitalcraftsman an inch closer to an answer to his question.

So, I will close the discussion by repeating myself:

```
Let us close the discussion with this:

Functions
  ...
Objects (renamed from Variables)
   "Lists objects available to the templates with its methods and fields (variables)."
So, all misplaced methods in Functions must be moved to its logical place in Objects (renamed from Variables).

```

rdwatters:
[I already answered his question above](https://github.com/gohugoio/hugoDocs/pull/105#issuecomment-319519967). LMK if I can help @digitalcraftsman.

bep:
@digitalcraftsman I assume this one is stale?

We need a fix for this sooner or later. We keep adding template functions -- and the docs gets messier by the day.

/cc @moorereason

moorereason:
If @digitalcraftsman doesn't have time to bring this PR up-to-date, I should have time in December to start a separate PR.

bep:
Maybe we should wait for "pages from data" -- this would be a good test case.

/cc @regisphilibert

regisphilibert:
> Maybe we should wait for "pages from data" -- this would be a good test case.

Not sure to follow, what would be the remote data source here ?

bep:
>Not sure to follow, what would be the remote data source here ?

Why does it have to be remote? 

https://github.com/gohugoio/hugoDocs/blob/master/data/docs.json

regisphilibert:
Oh yeah of course! Got it.

PR #116, opened 2017-07-31
Replace old RSS template with new embedded version

gaetawoo:
Current documentation shows an old version of the embedded RSS template. The newer version is in:
https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/template_embedded.go

This was copied and pasted into the documentation as embedded.


digitalcraftsman:
Hello Jeremiah, 

thanks for updating the documentation :+1:

Issue #117, opened 2017-08-01
'File an issue' link broken


Jos512:
Hi,

The 'File an issue' link that appears in the footer of every page is broken. The HTML it generates is:

```
      <div class="w-third pv3 dn db-l">
        <ul class="list ma0 pa0 w-50 center">
          <li><a href="" class="link white dim pv2 db">File an Issue</a></li>
          <li><a href="https://discourse.gohugo.io/" class="link white dim pv2 db">Get Help</a></li>
          <li><a href="https://gitter.im/spf13/hugo" class="link white dim pv2 db">Discuss the Source Code</a></li>
        </ul>
      </div>
```

Here's how that link in question looks like:

![2017-08-01_14-54-56](https://user-images.githubusercontent.com/13403160/28826202-7cfce3a6-76c9-11e7-8c6f-fead510bd0c3.png)

PR #120, opened 2017-08-01
Fix footer issue link

rdwatters:
See #117. This adds an additional param in config to meet the footer partial in @budparr 's theme here:

https://github.com/gohugoio/gohugoioTheme/blob/master/layouts/partials/site-footer.html#L22

Note that the value is for the GH org page and not a direct link to an `issues/new`, which might be confusing for users since we treat issues for docs separately from issues for projects. 

We might want to consider two links if we really want to make is easy, but that can wait for another day.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -53,6 +53,8 @@ twitter = "GoHugoIO"
   removefromexternalsearch = false
   ## Gh repo for site footer (include trailing slash)
   ghrepo = "https://github.com/gohugoio/hugoDocs/"
+  ## GH Repo for filing a new issue
+  github_repo = "https://github.com/gohugoio/hugo/issues/new"
   ### Edit content repo (set to automatically enter "edit" mode; this is good for "improve this page" links)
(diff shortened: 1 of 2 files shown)


digitalcraftsman:
Thanks Ryan! To prevent broken links it might be a good idea to use `with` to [check if the variables for the footer links are actually set](https://github.com/gohugoio/gohugoioTheme/blob/acf8b548f86809cee7bd5f8d3aeefcf340ab890c/layouts/partials/site-footer.html#L22-L24).

rdwatters:
Ah, very good point, but I think that belongs in the theme repo since it's at the templating layer, no?

digitalcraftsman:
That's correct. I just saw it while looking into this pr.

digitalcraftsman:
Fixed it in https://github.com/gohugoio/gohugoioTheme/commit/ac239e1525ab06cf0d64ca03d64bc0c3624d7fdd

PR #121, opened 2017-08-02
Add missing forward slash to baseurl example

gaetawoo:
A missing slash in baseurl can trip up a lot of new users (like it did me)


gaetawoo:
Resubmitting as a different PR

PR #123, opened 2017-08-02
Add missing closing and opening comment tags

gaetawoo:
The comment tags were split between lines and therefore caused the content within them to render as html.


digitalcraftsman:
Thanks again for your ongoing submissions :+1:

PR #125, opened 2017-08-05
Update title.md

ronliskey:
Added a pipe example.

### DIFF
--- a/content/functions/title.md
+++ b/content/functions/title.md
@@ -23,3 +23,12 @@ aliases: []
 ```
 {{title "BatMan"}}` → "Batman"
 ```
+
+Can be piped to combine with others.
+
+```
+{{ range $name, $items := .Site.Taxonomies.<YOUR_TAXONOMY> }}
+  <li><a href="{{ $.Site.BaseURL }}<YOUR_TAXONOMY>/{{ $name | urlize | lower }}">{{ $name  | humanize | title }} ({{ len $items }}</a>
+  </li>
+{{ end }}
+```
--- a/content/functions/title.md
+++ b/content/functions/title.md
@@ -24,11 +24,10 @@ aliases: []
 {{title "BatMan"}}` → "Batman"
 ```
 
-Can be piped to combine with others.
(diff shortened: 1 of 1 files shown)


content/functions/title.md
```diff
@@ -23,3 +23,12 @@ aliases: []
 ```
 {{title "BatMan"}}` → "Batman"
 ```
+
+Can be piped to combine with others.
+
+```
+{{ range $name, $items := .Site.Taxonomies.<YOUR_TAXONOMY> }}
+  <li><a href="{{ $.Site.BaseURL }}<YOUR_TAXONOMY>/{{ $name | urlize | lower }}">{{ $name  | humanize | title }} ({{ len $items }}</a>
```
rdwatters: Thanks for this submission, @ronliskey. Possible to rewrite this to use the safer/less fragile `relURL` or `absURL` functions for the anchor's URL construction rather than the global config value?

digitalcraftsman: Building a url this way can be error prone. It can happen that users defines the base url as `www.example.com` without a trailing slash in the end. In this case `www.example.com<YOUR_TAXONOMY>....` isn't a valid url.

Instead use something like this:

    {{ printf "%s/%s" <YOUR_TAXONOMY> ($name | urlize | lower) | absURL }}

As you can see we build the path to the resource and then let `absURL` (or `relURL` if you like) the prepending of `www.example.com`.

rdwatters: Oops. I wrote the exact same thing and didn't hit "submit" and left it pending. Please ignore the duplication @ronliskey

ronliskey: Thanks! Is the corrected version appropriate for the docs, or should such examples go somewhere else, maybe in discourse?

rdwatters: Definitely in the docs since that's the first place users will look for what's considered the "best practice" for general Hugo usage...

rdwatters: LGTM @digitalcraftsman

digitalcraftsman:
@ronliskey thanks for submitting this addition. I've merged your contribution as 5ff530ac.

PR #127, opened 2017-08-06
Explain how to add non-404 custom error pages.

ronliskey:
I spent too long last night doing what seemed logical at the time to create a custom 403 error page. I tried every permutation before realizing the system worked as intended, and 404 is handled differently from all other errors. A search shows many others waste time with the same experiments. I saw the discussions about this changing at some point in the future, but for now it would save needless frustration to explain the work-around.

### DIFF
--- a/content/templates/404.md
+++ b/content/templates/404.md
@@ -52,4 +52,16 @@ Your 404.html file can be set to load automatically when a visitor enters a mist
 * Amazon AWS S3. When setting a bucket up for static web serving, you can specify the error file from within the S3 GUI.
 * Caddy Server. Using `errors { 404 /404.html }`. [Details here](https://caddyserver.com/docs/errors)
 
+## Serving Other Custom Error Pages
+
+The above technique works only for 404 errors. To create additional custom error pages, such as 401 and 403, place them in the `static/` directory, and adjust your server settings for each file as noted in {{< relref "#Automatic Loading" >}}. 
+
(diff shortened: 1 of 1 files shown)


bep:
While this approach works, I'm not sure it is the best approach. What about creating content files (i.e. `md` or even `html` files) with some front matter, i.e.:

```
/content/403.md
/content/401.md
```

This way you get full template support, you can even put `layout = error` or something to have dedicated template for the error pages.

ronliskey:
Sounds good. I wasn't sure which you'd prefer, and went for the 'simpler' method, but this is certainly more flexible. Will update...

bep:
Curious: Why do you need 403 etc. error pages? In a normal "static site" situation, I don't see how you would use them?

ronliskey:
I found that directories on longer paths were accessible for index viewing. In `/tags/foo/bar', `foo` was listed all contents. I blocked index listing in `.htaccess`, but also wanted to provide a user friendly page. You can see it here:

http://ronliskey.com/tags/field-trips/

http://ronliskey.com/tags/

bep:
OK, but the better solution for you would be to use `disableKinds` and turn of the stuff you don't want generated, and the above would be a 404 error, which makes more sense to me in most cases.

ronliskey:
Not seeing much in the docs about `disableKinds`. Don't understand it yet. Off to do some experimenting...

PR #128, opened 2017-08-07
Add documentation for Blackfriday Extensions

kaushalmodi:


digitalcraftsman:
Hello @kaushalmodi,

thanks for taking the time to compile this more detailed list of all available extensions :+1:

kaushalmodi:
@digitalcraftsman Thanks for merging. I just checked https://gohugo.io/content-management/formats/#blackfriday-options, and I found an issue with the rendering of the `relref` links that worked fine locally on `hugo server`.. See the links in the *See Blackfriday extensions section for more information.* lines.

As the section is originally on `bfconfig.md` page in `hugoDocs` but on `formats` page on the final site, those `relref`s do not work. Can you please fix those?

kaushalmodi:
@digitalcraftsman I believe this will fix it.. can you please review: https://github.com/gohugoio/hugoDocs/pull/132

digitalcraftsman:
Thanks for pointing out this issue. If you only link a header on the same there's no need to use `relref`. Using `[link text](#id-of-header)` works just fine.

digitalcraftsman:
da978cc7 should fix it.

kaushalmodi:
Thanks. Any idea though why relref doesn't work on gohugo.io? Because they work fine on localhost.

On gohugo.io, the `relref`s stay unexpanded: `https://deploy-preview-132--gohugoio.netlify.com/content-management/formats/%7B%7B%3C%20relref%20%22#blackfriday-extensions" >}}`

digitalcraftsman:
That's strange. In none of the commits the links worked locally. My guess would be that the `bfconfig.md` file is just read and that the shortcodes aren't processed.

kaushalmodi:
Also I don't see (based on grepping) an actual use of `relref` anywhere on `hugoDocs`. The only occurrences are in code blocks and commented out shortcodes.

digitalcraftsman:
The search revealed the same. In the old doc the was a mixture of manual links and the use of `relref` to create cross references. `relref` should be able to create more stable links case something changes.

@rdwatters rewrote large portions of the docs. It might be helpful to know what he thought about `relref` vs manual links during the rewrite.

rdwatters:
> It might be helpful to know what he thought about relref vs manual links during the rewrite.

Largely a style preference; i.e., whenever I write long form content, I like to keep my links collected at the bottom, especially in documentation where a single link is referenced more than once). That's why you see links at the bottom of pages (although I'm sure I missed a few). 

Part of reworking everything was changing the filenames, which are used to create the slugs by default. I originally found myself using relref a lot, but then I would change filenames frequently to make it easier to keep the URL reflective of the filename.

Of course, you can change a `url` for a specific content file within the front matter, but then you're muddying site architecture/navigation with your content architecture, and I'm a fan of only the most crucial metadata traveling with an instance of a content type (hence my previous comments about the namespaces being part of FM, whereas the last site had `prev` and `next` in the FM of each file). Plus, this creates a disconnect between the filename, the inconsistencies in front matter, and the output directory structure/urls. The URLs for the last site were pretty goofy. 

> **I am by no means saying that relref is a bad feature**. In fact, it's really convenient and makes Hugo feel more like the typical DB-driven CMS. 

I just figured that while I was cleaning up, I might as well aim for consistency whenever possible. Moving forward, if contributors prefer the shortcode, I say have at it :smile: HTH. Thanks for contributing @kaushalmodi

PR #132, opened 2017-08-08
Fix reference to Blackfriday Extensions section

kaushalmodi:


kaushalmodi:
Nope.. it doesn't work.. help?

https://deploy-preview-132--gohugoio.netlify.com/content-management/formats/#blackfriday-options

The `relref`s are not rendering.. but they work fine locally on `hugo server`

PR #133, opened 2017-08-08
Details on what this variable could be used for

brunoamaral:
learned this one the hard way.

### DIFF
--- a/content/variables/git.md
+++ b/content/variables/git.md
@@ -19,6 +19,8 @@ toc: false
 wip: false
 ---
 
+This variable shows metada about a specific piece of content. It is useful for times when you are using Git and Hugo to manage a project or in a multi-author setting, for example. It will not be useful as a cache busting method.
+
 {{% note "`.GitInfo` Performance Considerations"  %}}
 Hugo's Git integrations should be fairly performant but *can* increase your build time. This will depend on the size of your Git history.
 {{% /note %}}


content/variables/git.md
```diff
@@ -19,6 +19,8 @@ toc: false
 wip: false
 ---
 
+This variable shows metada about a specific piece of content. It is useful for times when you are using Git and Hugo to manage a project or in a multi-author setting, for example. It will not be useful as a cache busting method.
```
rdwatters: \*metadata

That said, I think "metadata about a specific piece of content" is actually the description of front matter. I would instead recommend adding something to the effect of the following to the `note` admonition below instead:

> Hugo's Git integrations should be fairly performant but *can* increase build times for larger histories. As `.GitInfo` is specific to an individual content file, it is not an effective cache-busting method.

bep: I don't like this addition. I think this "using this as a cache busting method" was something @brunoamaral thought about and not something the Common Joe will think about. We cannot fill the docs with "this is not that ...", so unless more people chime in with the same confusion, we're keeping this as is. So, thanks for this, but we'll wait.

rdwatters: It's come up on other issues and the forum (i.e. cache busting), but I agree we don't need to add "Don't use X for this" for every Hugo feature. My suggestion was more about the language and placement :smile:

PR #137, opened 2017-08-12
fix incorrect syntax in content/news/http2-server-push-in-hugo.md

MunifTanjim:


bep:
Are you sure about this? I.e. what was wrong with the old?

MunifTanjim:
Header Rules Documentation: https://www.netlify.com/docs/headers-and-basic-auth/
Header Rules Validator: https://play.netlify.com/headers

If you check the previous one with the validator, it'll show the error.
Also, with the previous one, you'll get a error line in the Netlify deploy log.

bep:
1. I'm getting the headers I want from the site, which is really my one and only ultimate goal.
2. With your change, you add the Link headers to every response (images, css, ... everything), which  is not what I want.

If the Netlify validator is wrong, I suggest you take it directly with Netlify.

PR #150, opened 2017-08-19
fix terminal output example text

mbren:
In one of the Windows examples, some of the terminal output was rendering a \&gt; instead of >.


digitalcraftsman:
Hello @mbren,

thank you for fixing this issue :+1: 

Cheers,
Digitalcraftsman

PR #152, opened 2017-08-20
Remove typo for figure shortcode

GamingCoder:
The figure shortcode **does** provide a shorthand for the HTML5 figure element.

### DIFF
--- a/content/content-management/shortcodes.md
+++ b/content/content-management/shortcodes.md
@@ -74,7 +74,7 @@ Hugo ships with a set of predefined shortcodes that represent very common usage.
 
 ### `figure`
 
-`figure` is an extension of the image syntax in markdown, which does not provide a shorthand for the more semantic [HTML5 `<figure>` element][figureelement].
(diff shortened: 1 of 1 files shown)


rdwatters:
@GamungCoder Thank you for contributing, but the sentence is now incorrect with your change. 

There is no shorthand for `<figure>`in markdown.

bep:
>-`figure` is an extension of the image syntax in markdown, which does not provide a shorthand for the more semantic [HTML5 `<figure>` element][figureelement].

Does `figure` not create a `figure` element? If not, what does the sentence above then mean?

rdwatters:
"the image syntax in markdown" is the [referent](https://www.thoughtco.com/referent-grammar-1692033) being described in the second clause "does not provide a shorthand for the more semantic HTML5 <figure> element." 

The current sentence is correct in that Markdown syntax does *not* account for `<figure>`; it only accounts for ``.

We *could* rephrase as follows, but I think the current sentence is better:

> Hugo ships with a `figure` shortcode because markdown syntax does not account for the HTML5 `<figure>` element.

Regardless, the PR makes the sentence incorrect.

bep:
Ah, OK, now I ... kind of understand.

PR #156, opened 2017-08-21
Update Installing.md

stuartluscombe:
Adjust required Go version for compilation to 1.7+.
Remove requirement for govendor and separate go install command

When trying to follow this guide as-is on Go 1.6, govendor would download/install, but running `govendor get github.com/gohugoio/hugo` produced the error "Package "/home/stuart/src/go\n/src/github.com/gohugoio/hugo" not a go package or not in GOPATH."

Running instead `go get -v github.com/gohugoio/hugo` downloaded Hugo and all dependencies, but would not compile, giving an error of "github.com/gohugoio/hugo/commands/gendocshelper.go:61: enc.SetIndent undefined (type *json.Encoder has no field or method SetIndent)"

In order to resolve this problem it was necessary to upgrade my version of Go as the SetIndent method was not introduced into encoding/json until 1.7 (https://golang.org/doc/go1.7#minor_library_changes).

In my case I upgraded to Go 1.8 as that is what was available via a ppa, so without testing specifically on 1.7 I cannot confirm this works, but with the change being in the official docs I would assume it would.

### DIFF
--- a/content/getting-started/installing.md
+++ b/content/getting-started/installing.md
@@ -66,22 +66,15 @@ choco install hugo -confirm
 #### Prerequisite Tools
 
 * [Git][installgit]
-* [Go 1.5+][installgo]
-* [govendor][]
-
-#### Vendored Dependencies
-
-Hugo uses [govendor][] to vendor dependencies, but we don't commit the vendored packages themselves to the Hugo git repository. Therefore, a simple `go get` is *not* supported because the command is not vendor aware. *You must use `govendor` to fetch Hugo's dependencies.*
+* [Go 1.7+][installgo]
 
 #### Fetch from GitHub
 
 {{< code file="from-gh.sh" >}}
-go get github.com/kardianos/govendor
-govendor get github.com/gohugoio/hugo
-go install github.com/gohugoio/hugo
+go get github.com/gohugoio/hugo
 {{< /code >}}
 
-`govendor get` will fetch Hugo and all its dependent libraries to `$GOPATH/src/github.com/gohugoio/hugo`, and `go install` compiles everything into a final `hugo` (or `hugo.exe`) executable inside `$GOPATH/bin/`.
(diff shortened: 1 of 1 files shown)


stuartluscombe:
Thanks for the feedback.

I've just been trying to get govendor to work again, but everything I try just keeps coming back with the same error:

```
stuart@stuart-HP-EliteBook-745-G2 ~/go/bin $ govendor get -v github.com/gohugoio/hugo
Error: Package "/home/stuart/src/go\n/src/github.com/gohugoio/hugo" not a go package or not in GOPATH.
```

I can't see why govendor is adding in a \n to the path? Certainly I have the hugo files in my GOPATH

```
stuart@stuart-HP-EliteBook-745-G2 ~/src/go/src/github.com/gohugoio/hugo $ ls
appveyor.yml  CONTRIBUTING.md  helpers     media             source
...
```

I'll keep trying to work through this, but any suggestions would be greatly appreciated!

stuartluscombe:
After some testing I've discovered the following:

The `govendor get` command as given in the documentation just does not seem to  work, the same error (`not a go package or not in GOPATH.`) occurs with a fresh Go environment and after some searching I was not able to resolve this.

Running `go get github.com/gohugoio/hugo` pulls down hugo and a number of dependent libraries and also places a compiled, executable file in `$GOPATH/bin`.

If after running the above I run `govendor sync .` from within the `$GOPATH/src/github.com/gohugoio/hugo` folder then the dependent packages are pulled into the vendor folder. Running `go install github.com/gohugoio/hugo` at this point builds the code successfully.

One thing I did find is that the file size of the hugo binary was different with both the plain `go get` build and the one following the `govendor sync`:
```
-rwxrwxr-x 1 stuart stuart 18327168 Aug 22 19:25 hugo-goget
-rwxrwxr-x 1 stuart stuart 18742533 Aug 22 19:32 hugo-govendorsync
```

So while govendor does seem to add something to the mix I cannot immediately see what.

I would at the very least suggest the documentation include the following:

```
go get github.com/gohugoio/hugo
cd $GOPATH/src/github.com/gohugoio/hugo
govendor sync .
go install github.com/gohugoio/hugo
```

Please let me know what you think.

bep:
If you have problems with govendor I suggest you bring that up on the discussion forum.

PR #158, opened 2017-08-22
Abstract the type in the lookup order

gotgenes:
The _default pieces still had a concrete type of "post".


gotgenes:
I'm unsure how to trigger netlify to get it to reattempt building this commit.

digitalcraftsman:
Hello @gotgenes,

thanks for your submission. Generalizing the example makes sense.

A rebuild of the preview should be triggered once you commit a new change AFAIK. When clicking on "Show all checks" and then "Details" in the Netlify row, shouldn't you see your changes in the preview?

Cheers,
Digitalcraftsman

PR #159, opened 2017-08-22
Add missing css lang argument

gotgenes:
Makes the triple-backtick example equivalent to triple-tilde example.


digitalcraftsman:
Hello @gotgenes 

thank you for submitting this addition.

Cheers,
Digitalcraftsman

PR #160, opened 2017-08-25
updated .gitlab-ci.yml

aborruso:
The "Completion of the Hugo Quick Start" is an assumption, and there you install a hugo theme via git submodule, than it's necessary to add the rows I have added.

### DIFF
--- a/content/hosting-and-deployment/hosting-on-gitlab.md
+++ b/content/hosting-and-deployment/hosting-on-gitlab.md
@@ -40,6 +40,10 @@ In the root directory of your Hugo site, create a `.gitlab-ci.yml` file. The `.g
 {{< code file="gitlab-ci.yml" >}}
 image: monachus/hugo
 
+before_script:
+  - git submodule init
+  - git submodule update --force
+
 pages:
   script:
   - hugo


maiki:
Rather than adding those lines to the example deploy script, I think we ought to clarify the assumption.

For instance, the Quick Start uses MacOS as the example OS. However, it links to an installation doc for options.

In that spirit we should keep the example as simple as possible, and link to the [`.gitlab-ci.yml` docs](https://docs.gitlab.com/ce/ci/yaml/README.html) instead.

Adding a theme via submodule isn't the choice for everyone, and since the doc in question is an intermediate topic, they will have other options available to them.

bep:
I agree with @maiki  on this. We might get more clever in the future with an "OS selector" etc., but we have to keep the documentation at a short, clear and maintainable level -- and the way to do that is by delegation; i.e. llnk to external doc as much as we can.

aborruso:
Hi @bep  and @maiki first of all, thank you.

To mantain the documentation clear and maintainable level is a great goal, and hugo has good documentation.

But I'm a beginner.

I have read [this page](https://gohugo.io/hosting-and-deployment/hosting-on-gitlab/) and here one of the assumptions is the [Completion of the Hugo Quick Start](https://gohugo.io/getting-started/quick-start/). 
Than I have read it and used `git submodule ` and followed the instructions of Host on GitLab, but without success.
I think that a lot of beginner could do the same error.

What to you think to edit "Completion of the Hugo Quick Start" and remove `git submodule` part and use a more generic way?

Once again thank you

maiki:
> I have read this page and here one of the assumptions is the Completion of the Hugo Quick Start.
Than I have read it and used git submodule and followed the instructions of Host on GitLab, but without success.

The first assumption is "[w]orking familiarity with Git for version control". That covers submodules. I understand that the GitLab hosting page isn't beginner-friendly, but hosting isn't that easy for anyone; one needs to know the specifics of their hosting scenario, and for GitLab Pages, that means knowing enough about Git to setup your own `.gitlab-ci.yml` config.

The assumption about completing the Quick Start isn't to confirm a person knows about the git commands, it is instead to confirm a person knows how content and themes and config files all relate to each other in a Hugo site. The Quick Start isn't intended to teach git commands.

Hence my suggestion to link to a learning resource, for those folks that finished the quick start but may not have the requisite knowledge to host on GitLab. Maybe a line mentioning the Quick Start and a link to the GitLab doc.

> But I'm a beginner.

That is very relative when it comes to Hugo. It is a commandline-based static website generator (at the minimum). So if you want to make a case for the docs not making sense for a particular persona, you'll have to be more specific. :smile:

aborruso:
Hi @maiki I do not want a doc page for me :)

Probably my proposal is very stupid, and thank you for you time.

My step was:

- read "hosting-on-gitlab" page;
- than complete the Hugo Quick Start;
- than do what I read in "hosting-on-gitlab" and have no success.

I have used a Kinder Surprise toy instruction approach, with no brain.

In my opinion, there should be some note "hey, remember to add install submodule" in gitlab.

Once again thank you

PR #167, opened 2017-08-28
Fix wrongly named default value of publishDir from "publish" to "public"

marvhock:


digitalcraftsman:
Hello @marvhock 

thank you for submitting this improvement. Good catch!

Cheers,
Digitalcraftsman

PR #169, opened 2017-08-30
Changed config's syntax order matching description

ccamara:
Description states: 
> you need to change the Hugo publish directory in your site's config.toml and config.yaml, respectively:

but after that, the first piece of code was written using `yaml` syntax and the second using `toml`'s.

I have kept description and changed block's order.


digitalcraftsman:
Hello @ccamara 

thank you for submitting this improvement. This way it's definitely more consistent.

Cheers,
Digitalcraftsman

ccamara:
You're welcome! Since I am a completely newbie to Hugo, toml and deplyoing sites to gitlab and github, I will be detecting any lack of consistency, as I am reproducing every step. 

Thanks for your job!

PR #172, opened 2017-08-31
generating static assets on deploy rather than having to do it manually

sdomino:
Sorry, quick update.

Rather than making people run "hugo" manually to generate static assets, I've updated the boxfile to do it for them on deploy. Much better workflow.


digitalcraftsman:
No problem. Improvements are always welcomed.

PR #182, opened 2017-09-08
Add documentation for `cond` function.

### LINKED ISSUE #180: Document cond template func
See https://github.com/gohugoio/hugo/pull/3868

jetwash:
Fixes #180.

### DIFF
--- a/content/functions/cond.md
+++ b/content/functions/cond.md
@@ -1,0 +1,26 @@
+---
+title: "cond"
+date: 2017-09-08
+description: "Return one of two arguments, depending on the value of a third argument."
+categories: [functions]
+menu:
+  docs:
+    parent: "functions"
+signature: ["cond CONTROL VAR1 VAR2"]
+aliases: [/functions/cond/]
+hugoversion: 0.27
+relatedfuncs: [default]
+toc: false
+draft: false
+needsexamples: false
+---
+
+`cond` returns *VAR1* if *CONTROL* is true, or *VAR2* if it is not.
+
+Example:
+
+```
+{{ cond (eq (len $geese) 1) "goose" "geese" }}
+```
+
+Would emit "goose" if the `$geese` array has exactly 1 item, or "geese" otherwise.


content/functions/cond.md
```diff
@@ -0,0 +1,26 @@
...
+draft: false
+needsexamples: false
+---
+
+`cond` returns *VAR1* if *CONTROL* is true, or *VAR2* if it is not.
+
+Example:
+
+```
+{{ cond (eq (len $geese) 1) "goose" "geese" }}
```
digitalcraftsman: `cond (len $geese | eq 1) ...` might be a bit easier to read

bep:
This is appreciated, but just a reminder: This isn't released ... planning to cut a release on Monday.

bep:
As to documentation, I like the clarity of this:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

content/functions/cond.md
jetwash: I disagree. I find the pipe syntax to be quite overused. It's great for thinking about stuff functionally, but for stuff like this it's often easier not to think about it functionally.

jetwash:
Wow, yeah that's pretty good. TBH I was having a lot of trouble figuring out what to say without just telling people it's basically the ternary operator. Sometimes expert knowledge makes documentation a lot harder…

MDN is [CC-BY-SA 2.5](https://developer.mozilla.org/en-US/docs/MDN/About#Copyrights_and_licenses), could we just steal their text and adopt it to Hugo?

content/functions/cond.md
bep: I agree with @jetwash in this case about the pipe (which is very useful in other cases), but the prime example, in this case, is to conditionally set a variable based on some condition.

bep:
>MDN is CC-BY-SA 2.5, could we just steal their text and adopt it to Hugo?

With this short a text I would say it would come in under "fair use". I would not be surprised if Mozilla borrowed some snippets from Hugo, so this goes both ways.

bep:
But if you borrow more than just the basic description, I suggest you add "Source: MDN/ CC-BY-SA 2.5" below.

digitalcraftsman:
Any updates here?

content/functions/cond.md
```diff
@@ -0,0 +1,26 @@
+---
+title: "cond"
+date: 2017-09-08
+description: "Return one of two arguments, depending on the value of a third argument."
```
rdwatters: @jetwash 

Thanks for putting this together. How about the following for `description` instead?

> Returns one of two arguments according to the truthiness of a condition. 

Or

> Assigns one of two values to a variable according to the truthiness of a condition.

Still could be room for improvement, but the current description feels like it could apply to more than just `cond`.

digitalcraftsman:
Any updates here?

PR #190, opened 2017-09-14
I imagine hugo means git version 2.05

jodumont:
because git 2.5 still don't exist considering they are at version 2.14

### DIFF
--- a/content/hosting-and-deployment/hosting-on-github.md
+++ b/content/hosting-and-deployment/hosting-on-github.md
@@ -23,7 +23,7 @@ GitHub provides free and fast static hosting over SSL for personal, organization
 
 ## Assumptions
 
-1. You have Git 2.5 or greater [installed on your machine][installgit].
+1. You have Git 2.05 or greater [installed on your machine][installgit].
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
> because git 2.5 still don't exist considering they are at version 2.14

2.14 is greater than 2.5, meaning that 2.5 has to exist and that 2.14 is the newer version. Here's the release of [Git 2.5](https://github.com/git/git/releases/tag/v2.5.0).

Git 2.5 was released 2015, Git 2.**0**5 around 2010. Based on that I guess that Git 2.5 should work fine.

digitalcraftsman:
Any updates here?

i-give-up:
Git uses semantic versioning so I don't think "2.05" is a valid version string. "2.0.5" on the other hand is a version that exists.

As digitalcraftsman said, 2.5 comes before 2.14 so I don't think there's anything wrong with the document in question in its current state.

digitalcraftsman:
Due to inactivity and the discussion regarding the Git versioning I'll close this issue.

PR #191, opened 2017-09-18
Escape code fencing

williamhaley:
Looks like the code fencing is misformatted here.

oops

I believe the intention is to escape the backticks in the code fences so that the example looks like this.

ah


digitalcraftsman:
Thanks for your submission @williamhaley :+1:

PR #196, opened 2017-09-19
Add `none` as a value for `layout`

robsonsobral:
Useful for FAQs, for example.

### DIFF
--- a/content/content-management/front-matter.md
+++ b/content/content-management/front-matter.md
@@ -110,7 +110,7 @@ There are a few predefined variables that Hugo is aware of. See [Page Variables]
 : the meta keywords for the content.
 
 `layout`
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Hi Robson,

great to see you again and thanks for opening this pull request :+1: To proceed please sign the CLA by clicking on the yellow badge in the comment above. It only has to be done once.

bep:
I'm not a fan of this. It kind of works, and will continue so, but it will create a WARNING and we should not have tips in the docs that creates warnings ... This is a workaround, and until we get a better solution for this, these workarounds should live in the discussion forum, me thinks.

robsonsobral:
> [...] we should not have tips in the docs that creates warnings [...]

It's a valid point, @bep . I totally agree. I can try to put a conditional to bypass the template lookup in case of `none`, but I never ever coded using GoLang.

I gonna take a look at this, but i don't know exactly when.

digitalcraftsman:
Due to Bjørn Erik's comment and the possibility to handle this case in the code base of Hugo itself I'll close this issue.

PR #202, opened 2017-09-24
Fix typo in migration tools section

cberner:


digitalcraftsman:
Thanks for your submission @cberner :+1:

PR #207, opened 2017-09-26
Add a note about the new Emacs package ox-hugo

kaushalmodi:
Mention the new Emacs package ox-hugo: https://ox-hugo.netlify.com


digitalcraftsman:
Thanks for adding your Emacs package @kaushalmodi

PR #209, opened 2017-09-26
Change name of post files in example to use post prefix

marvhock:
I guess in the code structure example "event" was used as prefix for the post files by accident, so I changed that.


digitalcraftsman:
Thanks Marvin for fixing the code example in the documentation :+1:

PR #210, opened 2017-09-26
Document math.Log template function

### LINKED ISSUE #99: Document math.Log template func
The documentation might accidentely got lost during the merge of the old and new docs.

Here are the original ones: https://github.com/gohugoio/hugo/commit/34c566773a1364077e1397daece85b22948dc721#diff-b0459ca979b7cda94af36279c6f4c38fR436

digitalcraftsman:
Closes #99

### DIFF
--- a/content/functions/math.md
+++ b/content/functions/math.md
@@ -20,15 +20,15 @@ draft: false
 aliases: []
 ---
 
-| Function     | Description                                                                 | Example                        |
-|--------------|-----------------------------------------------------------------------------|--------------------------------|
-| `add`        | Adds two integers.                                                          | `{{add 1 2}}` &rarr; 3         |
-| `div`        | Divides two integers.                                                       | `{{div 6 3}}` &rarr; 2         |
-| `mod`        | Modulus of two integers.                                                    | `{{mod 15 3}}` &rarr; 0        |
-| `modBool`    | Boolean of modulus of two integers. Evaluates to `true` if result equals 0. | `{{modBool 15 3}}` &rarr; true |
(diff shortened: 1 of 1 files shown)


content/functions/math.md
```diff
@@ -20,15 +20,16 @@ draft: false
...
 
-| Function     | Description                                                                 | Example                        |
-|--------------|-----------------------------------------------------------------------------|--------------------------------|
-| `add`        | Adds two integers.                                                          | `{{add 1 2}}` &rarr; 3         |
-| `div`        | Divides two integers.                                                       | `{{div 6 3}}` &rarr; 2         |
-| `mod`        | Modulus of two integers.                                                    | `{{mod 15 3}}` &rarr; 0        |
-| `modBool`    | Boolean of modulus of two integers. Evaluates to `true` if result equals 0. | `{{modBool 15 3}}` &rarr; true |
-| `mul`        | Multiplies two integers.                                                    | `{{mul 2 3}}` &rarr; 6         |
-| `sub`        | Subtracts two integers.                                                     | `{{sub 3 2}}` &rarr; 1         |
-|--------------|-----------------------------------------------------------------------------|--------------------------------|
```
digitalcraftsman: @moorereason did you add the divider because floor, ceil and round are a logical unit or because those template funcs require the package name in order to be called?

If it's the latter case I'll move `math.Log` there as well.

moorereason: I put it in there to see how it looked but actually meant to delete it before committing!  You can do what you want with it.

digitalcraftsman: As divider not intended to provide any further meaning I'll remove it.

PR #212, opened 2017-09-26
Fix link type

maiki:


digitalcraftsman:
Thanks for your submission @maiki :+1:

PR #224, opened 2017-10-02
fix `ref` and `relref` shortcode usage patterns

vassudanagunta:
The documentname parameter to `ref` and `relref` shortcodes must include file extension. Including the file extension in the given usage patterns will make it much easier for users to follow.


digitalcraftsman:
Thanks for your submission @vassudanagunta.

PR #225, opened 2017-10-02
Remove delimiters in YAML and TOML examples

i-give-up:
The use of `---` and `...` delimiters in the YAML examples seem to
suggest that those delimiters are necessary in a .yaml/.yml file when in
fact they are optional. Also, it seems to be common practice to *not*
use `---` and `...` in .yaml/.yml files. Similarly, it's less common to
use `+++` in .toml files. The examples on TOML's Github repo don't use
it, and Hugo-generated .toml files don't use it either.  This commit
removes the delimiters mentioned above to conform with the more common
styles.

Additionally this modifies the example for shortcut to find
configuration option

- add `cd` to make it clear that user has to change to the hugo site's
  directory first
- separate the command and the output of the command

### DIFF
--- a/content/getting-started/configuration.md
+++ b/content/getting-started/configuration.md
@@ -34,10 +34,9 @@ In your `config` file, you can direct Hugo as to how you want your website rende
 
 ## YAML Configuration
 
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Thanks for your submission @i-give-up

PR #226, opened 2017-10-02
Clarify that .Lastmod automatically uses .GitInfo.AuthorDate

kaushalmodi:
If the `lastmod` field in the content front matter is not set if `.GitInfo` is
enabled, Page's .Lastmod is auto-set to .GitInfo.AuthorDate.

### DIFF
--- a/content/variables/git.md
+++ b/content/variables/git.md
@@ -51,4 +51,8 @@ The `GitInfo` object contains the following fields:
 `.Subject`
 : commit message subject (e.g., `tpl: Add custom index function`)
 
+## `.Lastmod`
+
+If the `.GitInfo` feature is enabled, **and** if the `lastmod` field in the content's front matter is not set, `.Lastmod` (on `Page`) is fetched from Git i.e. `.GitInfo.AuthorDate`.
+
 [configuration]: /getting-started/configuration/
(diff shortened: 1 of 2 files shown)


content/variables/git.md
```diff
@@ -51,4 +51,21 @@ The `GitInfo` object contains the following fields:
 `.Subject`
 : commit message subject (e.g., `tpl: Add custom index function`)
 
+### Debugging `.GitInfo`
```
bep: Please restrict the commit to what the commit message says. I like your Lastmod change below, not the debug part here, which has no place on this page.

kaushalmodi: OK, makes sense.. I will remove it from this commit. Let me know the correct location for this.. I just wanted to make a point that `range` cannot be used. I had made that mistake and this this showed up on discourse ( https://discourse.gohugo.io/t/iterate-over-gitinfo/8472 ). So I thought this should be useful to others too.

kaushalmodi:
I have force-pushed the requested change.

kaushalmodi:
@bep Pinging just in case you missed the update as suggested.

digitalcraftsman:
Thanks @kaushalmodi

PR #230, opened 2017-10-10
Update quickstart

marcanuy:
Quickstart step 4 makes reference to outdated instructions (https://gohugo.io/getting-started/quick-start/#step-4-add-some-content).

As it shows the command output, it used to build a draft by default, now it is generating a regular post with empty values. To make the post display something when running the server, example content should be added.

Also there is no need to start the server with the `-D` option as new there isn't any draft, and `date` has an example value, but if none is specified it will just ignore it displaying an error in console.

### DIFF
--- a/content/getting-started/quick-start.md
+++ b/content/getting-started/quick-start.md
@@ -81,24 +81,38 @@ hugo new posts/my-first-post.md
 ```
 
 
-Edit the newly created content file if you want. Now, start the Hugo server with [drafts](/getting-started/usage/#draft-future-and-expired-content) enabled:
+Having the front-matter skeleton, edit the newly created content at `content/posts/my-first-post.md` and add example content:
+
+~~~
++++
+title = "This is my first post"
+date = "2017-10-10"
+tags = []
+featured_image = ""
+description = "A new post added"
++++
+
+This is an example content.
+~~~
+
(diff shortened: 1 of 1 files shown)


bep:
>Quickstart step 4 makes reference to outdated instructions

What is outdated?

The quick start is made deliberately sparse, so I'm not keen on adding stuff. I'm happy to get corrections, but I'm not sure what you mean about the above.

marcanuy:
When creating content: `hugo new posts/my-first-post.md`, the created file has:

~~~
+++
title = ""
date = ""
tags = []
featured_image = ""
description = ""
+++
~~~

So nothing is displayed when serving or building the site at homepage.

The guide suggest to **start the Hugo server with drafts enabled:**, and shows `1 of 1 draft rendered`, this is why I thought it was referring to an older version of Hugo where drafts where *true* by default.

I think it would be good to add at least a basic title to the post to show anything when serving the site for first time, and also the `-D` flag can be removed from `hugo server -D`.

---
Using: Hugo Static Site Generator v0.29 linux/amd64 BuildDate: 2017-09-26T16:42:55-03:00

marcanuy:
I had a dependency problem with Ubuntu that used an old version that generated the empty post. It is creating the post with basic content and drafts=true as expected.

PR #247, opened 2017-10-25
shortcode-templates.md: Remove stray period.

XhmikosR:


digitalcraftsman:
Thanks for you contribution @XhmikosR

PR #249, opened 2017-10-29
Mention the available 'width' argument in 'figure' shortcode

kaushalmodi:


digitalcraftsman:
Thanks @kaushalmodi for your addition and for improving the figure shortcode :+1:

PR #250, opened 2017-10-30
Updated permalink example URL

### LINKED ISSUE #244: URL Management / time zones / Documentation wrong
2 part-er for docs at <https://gohugo.io/content-management/urls/#permalinks>

Using the example from the documentation:
```
[permalinks]
  post = "/:year/:month/:title/"
```
> Only the content under post/ will have the new URL structure. For example, the file content/post/sample-entry.md with date: 2017-02-27T19:20:00-05:00 in its front matter will render to public/2017/02/sample-entry/index.html at build time and therefore be reachable at https://example.com/2013/11/sample-entry/.

1. The documentation is unclear as to whether `/:year/:month/` and in my case `/:year/:month/:day` are using UTC or local time. Thus, to use the example date of `2017-02-27T19:20:00-05:00` would that be rendered as `/2017/02/28/` (UTC) or as `/2017/02/27/` (local).
2. The part of _...and therefore be reachable at `https://example.com/2013/11/sample-entry/.`_ is just wrong and doesn't fit with the rest of the paragraph, I'm guess it is just a mistake.

( I moved this ticket here from a mis-placed issue in [hugo](https://github.com/gohugoio/hugo/issues/3986) )

jamesrhea:
Fixes #244

### DIFF
--- a/content/content-management/urls.md
+++ b/content/content-management/urls.md
@@ -43,7 +43,7 @@ permalinks:
   post = "/:year/:month/:title/"
 {{< /code >}}
 
-Only the content under `post/` will have the new URL structure. For example, the file `content/post/sample-entry.md` with `date: 2017-02-27T19:20:00-05:00` in its front matter will render to `public/2017/02/sample-entry/index.html` at build time and therefore be reachable at `https://example.com/2013/11/sample-entry/`.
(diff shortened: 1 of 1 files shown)


digitalcraftsman:
Thanks @mbleigh for fixing #244  :+1:

PR #252, opened 2017-11-01
shortcode-templates.md: update year example

julienschmidt:
The example used a deprecated way to get the current year:
`Page's Now is deprecated and will be removed in Hugo 0.31. Use now (the template func).`


digitalcraftsman:
Thanks Julien for fixing the example :+1:

PR #259, opened 2017-11-05
Add missing params key

chris-rudmin:


content/content-management/multilingual.md
```diff
@@ -43,7 +43,7 @@ title = "Mon blog"
 weight = 2
 [Languages.fr.params]
 linkedin = "lien-francais"
-[Languages.fr.navigation]
```
bep: I think the correct fix here is to remove the params above, i.e.  [Languages.fr.params] >  [Languages.fr]

bep:
> if so, do you know why?

Not sure why you would need a "why", but this has been discussed in other places -- there is an open issue about it in the main repo. The short gist of it is that everything in a language section is put into `params` for that language.

chris-rudmin:
@bep I must say very confusing. Is this the correct fix then?

chris-rudmin:
IMO, if the ‘why’ is not clear, people might unknowingly run into namespace conflicts with params and localizations...

PR #272, opened 2017-12-12
Include note to allow hugo write permissions outside home dir

Tails:
I can't ind the related issue right now where I also commented, but this documentation patch would fix the issue I had where Hugo couldn't create a new site in a partition that was mounted in my home folder using the snap package.


digitalcraftsman:
Thanks for adding the note @Tails :+1:

PR #274, opened 2017-12-13
Fix ref and relref function pages.

### LINKED ISSUE #264: Fix ref/relref template funcs doc
https://gohugo.io/functions/ref/

Both are missing the very important `Page` parameter. This is not needed in the shortcode case, because that is implicit.

felicianotech:
Fixes #264.

Gathered this information from a couple of Hugo forum posts as well as testing things out on my own website.


JuanKman94:
What's a "Regular" page, anyway? I can't find anything relating to that in the documentation

PR #280, opened 2017-12-15
Mention the significance of leading 0 in int fn string input

kaushalmodi:
https://discourse.gohugo.io/t/unable-to-cast-09-of-type-string-to-int/9614

### DIFF
--- a/content/functions/int.md
+++ b/content/functions/int.md
@@ -24,3 +24,28 @@ Useful for turning strings into numbers.
 ```
 {{ int "123" }} → 123
 ```
+
+{{% note "Usage Note" %}}
+If the input string is supposed to represent a decimal number, and if it has
+leading 0's, then those 0's will have to be removed before passing the string
+to the `int` function, else that string will be tried to be parsed as an octal
+number representation.
+
+The [`strings.TrimLeft` function](/functions/strings.trimleft/) can be used for
+this purpose.
+
+```
+{{ int ("0987" | strings.TrimLeft "0") }}
+{{ int ("00987" | strings.TrimLeft "0") }}
+```
+
+**Explanation**
+
+The `int` function eventually calls the `ParseInt` function from the Go library
+`strconv`.
+
+From its [documentation](https://golang.org/pkg/strconv/#ParseInt):
+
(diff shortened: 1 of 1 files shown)


moorereason:
Can you also leave a hint about using `strings.TrimPrefix` to remove leading zeros?

kaushalmodi:
@moorereason I prefer `replaceRE` as it is more robust. Use of `strings.TrimPrefix` needs to be hard-coded i.e. one needs to know exactly what string prefix needs to be removed.

### `replaceRE` examples

All of these work (tested)

```
            {{ int ("987" | replaceRE "^0+" "") }}
            {{ int ("0987" | replaceRE "^0+" "") }}
            {{ int ("00987" | replaceRE "^0+" "") }}
            {{ int ("000987" | replaceRE "^0+" "") }}
```

Note that the `replaceRE` expression can stay the exact same (`replaceRE "^0+" ""`) for any case.

### `strings.TrimPrefix examples

These work (tested):

```
            {{ int ("987" | strings.TrimPrefix "0") }}
            {{ int ("0987" | strings.TrimPrefix "0") }}
            {{ int ("00987" | strings.TrimPrefix "00") }}
```

But these don't (tested) as the prefix doesn't match exactly:

```
            {{ int ("00987" | strings.TrimPrefix "0") }}
            {{ int ("0987" | strings.TrimPrefix "000") }}
```

kaushalmodi:
@moorereason I have updated this PR with `replaceRE` example, in case you'd like to move forward with that.

moorereason:
My bad.  I should have said `strings.TrimLeft`.

kaushalmodi:
@moorereason Done! Thanks for the review. The `strings.TrimLeft` is a good suggestion especially for folks not comfortable with regex. I did not know of that fn earlier.

*I know I will always reach out for `replaceRE` for cases like these 😄 *

BTW performance wise, is one better than another?

moorereason:
The Trim funcs should be faster than RE.

kaushalmodi:
@digitalcraftsman Can you please merge this?

kaushalmodi:
@digitalcraftsman @bep Can you please merge this? Thanks.

PR #281, opened 2017-12-16
Update roadmap

maiki:
Found an e wandering around, fortunately it knew its address.


digitalcraftsman:
Thanks for your contribution @ariejan :+1:

PR #288, opened 2017-12-21
Update rss.md

YaguraStation:


bep:
Thanks for this, but I think we will have to redo this. Neither values are actually correct for the recent Hugo versions. The correct way is to use `Permalink`.

PR #300, opened 2018-01-17
Update quick-start.md

arikalmen:
The next step for creating the first post results in an error if you try to create a post without moving the theme.

`Error: Unable to find theme Directory: /Users/arikalmen/arikalmen.com/themes/after-dark`


bep:
I believe that must be a symptom of something else. I have tested the quick start several times without any error.

PR #304, opened 2018-01-20
Fix example code.

DravenK:
In the example, the end symbol is missing and the content cannot be output.


DravenK:
I made some mistakes.

PR #305, opened 2018-01-21
Remove duplicate template lookup paths

bengadbois:


bep:
That file is auto-generated. I suggest you create an issue in the main repo.

PR #310, opened 2018-01-24
remove obsolete link to hugo roadmap

salim-b:
there's no roadmap (anymore), at least I couldn't find it anywhere on gohugo.io


bep:
Thanks for these. For the future, it would be good if you could start your commit title with a capital letter ("Remove ...").

salim-b:
> Thanks for these. For the future, it would be good if you could start your commit title with a capital letter ("Remove ...").

Aight, will do!

PR #320, opened 2018-01-31
Add note regarding custom 404 and hugo server

fbnlsr:
Ref #874


kaushalmodi:
Thanks for your PR.

I tried to address this using a slightly different word phrasing in https://github.com/gohugoio/hugoDocs/commit/6b626fac61e43894a4c560073ee27ef33e8e6a37.

PR #329, opened 2018-02-05
Add basics for links in multilingual context

Jibec:
https://discourse.gohugo.io/t/solved-reference-multilingual-content-from-template/9461/19


Jibec:
oh, this only works in templates...

Issue #322, opened 2018-02-01
Add Showcase


bep:
Work branch: 

* https://github.com/gohugoio/hugoDocs/tree/next
* https://next--gohugoio.netlify.com/

Let us finish it in this issue.

bep:
@budparr I have refactored it to use bundles and image resources. There are probably some more small items on that list (I'm not sure about the image sizes, but I have to say that the new image API in Hugo is a pleasure to use -- just have to remember running `hugo --gc` before committing to Git).

On my TODO list:

* Add some simple metadata (title) to the featured image
* Complete the creator etc. data model
* On a slightly unrelated topic, I think we need a "global RSS feed" with news+showcase. I guess we can add a RSS template to home page. That way we can release the showcases a little-by-little and make it news items on the Twitter feed etc. (undraft them).

There is some "funkynes" on the single page that you could look at:

* The prev-next looks ... odd?
* When there are content (https://next--gohugoio.netlify.com/showcase/esolia/) it looks a little untidy. It will probably help with more text/paragraphs, but maybe the featured image should be integrated better with the text? Not sure.

bep:
Also:

The list page only shows the first item on mobile:

https://next--gohugoio.netlify.com/showcase/

budparr:
> The list page only shows the first item on mobile:

I'm guessing that's an issue with lazyload. I'll have a look when I do the rest.

bep:
@spf13 could you share your thoughts on this, you are way better in this area than me ...

When we get technical stuff working, I was thinking that instead of just publishing a set of screenshot we could

1. Ask the site owners to tell/write a little story about the site. Some self-advertising would be fine, but it should at least mention Hugo ...
2. Then publish/updraft those one by one and make it a "news item".

I think I will reach out to the paceup.com CEO, as he has shown great interest in Hugo and that site is both multilingual and uses all the new features.

I'm assuming that some will not bother to take the time to respond, so we will eventually just post the screenshot with a link etc. ...

What do you think?

Also, what kind of metadata model would be natural to ask them to fill out (most of them being optional)?

DirtyF:
@bep you mean like http://www.trysmudford.com/perfomance-wins-with-hugo-and-netlify/ ? 😉

bep:
@DirtyF exactly; in that case I would ask the author if we could cross-post that same article. But I would settle for less; the most important part (in most cases) is to show to people that the site is built with Hugo.

budparr:
[...]

PR #330, opened 2018-02-06
Iterating on Showcase

budparr:
I've cleanup up the single page template and re-arranged the page to make the navigation work better.

Reference #322

### DIFF
--- a/themes/gohugoioTheme/layouts/partials/previous-next-links-in-section-with-title.html
+++ b/themes/gohugoioTheme/layouts/partials/previous-next-links-in-section-with-title.html
@@ -2,11 +2,11 @@
 {{/* this div holds these a tags as a unit for flex-box display */}}
 	<div date-pref class="white">
 		{{if .NextInSection}}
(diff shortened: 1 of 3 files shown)


bep:
The "Showcase" h1 title does not make much sense hanging there. I would just remove the h1 element.

budparr:
May seem obvious to me and you where you're at, but I think people need wayfinding.

bep:
>wayfinding

Can you elaborate on that ...?

bep:
This looks very good:

budparr:
Yeah, just letting people know where they are. It's quite clear in the docs, without say docs, but here it's just an image, basically.

bep:
This looks untidy:

image

budparr:
> Untidy
 
Are you still speaking of the "Showcase" heading?

bep:
We can add showcase to the "<title>" element, but not alone in a h1 attribute.

bep:
>Are you still speaking of the "Showcase" heading?

That is the only difference between the 2 screenshots.

budparr:
Only seeing one screenshot. But at any rate, I disagree, but defer to you.

budparr:
Done

PR #335, opened 2018-02-06
Fix an issue where whitespace wrap caused scrollbars on some browsers

budparr:
Ref #322 
Fixes the issue from this comment in https://github.com/gohugoio/hugoDocs/issues/322#issuecomment-363543231


kaushalmodi:
Looks great now! https://deploy-preview-335--gohugoio.netlify.com/showcase/brightline-org

image

PR #338, opened 2018-02-07
Mention Talkyard, an embedded commenting system for Hugo and others.

kajmagnus:


kajmagnus:
Maybe at the same time you'd like to remove `https://txtpen.com/` — seems it's gone, the website won't load and doesn't reply to `ping`.

kajmagnus:
Submitting new, with different link & that also fixes the `txtpen` commenting system broken link.

PR #352, opened 2018-02-12
WIP- Make "Last Updated" table searchable

kaushalmodi:
[**PREVIEW**][1]

---

- The maintenance-pages-table partial now takes a dict argument with keys "id"
  and "pages".
  - "id" is the unique table id to be assigned to each maintenance page table.
  - "pages" is the list of pages (which was the original one and only context
     passed to this partial).
- Increase "Last Updated" table entries from 10 to 500.
- Use https://datatables.net/ table filtering/searching/ordering script to
  manage that huge "Last Updated" table.

---

**Need help**

How do I not make that table fit the whole page? Once that is fixed, I believe this PR would be ready for merging.. WDYT @bep?

I know that this happens because of:

```html
<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.10.16/css/jquery.dataTables.min.css">
```

But I cannot also completely remove that, else the formatting of paging links at the bottom of the table get messed up. 

[1]: https://deploy-preview-352--gohugoio.netlify.com/maintenance/

### DIFF
--- a/layouts/maintenance/list.html
+++ b/layouts/maintenance/list.html
@@ -1,4 +1,14 @@
 {{ define "main" }}
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.10.16/css/jquery.dataTables.min.css">
+<script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.10.16/js/jquery.dataTables.min.js"></script>
+<script type="text/javascript">
+ $(document).ready(function() {
+     $('#recent').DataTable( {
+         "order": [[ 0, 'desc' ]]
+     } );
+ } );
+</script>
 <div class="w-100 ph4 pb5 pb6-ns pt1 mt4 pt3-ns">
     <div class="flex-l">
         <div class="order-2 w-100 w-20-l ph5-m ph0-l mb4 sticky">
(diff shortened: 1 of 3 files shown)


kaushalmodi:
@bep I see that the deploy previews do not run with `--enableGitInfo`.. was wondering if it's OK to enable that by default i.e. enable gitinfo in the hugoDocs config.toml itself?

bep:
We should maybe start by asking: Do we need this?

It is not free (as in: No work now or in the future), so we should really need it.

Do we?

kaushalmodi:
> We should maybe start by asking: Do we need this?

I found DataTables useful in showing huge tables in digestable form (as you see in that preview). If someone wants to see any recent change in "bundle" documentation, they can type that in the search field. So this feature is like something someone wished was already present :)

>It is not free 

To be honest, I was aware that this was MIT licensed, but never saw the [Purchase][1] page.. and now looking at the Purchase page, it still says:

> DataTables is free open source software, available under the MIT license, and you are free to download it and use as you see fit.

So the purchase is necessary only if support (forum access) is needed. So it's not **completely free**. 

>(as in: No work now or in the future), so we should really need it.

I did not follow that.

---

In summary, I don't feel too strongly about this PR getting merged or not.. thought of sharing this as it could be useful for folks not using git.

[1]: https://datatables.net/purchase/index

bep:
>To be honest, I was aware that this was MIT licensed,

I did not mean money. Time. Your time. My time. 

> If someone wants to see any recent change in "bundle" documentation,

Why not go to the bundle page and look at the footer? You are talking hypotheticals. The reason I spent time on this feature in the first place was to list pages marked as in need of some attention (TODO shortcode etc.). The "last changed" was a nice side effect.

So, to do even more work on this page, requires real and important use cases, not just "it would have been nice/cool" etc.

kaushalmodi:
> Your time. My time.

I agree. This PR just added a partial coolness factor.. because the CSS issues still need to be fixed :P

So, am closing this PR.

---

To my side question, should gitinfo be enabled in the hugoDocs config.toml itself?

PR #356, opened 2018-02-15
Update hosting-on-gitlab.md

onedrawingperday:
I am sending this proposal because Hugo users need to be warned in advance about the implications of hosting their Hugo site with a custom domain on Gitlab Pages.

See the current Gitlab Pages custom domains notification: https://about.gitlab.com/2018/02/05/gitlab-pages-custom-domain-validation/

And the past IP change that broke custom domains on Gitlab Pages last May: https://gitlab.com/gitlab-com/marketing/issues/883

Gitlab Pages is a free service  but not the only one and others tend to warn people via email about breaking changes unlike Gitlab. So there is no excuse.

### DIFF
--- a/content/hosting-and-deployment/hosting-on-gitlab.md
+++ b/content/hosting-and-deployment/hosting-on-gitlab.md
@@ -22,6 +22,8 @@ aliases: [/tutorials/hosting-on-gitlab/]
 
 [GitLab](https://gitlab.com/) makes it incredibly easy to build, deploy, and host your Hugo website via their free GitLab Pages service, which provides [native support for Hugo, as well as numerous other static site generators](https://gitlab.com/pages/hugo).
 
+N.B. You need to exercise caution when deploying your Hugo website with a custom domain to Gitlab Pages. Sometimes Gitlab introduces changes that can cause site outage with no prior user notification.
+
 ## Assumptions
 
 * Working familiarity with Git for version control


maiki:
My hot take is: if something requires a warning, it shouldn't be listed on that page.

We aren't reviewing those as a project, we just want to have a list of links for users to go check out, and while the uptime issue is serious, I don't think it is for goHugo to take on. ^_^

onedrawingperday:
@maiki and @kaushalmodi if you think that a warning is inappropriate maybe at @bep or @digitalcraftsman or @quoha should close this pull request. I'm leaving it as is for others to review it.

I am of the opinion that Gitlab Pages should be completely removed from the Hugo hosting section. Sending people there is not a very good idea IMNHO and it does not reflect well on the Hugo project itself. 

I originally found out about Gitlab Pages through this very section in the Hugo Docs back in the day. And I have come to regret this choice, in light of recent and not so recent events.

Anyway. I have said all there is to say about this.

moorereason:
I agree with @maiki in that we shouldn't get in the practice of grading/reviewing services.  If we're worried about appearing to endorse these hosting services, we should add a disclaimer to all pages about 3rd party services.  It's up to the user to do their due diligence in researching the service.

onedrawingperday:
@moorereason there should be a minimum threshold for listing a service. Uptime is the bare minimum. And in my case I've researched before hosting with them.

But unfortunately they have become very unreliable.

Disclaimers are generic little things but it's better to have one than none.

So +1 for putting a disclaimer on third party hosting services listed in the Docs.

onedrawingperday:
Closing this since this Pull Request has been rejected

See new issue at: https://github.com/gohugoio/hugoDocs/issues/358

maiki:
I wanted to note that I opposed taking steps to warn about that service. To clarify, I meant that if we felt we needed a warning for GitLab Pages, we ought to instead remove it.

Netlify works for a huge demographic, and I think most of us personally recommend that in the forums all the time. But for purposes of the hosting page, that just means it is listed with the others. If something happens to the service or company, we'd want to change the page to reflect that as well.

Also, hope your new host works out better, @onedrawingperday. ^_^

PR #362, opened 2018-02-17
Use GOPATH variable if defined in installation from source

Calinou:
This makes the command work on systems with a custom GOPATH environment variable set. If not defined, it will fallback to `$HOME/.go`.


Calinou:
Just saying, I made a typo in the commit description: the fallback path is `$HOME/go`, not `$HOME/.go`. Sorry!

PR #372, opened 2018-02-26
Shortcode tips

bmansheim:
These are things that I noticed. I might not be understanding something but if these are correct they could help others.

### DIFF
--- a/content/content-management/shortcodes.md
+++ b/content/content-management/shortcodes.md
@@ -50,6 +50,12 @@ Here are two examples of paired shortcodes:
 
 The examples above use two different delimiters, the difference being the `%` character in the first and the `<>` characters in the second.
 
+{{% tip %}}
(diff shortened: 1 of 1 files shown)


bep:
I'm pretty sure you don't need a shortcodes folder; or, if that is true, it is also a bug.

anthonyfok:
bep wrote on 2018-02-26: 
> I'm pretty sure you don't need a shortcodes folder; or, if that is true, it is also a bug.
> 
> What version of Hugo are you using?

Hi @bmansheim, regarding bep's question to your "Shortcode tips" pull request to hugoDocs, could you please follow up and elaborate?  Many thanks!

bmansheim:
It was a while ago and I think I just didn't understand how shortcodes worked. If the 2nd point also doesn't make sense, go ahead and close the issue.

Thanks

PR #375, opened 2018-02-27
Update rss.md

inwardmovement:
`rel="feed"` is not validated by w3c:
> Bad value `feed` for attribute `rel` on element `link`: The string `feed` is not a registered keyword.

https://validator.w3.org/


kaushalmodi:
Thanks for the PR. Looks like this got fixed around the same time you submitted your PR.

PR #376, opened 2018-02-27
New Forestry logo

itsnwa:


bep:
@scottgallant I assume that this change comes from Forestry?

scottgallant:
@bep Yes. Thanks for checking :)

PR #378, opened 2018-02-27
Add docs about comments within templates

paulcmal:
Following a [discussion](https://discourse.gohugo.io/t/comments-are-rendered-then-stripped-bug/10692) on the forums, here's some documentation about how to deal with comments within templates.

Don't hesitate to edit my PR to better match your docs standards.

### DIFF
--- a/content/templates/introduction.md
+++ b/content/templates/introduction.md
@@ -361,6 +361,41 @@ Go considers the following characters whitespace:
 * carriage <kbd>return</kbd>
 * newline
 
+## Commenting templates
+
+In order to keep your templates organized and share information throughout your team, you may want to add comments to your templates. There are two ways to do that with Hugo.
+
+### Go templates comments
+
+Go templates support `{{/*` and `*/}}` to open and close a comment block. Nothing within that block will be rendered.
+
+For example:
+
+```
+Bonsoir, {{/* {{ add 0 + 2 }} */}}Eliott.
+```
+
+Will render `Bonsoir, Eliott.` and not care about the syntax error we have within the comment block.
+
+### HTML comments
+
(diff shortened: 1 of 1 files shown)


content/templates/introduction.md
```diff
@@ -361,6 +361,41 @@ Go considers the following characters whitespace:
...
+<!-- {{ $author := "Emma Goldman" }} was a great woman. -->
+{{ $author }}
+```
+
+Will render `Emma Goldman`. The templating engine will strip the content produced within the comment block (that is, nothing) but will still assign a value to the `author` variable that we can reuse outside of the comment block. If we had made an error within the comment block, the build would have been stopped with an `is an incomplete or empty template` error.
+
+If you really need to produce HTML comments from your templates, take a look at how we deal with [Internet Explorer conditional comments](#example-4-internet-explorer-conditional-comments) in Hugo. If you need to access variables within your HTML comments, just pipe printf to safeHTML. For example:
+
+```
+{{ printf "<!-- Our website is named: %s -->" .Site.Title | safeHTML }}
```
kaushalmodi: May be you should mention the `printf .. safeHTML` method first as it does not cause unnecessary evaluation of template code.

The part about Go template code wrapped with HTML comment delimiters should be mentioned after that with a caution note... that says something along the lines of "Don't do this".

kaushalmodi:
Another related discussion: https://discourse.gohugo.io/t/content-directories-per-environment/11026/7?u=kaushalmodi

paulcmal:
Is it more clear like this @kaushalmodi ? :)

kaushalmodi:
I squashed, rebased this manually in https://github.com/gohugoio/hugoDocs/commit/7851345c8ac913f43b7e92c329916b47c7c92bf4, with some sentence rephrasing.

Issue #382, opened 2018-02-28
partial vs template recommendation unclear


rawtaz:
In https://gohugo.io/templates/introduction/#includes it says "Starting with Hugo v0.12, *you may* also use the `partial` call for partial templates".

In https://gohugo.io/templates/partials/#use-partials-in-your-templates it says "When using Hugo v0.12 and newer, *be sure to* use the `{{ partial "<PATH>/<PARTIAL>.html" . }}` syntax".

These two pieces of text gives different impressions - it's unclear whether or not you SHOULD use `partial` instead of `template` with Hugo 0.12 and forward, or if it's just an option you have but without recommendation.

In short, one of them should be clarified to match the other :)

kaushalmodi:
@bep How about putting the exact same admonition block

> Before v0.12, Hugo used the template call to include partial templates. When using Hugo v0.12 and newer, be sure to use the {{ partial "<PATH>/<PARTIAL>.html" . }} syntax. The old approach will still work but has fewer benefits.

from https://gohugo.io/templates/partials/#use-partials-in-your-templates in https://gohugo.io/templates/introduction/#includes too, .. and remove the "Starting with Hugo v0.12, you may also use the partial call for partial templates:" line in the latter.

PR #383, opened 2018-02-28
Make the recommendation to use partial over template consistent

kaushalmodi:
Also, mention internal templates.

Fixes https://github.com/gohugoio/hugoDocs/issues/382.


kaushalmodi:
@bep Good to merge now? I have removed the mention of Hugo 0.12 in relation to partial.. should be cleared for new users.

PR #388, opened 2018-03-05
Remove deprecated Hugoidx, add native hugo solution

eddiewebb:
- Removed HugoIDX which is listed as deprecated 
>This project has been retired
>This project has been retired due to upstream API changes in Hugo.

- Adds example gist for client side searching with jquery, fuse.js. Requires no additional build-time process/workflow/tools except hugo!

### DIFF
--- a/content/tools/search.md
+++ b/content/tools/search.md
@@ -20,10 +20,10 @@ toc: true
 
 A static website with a dynamic search function? Yes. As alternatives to embeddable scripts from Google or other search engines, you can provide your visitors a custom search by indexing your content files directly.
 
-* [Hugoidx](https://github.com/blevesearch/hugoidx) is an experimental application to create a search index. It's built on top of [Bleve](http://www.blevesearch.com/).
 * [GitHub Gist for Hugo Workflow](https://gist.github.com/sebz/efddfc8fdcb6b480f567). This gist contains a simple workflow to create a search index for your static website. It uses a simple Grunt script to index all your content files and [lunr.js](http://lunrjs.com/) to serve the search results.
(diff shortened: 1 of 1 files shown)


eddiewebb:
You can see sample of search at https://edwardawebb.com/search/?s=dev

kaushalmodi:
@eddiewebb Pinging you here, as Gist hasn't *yet* implemented notifications.. *very strange..*

Here are my suggestions for changes in your post:

1. I had to add tokenize to get search results as expected. Please test that on your site too.
```js
var fuseOptions = {
    shouldSort: true,
    tokenize: true,
    matchAllTokens: true,
    includeMatches: true,
    threshold: 0.1,
    location: 0,
    distance: 100,
    maxPatternLength: 32,
    minMatchCharLength: 3,
    keys: [
        "title",
        "contents",
        "tags"
    ]
};
```
2. Remove unnecessary front-matter from `content/search.md`.. e.g. why have `sitemap`, `date`, etc. in there?
3. Keep the `search.html` layout template in `layouts/_default/` instead of `layout/page/`.. the `page` type/section seems to be specific to your theme.

That's all :)

eddiewebb:
Thanks @kaushalmodi,

I'll make those changes in the gist and my related resume theme that was just added.  Glad you were able to get the basic matching working. I'm also exploring other search tools as fuzzy searching is not really needed in this use case.

eddiewebb:
I've updated the gist and my theme with your feedback. Unfortunately `tokenize:true` has the consequence that the match indices returned by fusejs now represent the index of the token, not the characters, hence the mismatch in highlighting. To correct this I just highlight the searchquery if found instead of the fuzzy match.  Decreasing threshold and enabling tokenization should mostly produce exact matches, so that should be sufficient.

I've also adjust the weights so that title takes precedence over content, which takes precedence over tags.

kaushalmodi:
Thanks! Really happy to see you being receptive to my suggestions! I confirm the fixes for all the issues/suggestions I raised above. This ( https://scripter.co/search?s=leaf+bundle ) works now! 👍

PR #395, opened 2018-03-07
Fix new-post path in quick-start

cemsbr:


bep:
What was wrong with "post"?

cemsbr:
Nothing wrong with "post", but "posts". I was doing the quick-start with bootstrap4-blog template. When using the plural "posts", a new section besides "Home" was created and nothing new appeared in Home. After trying the singular "post", the post summary appeared in Home and there was not another section besides Home, as expected.

bep:
>Nothing wrong with "post", but "posts". I was doing the quick-start with bootstrap4-blog template. 

Yea, my typo. My point is that the "quick start" works fine as-is. I have tested it multiple times.

You should rather open an issue with the bootstrap theme to make it more flexible with section selection (possible use the `.Params.mainSections` values).

PR #396, opened 2018-03-07
Updated the directory path for creating section list.html

itshimesh:
After scratching my head for a couple of minutes and trying to figure out why my custom section list.html was not overriding the default one, I decided to try and follow the convention for single.html and it works. This should help others save that time.


kaushalmodi:
I just tried renaming my `_default/list.html` to `_default/section.html`, and everything worked as before. It might be better to first start a discussion on https://discourse.gohugo.io/ to discuss the problem you are seeing (also share your site source there showing the issue).

Closing this PR.

PR #399, opened 2018-03-08
Remove template reference to ordinal numbers

growdigital:
> Spelled-out cardinal numbers … are not currently supported

So, remove reference to them in "Hugo Date and Time Templating Reference"


kaushalmodi:
Reference: https://discourse.gohugo.io/t/date-format-not-adding-ordinals/10944

PR #400, opened 2018-03-09
add THEME location for lookup order under Main RSS

growdigital:
I'm doing all my templates in my theme, `/<THEME>/layout/rss.xml`, seems to work fine, so I'm guessing that the documentation for the lookup order for the main RSS template can be updated to include <THEME>.

Please tell me if I'm wrong!


bep:
Yes, your are right ... but I wrote this great little "layout lookup table generator" some time ago, which uses the actual implementation so we know it is correct. But there still seem to be some manually maintained listings in there. I will fix this in another commit and link to this to show how.

PR #404, opened 2018-03-09
Update Output Formats documentation (v0.38)

kaushalmodi:
**Merge AFTER v0.38 release**

Updates Output Formats documentation as per the change in v0.38.

### DIFF
--- a/content/templates/output-formats.md
+++ b/content/templates/output-formats.md
@@ -1,7 +1,7 @@
 ---
 title: Custom Output Formats
 linktitle: Custom Output Formats
-description: Hugo can output content in multiple formats, including calendar events, e-book formats, Google AMP, and JSON search indexes, or any custom text format. 
+description: Hugo can output content in multiple formats, including calendar events, e-book formats, Google AMP, and JSON search indexes, or any custom text format.
 date: 2017-03-22
 publishdate: 2017-03-22
 lastmod: 2017-03-22
@@ -60,11 +60,11 @@ suffix = "htm"
 mediaType = "text/html"
 ```
 
-**Note** that for the above to work, you also need to add an`outputs` definition in your site config.
+**Note** that for the above to work, you also need to add an `outputs` definition in your site config.
 
-## Output Formats
+## Output Format Definitions
 
-Given a media type and some additional configuration, you get an `Output Format`:
+Given a media type and some additional configuration, you get an **Output Format**.
 
 This is the full set of Hugo's built-in output formats:
 
@@ -122,9 +122,26 @@ The following is the full list of configuration options for output formats and t
 
 ## Output Formats for Pages
 
-A `Page` in Hugo can be rendered to multiple representations on the file system. By default, all pages will render as `HTML` with some of them also as `RSS` (homepage, sections, etc.).
+A `Page` in Hugo can be rendered to multiple *output formats* on the file
+system.
 
-This can be changed by defining an `outputs` list of output formats in either the `Page` front matter or in the site configuration (either for all sites or per language).
+### Default Output Formats
+Every `Page` has a [`Kind`][page_kinds] attribute, and the default Output
+Formats are set based on that.
+
+| Kind           | Default Output Formats |
+|--------------- |----------------------- |
+| `page`         | HTML                   |
+| `home`         | HTML, RSS              |
+| `section`      | HTML, RSS              |
+| `taxonomyTerm` | HTML, RSS              |
+| `taxonomy`     | HTML, RSS              |
+
+### Customizing Output Formats
+
+This can be changed by defining an `outputs` list of output formats in either
+the `Page` front matter or in the site configuration (either for all sites or
+per language).
 
 Example from site `config.toml`:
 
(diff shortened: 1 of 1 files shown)


content/templates/output-formats.md
```diff
@@ -122,9 +122,19 @@ The following is the full list of configuration options for output formats and t
...
-A `Page` in Hugo can be rendered to multiple representations on the file system. By default, all pages will render as `HTML` with some of them also as `RSS` (homepage, sections, etc.).
+A `Page` in Hugo can be rendered to multiple *output formats* on the file
+system.
 
-This can be changed by defining an `outputs` list of output formats in either the `Page` front matter or in the site configuration (either for all sites or per language).
+### Default Output Formats
+By default, **all** pages will render to the `HTML` output format. Every `Page`
+also has a [`Kind`][page kinds]. Except for the `page` `Kind`, by default, all
+other *Kinds* will also render to the `RSS` output format.
+
```
bep: I think the language here is a little technical.

I would suggest something in the line of: "By default, all ... HTML ... Also, every list page will, by default, render to the `RSS` output format."  

"list page" may not be perfect, but it is easy for people to understand.

kaushalmodi:
Preview of changed pages: [1], [2]

[1]: https://deploy-preview-404--gohugoio.netlify.com/templates/output-formats/#readout
[2]: https://deploy-preview-404--gohugoio.netlify.com/templates/section-templates/#page-kinds

content/templates/output-formats.md
```diff
@@ -142,12 +152,19 @@ outputs:
   page: ["HTML"]
 ```
 
+In the above examples, the *output formats* for `section`, `taxonomyTerm` and
+`taxonomy` will be implicitly set to `["HTML"]`.
+
+{{% note %}}
+If you add an `outputs` config section, the *output format* for all the
+unspecified *Kinds* will be set to only `HTML`.
```
bep: I think you take the "half-empty" approach to this. I would much prefer something in the line of:

>If you add an `outputs` config section, you replace the Hugo default `output` configuration with your own and must provide a complete configuration. You will get `HTML` for any missing page `Kind` entry.

content/templates/output-formats.md
```diff
@@ -142,12 +152,19 @@ outputs:
...
+{{% /note %}}
 
-* The output definition is per `Page` `Kind` (i.e, `page`, `home`, `section`, `taxonomy`, or `taxonomyTerm`).
-* The names used must match the `Name` of a defined `Output Format`.
-* Any `Kind` without a definition will default to `HTML`.
+* The output definition is per [`Page` `Kind`][page kinds] (i.e, `page`, `home`, `section`, `taxonomy`, or `taxonomyTerm`).
+* Any `Kind` not defined in this `outputs` config section will default to `HTML`.
```
bep: You have already said this before.

content/templates/output-formats.md
kaushalmodi: OK

content/templates/output-formats.md
kaushalmodi: But it's clearer that way. People will anyways needs to understand Kinds to be able to configure outputs.

kaushalmodi: And also, doesn't feel right mis-documenting.. RSS, 404, etc kinda are not "list pages".. I think that's what you were inferring too.

content/templates/output-formats.md
kaushalmodi: OK, will remove it from here.

content/templates/output-formats.md
kaushalmodi: I have attempted to make this  a bit more clearer in my last force-pushed commit to this PR, while retaining the correctness.

content/templates/output-formats.md
kaushalmodi: Done (this, and also the next hunk where you commented).

regisphilibert:
Even though home, page and section are rather obvious. I have been and saw a lot of people being confused as to what is a `taxonomy` page and a `taxonomyTerm` page. This PR is the perfect opportunity to fix that.
Following the recent [exchange in this thead](https://discourse.gohugo.io/t/solved-page-kind-are-swapped-for-taxonomy-and-taxonomyterm/10780/8), I think we could add a better description to each kind along the line of the following:

__home___: the homepage
__page__: a page showing one entry
__section__: a page listing entries from a section
__taxonomy__: a page listing entries from a given taxonomy term
__taxonomyTerm__: a page listing entries in a given taxonomy.

I use the word entry to avoid common confusion surrounding the word `page` in Hugo. But there maybe a better wording.

kaushalmodi:
@regisphilibert Thanks for the feedback. I have now rephrased that part taking in your input.

regisphilibert:
> A page listing all *taxonomies* under a *taxonomy term*

I think this wording is confusing. I'm sure not every body agrees on what a taxonomy is, but the documentation clearly states what a taxonomy is and what a term using its movie example. Following this logic:

A `taxomyTerm` page lists all **terms** belonging to a given **taxonomy**. 
So if the taxonomy is sport, the `taxonomyTerm` page of sport will list 
- football
- baseball
- tennis

While the `taxonomy` page of football will list every entries belonging to `football`.

regisphilibert:
Also I believe the evocation of the layout and content file is a bit confusing as it does not always follow the logic of the kind naming. This is the simple approach I would recommend:

home
 : The home page
 
 page
 : A page showing one entry or regular page
 
 section
 : A page listing entries from a given section
 
 taxonomy
 : A page listing entries from a given taxonomy term

 taxonomyTerm
 : A page listing terms from a given taxonomy

kaushalmodi:
How about now?

regisphilibert:
Yes :) I still have reservation about the HTML structure illustrations as not everybody goes looking under the public/ hood. But that’s just my view.

kaushalmodi:
> I still have reservation about the HTML structure illustrations

I didn't understand the reference.

kaushalmodi:
@bep Do you want to move me as a PR to hugo core, or do we just wait for v0.38 release?

bep:
> just wait for v0.38 release?

Wait. Less work.

PR #411, opened 2018-03-10
Fix some typos

psychelzh:


content/variables/taxonomy.md
```diff
@@ -28,7 +28,7 @@ For example, the following fields would be available in `layouts/_defaults/terms
 : The singular name of the taxonomy (e.g., `tags => tag`)
 
 .Data.Plural
-: The plural name of the taxonomy (e.g., `tags => tags`)
```
bep: The above is the Markdown syntax for a definition list item, which we want to keep.

psychelzh: @bep Thank you for your review, but the later commit fixed that.

bep: Ah, ok. Sorry.

Issue #363, opened 2018-02-17
Add "toggle" config language shortcode


regisphilibert:
It was discussed earlier this month that the docs should be able to present one code block with multiple tabs to toggle config languages (yaml, toml, json)

@bep created the [transform.Remarshal](https://github.com/gohugoio/hugo/commit/d382502d6dfa1c066545e215ba83e2e0a9d2c8d7) template function for this purpose.

regisphilibert:
@bep, @budparr, @kaushalmodi (anyone I should have mentioned...)

I'm only thinking design as yet but, already I need some point of views...
Here are three concept for the look of this codeblock.
We will have up to four elements in there:
- filename
- 3 config lang tabs
- copy button
To make sure this doesn't break if the filename is too long we may have to use flexbox, so won't rely solely on tachyons.

![capture d ecran 2018-02-17 17 43 32](https://user-images.githubusercontent.com/1480503/36346567-34c3e7c8-140e-11e8-822f-ceb9a2e32589.png)

Above is the cleanest way but in case a longer file name it won't look good, see below:

![capture d ecran 2018-02-17 17 58 15](https://user-images.githubusercontent.com/1480503/36346569-4bdf4c04-140e-11e8-8738-7cb3ca794505.png)

So I'm thinking more something like this:
![capture d ecran 2018-02-17 17 45 56](https://user-images.githubusercontent.com/1480503/36346572-5e264e58-140e-11e8-82d0-d8a3ffd69d82.png)

Because we don't need the file extension, we can always hide its overflow.

![capture d ecran 2018-02-17 18 22 59](https://user-images.githubusercontent.com/1480503/36346612-a4f13bd0-140f-11e8-88f4-13ccc0314bc0.png)

I'm thinking number 2 but if you have a better design concept, or if you think i'm overlooking something, feel free to jump in.

bep:
>I'm thinking number 2 

By that you mean the second screenshot? I kind of liked your third.

regisphilibert:
> By that you mean the second screenshot? I kind of liked your third.

No I meant the same as you, I made it clearer.

bep:
But this toggle should persist somehow. So when "I" pick

regisphilibert:
> But this toggle should persist somehow. So when "I" pick TOML then I want it to continue TOML.

1. Something like general user preference thing? 
2. Or just for the current page, bind one toggle event to the other code blocks on the page?

The first one will be more complicated and need more ux/prog specing.

[...]

PR #420, opened 2018-03-16
Config toggle

regisphilibert:
Will comment in issue


regisphilibert:
Next is old (7 Feb), this PR is a branch out of master, so I'd rather you guys merge it yourself rather than just me (odd feeling). Original branch from my fork is here: https://github.com/regisphilibert/hugoDocs/tree/config-toggle

bep:
This PR has a Git history that I cannot (or will not) spend time on merging, I will only say this:

1) Rebase is your friend
2) The `next` branch is by definition experimental -- so you cannot ruin anything. But it helps to have a tidy Git history for when you want it to go into master.

kaushalmodi:
@regisphilibert Looks like you had rebased config-toggle off master, so I just force pushed that to next.

Though, should we merge this directly into hugoDocs? Because a major chunk is the change in gohugoioTheme?

bep:
>Though, should we merge this directly into hugoDocs? Because a major chunk is the change in gohugoioTheme?

That is a pragmatical approach and the benefit of the subtree approach. Once it looks good in hugodocs we can forward the changes to the theme.

regisphilibert:
> That is a pragmatical approach and the benefit of the subtree approach. Once it looks good in hugodocs we can forward the changes to the theme.

My git self-confidence has never been so low (Kidding. Of course it has been).

I'm not sure on what I can rebase now that my branch already contains master's commit history though. 🤔

kaushalmodi:
Also the js needs to be minified (which I just learned about: `npm run build:production` from https://github.com/gohugoio/gohugoioTheme/blob/master/src/readme.md).

kaushalmodi:
@regisphilibert 

> I'm not sure on what I can rebase now that my branch already contains master's commit history though.

If this is about porting the hugoioTheme repo specific changes to that repo, I would just `cp -f` those files into that repo, then stage only the relevant parts, and discard the rest/outdated hunks.

.. and then minify the stuff in static/dist and commit everything.

regisphilibert:
Ok, if I want to start clean what would be the best way to proceed ? Maybe not use a fork. 
1. branch from Next directly
2. rebase master on my newly created branch
3.  add toggle code
4. commit
5. push

regisphilibert:
@kaushalmodi It's about the history of this PR. About the theme, I would need to know more about subtrees, never worked with those.

kaushalmodi:
Btw this works: https://next--gohugoio.netlify.com/getting-started/code-toggle/

I haven't yet thoroughly checked but looks good.. the JS would be slow to load as it is unminified at the moment.

regisphilibert:
@kaushalmodi How come it works without the PR working ? 🤔

bep:
That looks great! 

And even more fantastic if we could kind get rid the flicker effect on "remembering the tab".

regisphilibert:
> And even more fantastic if we could kind get rid the flicker effect on "remembering the tab".

Yes we had a lengthy exchange about that remember ? :)

There's no easy fix. We could hide the code-toggles until we have shown the saved pane. But there would still be a kind of flick.

If any one has suggestions, I take them!

regisphilibert:
Question: What is the best process when working on hugoioTheme ? Working on right here in hugoDocs or in the theme's repo ?

bep:
>If any one has suggestions, I take them!

I will think about it, but it is a minor detail. This is a big, big improvement to what we have.

regisphilibert:
Ok I still have a lot of stuff to learn git wise...
You guys don't have to spend too much time explaining to me, but I'll summarize what I figured out. And you can come back to it when/if you have the time:

@kaushalmodi just singled out my commit from the PR and pushed it to Next ?
It has not been merged to the theme yet, so for now I'll keep improving my code from within hugoDocs' next branch.
Then we'll merge hugoDocs/themes/gohugoioTheme back to the theme's repo when done. (Curious to see how this step works)

regisphilibert:
I'll close the PR. Please see my notes on https://github.com/gohugoio/hugoDocs/issues/363 for further specing / suggestions.

kaushalmodi:
@regisphilibert 

> How come it works without the PR working ?

I just stole your Github keys :P j/k

*Github PR system is what Microsoft Word is to plain text documentation.. More frills, less power.* 

So I just fetch your remote and play with the commits locally, and then I can (force) push that to whatever branch on whatever remote where I can commit access.

> just singled out my commit from the PR and pushed it to Next ?

This is how all commits look to me:

image

- regis/master was on upstream/master~1, and regis/config-toggle was based off that.
- So I created a local branch off regis/config-toggle and rebased it off upstream/master.
- Then I checked out my local next branch and force reset it to my local config-toggle
- .. and force pushed that to upstream/next (which Netlify picked up to build that preview)

regisphilibert:
Thank you so much for the explanation @kaushalmodi ! Always nice to learn new stuff.

PR #426, opened 2018-03-19
remove duplicate "Example from site `config.toml`"

xa0082249956:


xa0082249956:
Sorry I misunderstand that.

PR #430, opened 2018-03-27
Add new .Site variable IsServer.

felicianotech:
This adds the docs for the new `.Site` variable flag I made.

This **should not** be merged unless the corresponding code gohugoio/hugo#4541 is merged first.


bep:
This is fine, but ... It would be even beter if you could just add the docs to the same PR in the same repo. That way I don't have to remember when to merge what.

felicianotech:
Closing this in favor of docs in to code PR: https://github.com/gohugoio/hugo/pull/4541

PR #438, opened 2018-04-04
csv & json files location outside of data/

divinerites:


kaushalmodi:
Can you please elaborate on this PR? Why is that statement needed? 

Would be good if you can document a use case in this thread so that we understand the motive of this change.

divinerites:
No problem :-)

CSV files can be processed but they need to be outside of /data/ folder. (see https://discourse.gohugo.io/t/error-message-failed-to-read-data-from-csv-even-it-is-working-fine/10763/7).

On this documentation page data-template.md, this fact is not clear. We talk about YAML, JSON, or TOML in data folder. Then after that we talk abour CSV files via getCSV, but it make the CSV file location ambigus.

So I propose to make a clear statement (and you have to edit my giblish so called english :-) that CSV should be located outside of data folder.

divinerites:
Ahhhh, and reading me again, I guess only the csv file (json is fine) should be outside of /data/. Sorry.

kaushalmodi:
@divinerites Thanks! I have reworded that and moved it to a different location in the doc; commited in https://github.com/gohugoio/hugoDocs/commit/f462b620f38b9ccd81b544c609fd9a344b254479.

PR #439, opened 2018-04-06
use code-toggle shortcode when relevant

regisphilibert:
Content Management ✅

(Except for one or two cases with useful comments resided in the code block)


bep:
This is great!

PR #454, opened 2018-04-18
Improve .Scratch entry

regisphilibert:
Hi Guys,

I finally set myself to improve the Scratch page with concised but yet helpful examples. I think the previous ones failed to include less technical users.

Feel free to improve the methods' one line descriptions, I know @bep is really good at that!

I thought about plugging my scratch post in there, but not sure it is really needed now.

cc/ @budparr @kaushalmodi

### DIFF
--- a/content/functions/scratch.md
+++ b/content/functions/scratch.md
@@ -26,6 +26,10 @@ In most cases you can do okay without `Scratch`, but due to scoping issues, ther
 See [this Go issue](https://github.com/golang/go/issues/10608) for the main motivation behind Scratch.
 {{% /note %}}
 
+{{% note %}}
+For a detailed analysis of `.Scratch` and in context use cases, see this [post](https://regisphilibert.com/blog/2017/04/hugo-scratch-explained-variable/).
+{{% /note %}}
+
 ## Methods
 
 `Scratch` is added to both `Page` and `Shortcode` -- with following methods:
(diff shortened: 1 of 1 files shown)


bep:
>I thought about plugging my scratch post in there, but not sure it is really needed now.

Scratch that ... :-)

content/functions/scratch.md
```diff
@@ -20,66 +20,90 @@ draft: false
...
+
+```go-html-template
+{{ .Scratch.Delete "greetings" }}
 ```
 
-{{% note %}}
-The examples above uses the special `$` variable, which refers to the top-level node. This is the behavior you most likely want, and will help remove some confusion when using `Scratch` inside page range loops -- and you start inadvertently calling the wrong `Scratch`. But there may be use cases for `{{ .Scratch.Add "key" "some value" }}`.
-{{% /note %}}
+## Scope
+The scope of the backing data is global for the given `Page` or `Shortcode`, and spans partial and shortcode includes.
```
kaushalmodi: Do you also want to mention the use case for "root level" scope `$`? I see that this commit completely removes that mention.

regisphilibert: Yes, I always found it more confusing than helping. $ is universal to any page context method. I don't think we should focus too much on that here. But I may be wrong.

bep: I agree with @regisphilibert -- I think these gotchas should be in one (or two: faq also maybe) place, and not sprinkled around everywhere. It removes focus.

regisphilibert:
Alright, I added the ref to the post.

Issue #65, opened 2017-07-20
Remove unused files below /static


bep:
There seem to be some unusded CSS etc. Probably due to a imperfect merge.

rdwatters:
This seems like a reasonable place to ask the question of where to keep additional design assets; e.g.; my affinity designer files when I recreated the Hugo theme, or any other additional Illustrator, etc files used. Do you have a preference @bep?

bep:
I have in my own personal projects put those below /work -- which sounds reasonable.

anthonyfok:
There were some files that I decided not to remove in PR #456, especially the various "Made with Hugo" or "Powered by Hugo" logos:

* static/share/powered-by-hugo.png
* static/share/made-with-hugo-long-dark.png
* static/share/powered-by-hugo-long.png
* static/share/made-with-hugo-long.png
* static/share/made-with-hugo-dark.png
* static/share/powered-by-hugo-dark.png
* static/share/powered-by-hugo-long-dark.png
* static/share/made-with-hugo.png
* static/share/hugo-tall.png

though these are not exposed by the current new Hugo website design yet.  Not sure what to do with them yet, hehe.

PR #456, opened 2018-04-21
Remove unused files below /static

anthonyfok:
These font, CSS, JavaScript and image files under /static
are no longer used, or were duplicated elsewhere,
after the revamp of the Hugo website in July 2017.

This should reduce the size of the Hugo source tarball by 13 MB or 48%.

Fixes gohugoio/hugoDocs#65


regisphilibert:
Glad I could help!

PR #458, opened 2018-04-21
Update delimit.md

hellbox:
Small fix: the original file called `.Params.tags`, but the title of the tags array in the front matter is `keywords`. This updates the two mentions to correctly call them out.


kaushalmodi:
@hellbox Thanks for noticing.. The actual fix is to undo the unintended change in https://github.com/gohugoio/hugoDocs//commit/158cee1b5c5fe6e6980c4978b038d3ff6f869ba6.. i.e. revert `keywords` back to `tags` :)

This is now fixed in https://github.com/gohugoio/hugoDocs/commit/8c67dc89a1d78094dd58900f9c10de57ca55685d.

PR #461, opened 2018-04-24
Add Chinese language for menus

LinuxSuRen:
I just translate the menu in this pr.
@anthonyfok


bep:
@anthonyfok you or some Chinese speaking person needs to follow up on this and maintain this. If we are going to translate beyond what is in this PR we probably need to talk a little. This is of great value, but it also comes at a cost: If I don't understand the language, I cannot do any PR merges etc.

I have 1 comment to this PR. Assuming we get more translations in the future, I suggest we reorganize the content into sub folders, e..g.

```bash
content/en
content/cn
```

The practical thing is to move en first, but @anthonyfok can just _do that_.

anthonyfok:
@LinuxSuRen Wow, you are moving a lot faster than I expected!  First, thank you for your effort!

@bep As always, thank you for your insight.  And yes, we do need to talk, and there are indeed some issues that we'd need to look into:

1. I suggest working off a new branch, say, m17n, instead of master, as we set up the multilingual infrastructure and work out all the kinks in the process.
2. Maybe we shouldn't copy all the English .md pages to .zh.md just yet especially when no translation has been done on them.
3. We need to provide the language namespace for both Simplfied and Traditional Chinese, hence we need to consider both "zh-Hans" and "zh-Hant" instead of just "zh".
4. Could we generate both zh-Hans and zh-Hant HTML from one "zh" source?  Similar to how it is done with Chinese Wikipedia, see, for example:
    * https://en.wikipedia.org/wiki/Chinese_Wikipedia
    * https://meta.wikimedia.org/wiki/Automatic_conversion_between_simplified_and_traditional_Chinese
5. We may need to think of ways to keep the translations in sync.  For example, with the Debian website at https://www.debian.org/, a translated page keeps record of the CVS ID of the original English page, and thus a change of the English page would trigger some kind of notification to the translators and readers that the translation may have become outdated.

These are some of my concerns, though in reality we could start with "zh" for just Simplified Chinese (after all, that's the biggest market, haha) and deal with zh-Hant later.

> The practical thing is to move en first, but @anthonyfok can just _do that_.

Great idea!  I shall learn how to do that.

anthonyfok:
Good news!  @liuzl has ported https://github.com/BYVoid/OpenCC, which converts between Simplified and Traditional Chinese, into a pure Go implementation called GoCC: https://github.com/liuzl/gocc

Better yet, @liuzl uses Hugo on his website!  https://zliu.org/

LinuxSuRen:
@anthonyfok Your news looks great. We should enhance Hugo's multi-language capability.

LinuxSuRen:
I think maybe we can continue to translate Hugo's docs before we figure out how to keep sync between the English version and other languages.
Yes, we should keep the diffrent language page in each subdirectory.
@anthonyfok @bep

bep:
Re the above:

* Do not copy ALL. Take one by one. There are ways to merge in missing translations from other languages in Hugo. It requires some thought/work, but it is much better to have just the translated files in the content/mylanguage
* All sites/pages in Hugo form a object graph that can be queried, so it should be possible to create "maintainance lists".

config.toml
```diff
@@ -250,10 +378,10 @@ anchor = "smart"
     url = "https://themes.gohugo.io/"
 
     [[menu.global]]
-      name = "Showcase"
+      name = "作品展示"
       weight = 20
       identifier = "showcase"
-      url = "/showcase/"
+      url = "/zh/showcase/"
```
anthonyfok: @LinuxSuRen These you shouldn't be changing, otherwise the English version is gone...

content/_index.zh.md
```diff
@@ -0,0 +1,49 @@
...
+    copy: "Hugo provides a robust theming system that is easy to implement but capable of producing even the most complicated websites."
+  - heading: "Capable Templating"
+    cta: Get Started.
+    link: templates/
+    color_classes: bg-primary-color-light black
+    image: /images/home-page-templating-example.png
+    copy: "Hugo's Go-based templating provides just the right amount of logic to build anything from the simple to complex. If you prefer Jade/Pug-like syntax, you can also use Amber, Ace, or any combination of the three."
+---
+
+Hugo is one of the most popular open-source static site generators. With its amazing speed and flexibility, Hugo makes building websites fun again.
```
anthonyfok: @LinuxSuRen No translation was made to this file, hence it shouldn't be committed now.

config.toml
LinuxSuRen: ok, Thank you

LinuxSuRen:
@anthonyfok It also got failed after ci build done. It's same with my local environment.
But it seems like don't affect anything.
@bep Do you know what's going on?

bep:
>@bep Do you know what's going on?

There is a "nil" pointer in the home page template -- which points to a missing check in the template itself, but also at some missing data. One would need to investigate to figure that out.

LinuxSuRen:
@bep Thank you for your explanation, but I can't figure it out the reason.

LinuxSuRen:
@anthonyfok @bep 
I think of the problem is a bug of Hugo Theme. report by CI build.
I just create a whole new site with Hugo Theme, got the same error.

anthonyfok:
My initial investigation: Simply adding

```toml
[languages]
  [languages.en]
  [languages.fr]
```

to config.toml would trigger the `Failed to render "/path/to/hugoDocs/themes/gohugoioTheme/layouts/index.html": reflect: Zero(nil)` error, which appears to come from the call to `{{ template "home_showcase_item }}` in themes/gohugoioTheme/layouts/partials/home-page-sections/showcase.html

But I am not quite sure how to debug further yet.  :blush:

bep:
@anthonyfok there is some logic in the showcase that assumes there is a showcase bundle ... Which is currently not the case in the Chinese site:

https://github.com/gohugoio/hugoDocs/blob/master/themes/gohugoioTheme/layouts/partials/home-page-sections/showcase.html#L7

If you just add a `if $showcasePages` or something to the above it should ... improve.

content/content-management/_index.zh.md
```diff
@@ -0,0 +1,20 @@
...
+    weight: 1
+keywords: [source, organization]
+categories: [content management]
+weight: 01	#rem
+draft: false
+aliases: [/content/,/content/organization]
+toc: false
+---
+
+静态网站生成器需要扩展前端和一些模板，使得具有伸缩行和可管理性。Hugo 不仅是给研发人员设计的，也同样适用于管理人员和写作的人。
```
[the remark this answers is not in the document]
LinuxSuRen: Thanks. You get the point.

LinuxSuRen:
@anthonyfok Sorry first, your last comment is lost because of my misoperation.

anthonyfok:
Translating to "front matter" as 前端功能 is inappropriate because it translates back to English as "front-end function".  Avoid the term "前端" because it is already prevalently known as "front-end" (端=end) in web development.

The term "front matter" came from book publishing, meaning 书籍正文前的材料〔序言、目录、用法说明等〕.  Some times known as 前页.

It is worthy to look at how others have dealt with this term.

For example:

* Hexo leaves it untranslated as "front-matter" in their Chinese documentation: https://hexo.io/zh-cn/docs/front-matter.html

* Jekyll CN translates it to 头信息 (head/header information):  https://jekyllcn.com/docs/frontmatter/

* Another Hugo documentation effort here http://www.zimustudio.com/hugo.html uses 文件头

Their translation might not be perfect, they are straightforward and won't be mistaken for something else, i.e., not confusing, hence worthy for our consideration.

LinuxSuRen:
I don't think of all those words are suitable. Maybe we should keep the original word.
@anthonyfok

bep:
@anthonyfok would it be more effective to create a "working branch" for this?

LinuxSuRen:
I agree. We can publish them when most of the content is translated well.

LinuxSuRen:
Maybe the work of Chinese localization is a long work, more Chinese people perhaps get starting to use Hugo if we can do with it as soon as possible.
In my team, many members are not well in English, but we all want to use Hugo.
So, after we finish the work, anyone also can improve Hugo's docs.

anthonyfok:
> I don't think of all those words are suitable. Maybe we should keep the original word.
@anthonyfok

Jekyll, Hexo and Hugo users are already using terms like 头信息 and 文件头 to mean the "front matter" since 2008 (found on over 1000 pages, please search on Google yourself), so please do not dismiss these translations so lightly.

What I would suggest, however, is to put the original English in parentheses beside the Chinese translation at least once on each page, i.e. "头信息 (front matter)".

anthonyfok:
"express", as a noun (作为名词使用), means  快车 (e.g. 东方快车 Oriental Express), 快递, or a type of rifle (步枪)

To mean "表达方式", you do not use "Fix express" (修理快车?), but rather "Fix expression".

LinuxSuRen:
Many people are better than me in translation, I just think it will wandrful if the main site supports the Chinese language. I can learn from those web site already exists.
And that will be great if you can invite more Chinese Hugo users to do these.

anthonyfok:
> @anthonyfok would it be more effective to create a "working branch" for this?

Yes, most definitely.  :-)

While I wrote earlier:

> I suggest working off a new branch, say, m17n, instead of master, as we set up the multilingual infrastructure and work out all the kinks in the process.

But since getting the existing website ready for m17n takes only a few commits (move to content/en, add a language switch menu) thanks to Hugo's mature m17n support...

... it is probably more appropriate to name that working branch "chinese" or "i18n-zh".

anthonyfok:
@LinuxSuRen, sorry for taking so long to get back to you.

The move of all English pages from content/ to content/en/ is now complete.  Please adjust your pull request accordingly.  For more background information, see:

* https://github.com/gohugoio/hugo/issues/4523
* https://github.com/gohugoio/hugo/commit/eb42774e587816b1fbcafbcea59ed65df703882a

Namely, please add `contentDir = "content/zh"` under `[Languages.zh]`, and move the translated Chinese markdown files into content/zh/.  For example, content/content-management/_index.zh.md should be moved to content/zh/content-management/_index.md, etc.

Many thanks for your help!

LinuxSuRen:
@anthonyfok Got it.

LinuxSuRen:
@anthonyfok Already done with it. Please review again. Thanks a lot.

huermos:
@anthonyfok Hello, I have started a translation project last month, and I have already translated some of posts, "about" and "getting-started" sections are almost done. (https://github.com/huermos/hugo-docs-chs) I'd like to continue to work here. Could you give me a brief process, like how to merge etc.

huermos:
@anthonyfok I did my first try to create a pull request. Let me know if I miss or mistake anything.

PR #467, opened 2018-05-01
 Move content/ into new contentDir content/en/

anthonyfok:
in preparation for adding new translations such as Chinese.
Thanks to @bep for the suggestion in anticipation of more languages in the future.

Paths for "readfile" shortcode are adjusted accordingly.

~~Also, because `--contentDir` is new in Hugo 0.38, the `min_version` in theme.toml is bumped to "0.38.2".~~ (superfluous, removed in a forced push)

See #461


LinuxSuRen:
@anthonyfok So, I need to modify my PR #461 ?

PR #468, opened 2018-05-02
Fix snap classic command in installing.md

Raqbit:
A simple fix for the installation command of the snap in classic mode.


anthonyfok:
Thank you for catching that!
Committed as 3c2fc32f12909f38c5c2595d3d64ad66616fc3bb
with compensation for changed paths (content/ → content/en/)

PR #476, opened 2018-05-08
Rename `taxonomies.html` to `taxonomy.html`

ashleytqy:
Rename `taxonomies.html` to `taxonomy.html`


kaushalmodi:
Thanks for catching that typo 👍

Issue #307, opened 2018-01-24
Update ref/relref documentation about optional extensions


kaushalmodi:
Update the `ref`/`relref` documentation with:

- Mention that the extension portion of the document name is optional. 
- Once v0.35 is released, also document the feature requested in https://github.com/gohugoio/hugo/issues/4312.. Now the index.?? portion for page bundles is optional too.

bep:
I suggest waiting for 0.35 and make a "overhaul" of this.

Jibec:
I would love to see this included in documentation, this is vital for internal links while and multilingual site, so the link is language agnostic!

kaushalmodi:
- [ ] Clarify that `ref` and `relref` can be used to link only regular pages (not section pages, taxonomy pages, etc.) ([ref][1])

[1]: https://discourse.gohugo.io/t/linking-to-the-index-page-for-a-section/10427?u=kaushalmodi

jmooring:
This was resolved on September 5, 2020 with https://github.com/gohugoio/hugoDocs/pull/1213.

Note that these statements were removed from the documentation because they are no longer true:

> ref looks up Hugo “Regular Pages” only. It can’t be used for the homepage, section pages, etc.

> relref looks up Hugo “Regular Pages” only. It can’t be used for the homepage, section pages, etc.

PR #477, opened 2018-05-10
Fix erroneous explanation of ref and relref shortcodes

vassudanagunta:
Also replaced "fragmentary links" with proper terminology.


kaushalmodi:
Thanks for working on fixing the ref/relref docs; they were long overdue: https://github.com/gohugoio/hugoDocs/issues/307.

The docs also need to mention that the `.md` extension is optional.. so it technically does not need the file logical name.. just the base name without externsions. 

Would you have time to deal with everything in that referenced issue above?

vassudanagunta:
@kaushalmodi, I'm sorry brother, I can't afford to spend time documenting features I don't use, much less those I am not convinced are a good idea, much less those ideas [whose shortcomings I already spend ***a lot*** of time mitigating](https://discourse.gohugo.io/t/getpage-ref-and-relref-renovations/12103) even though I don't use them because I do something [entirely different in my fork](https://github.com/Markaround/hugo-af#differences-from-mainline-hugo). 

The non-triviality of documenting these shortcodes (otherwise you'd just do it rather than ask me) is a sign that maybe I'm right. 

I'd save a lot of time if I didn't do more than make hugo work the way I want it to in my fork. But I spend considerable time contribute analysis and code upstream to parts of Hugo I never use simply because I'm trying to be a good open source citizen. Even though the overhead for such contributions is rather huge.  

I think Hugo's development has issues that hurt its robustness, its adoption as well as contributions from developers. But that's a whole other discussion. 😉

PR #478, opened 2018-05-11
Updates pygments gallery link

bradryanbice:
Different link shows more options for pygments styles.

### DIFF
--- a/content/en/content-management/syntax-highlighting.md
+++ b/content/en/content-management/syntax-highlighting.md
@@ -72,7 +72,7 @@ pygmentsCodefences
 : Set to true to enable syntax highlighting in code fences with a language tag in markdown (see below for an example).
 
 pygmentsStyle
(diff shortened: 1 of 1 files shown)


kaushalmodi:
Hello, I just reviewed the links.. the old and new links are different.. older one is for Pygments themes and newer one is for Chroma themes, and they both are functional. So it doesn't justify removing the old link. Also the presentation of the old link is better.. all themes are visible on a single page.

If you are the developer of the new xyproto link, I was curious if you can make all themes visible on a single page..

As it stands, it doesn't seem worthy to removed the old, but working Pygments themes link.

bradryanbice:
Fair enough. I'm not the developer of the page, just saw that most of those styles worked as well and thought it would be a helpful addition.

bradryanbice:
@kaushalmodi FYI this link shows all the styles on one page: https://xyproto.github.io/splash/docs/all.html

kaushalmodi:
[Really?][1] :P

[1]: https://github.com/xyproto/splash/issues/1

PR #480, opened 2018-05-13
Added a simple image example to help understanding

fieldingtron:
it was not clear how to do a simple example


davidsneighbour:
What happens when the code for `$img` matches multiple images? I think a better sample would be one that matches a specific image.

PR #481, opened 2018-05-14
Arch Linux: Partial upgrades are unsupported

GloverDonovan:
Using `pacman -Sy` is considered [bad practice](https://wiki.archlinux.org/index.php/System_maintenance#Upgrading_the_system) and should not be recommended here.


kaushalmodi:
I don't use Arch, but based on those thumbs ups to your suggestion, and also:

- https://www.archlinux.org/pacman/pacman.8.html#_examples (Official examples use **-Syu**)
- https://www.reddit.com/r/archlinux/comments/5111je/pacman_sy_vs_syu/

I'll merge this.

Issue #423, opened 2018-03-17
Remove duplication of Menu entry variables


kaushalmodi:
The same variables are explained in 2 places:

1. https://gohugo.io/variables/menus/
2. https://gohugo.io/content-management/menus/

PR #492, opened 2018-05-22
Update MenuEntry properties documentation

kaushalmodi:
This PR overhauls the documentation for the much awaited MenuEntry object properties (variables and functions).

.. and also fixes https://github.com/gohugoio/hugoDocs/issues/423 as a side-effect.


kaushalmodi:
@bep I am not able to find good explanation of what `IsCurrentMenu` and `HasCurrentMenu` do. Can you please summarize what they do? That's the only missing piece in this PR.

kaushalmodi:
I just updated this PR with the summaries of `IsCurrentMenu` and `HasCurrentMenu` as I understand it.

Can someone give a quick skim through this PR? 

If there are no concerns, I'll commit this in a day.

kaushalmodi:
Preview:
- https://deploy-preview-492--gohugoio.netlify.com/variables/menus/
- https://deploy-preview-492--gohugoio.netlify.com/functions/ismenucurrent/
- https://deploy-preview-492--gohugoio.netlify.com/functions/hasmenucurrent/

bep:
>This is a Page object function returning a boolean value. It returns true if the PAGE is the same object as the .Page in MENUENTRY in a given MENU.


I think it is important to use the word "method". 

`IsMenuCurrent` is a method on `Page`.

And the description does not make much sense to me, at least not on the practical level. What does it do, give?

I did not write this method, so others may chime in and suggest a better text.

kaushalmodi:
> I think it is important to use the word "method".

Thanks. I will update that. Also, from what I see, wouldn't probably every "function" documented be a **method** too? E.g. `.Title`, `.HasChildren`, etc.. in this PR are all tied to the MenuEntry object.

So should I use method for everything?

Also can you give few examples of methods and functions in Hugo so that I can understand the difference better?

> And the description does not make much sense to me, at least not on the practical level. What does it do, give?

Based on that example in [menu templates][1] , if you are looping through menu entries in `.Site.Menus.MENU` in the context of PAGE, and if the menu entry loop variable is MENUENTRY, and if PAGE==MENUENTRY.Page, then `IsMenuCurrent` returns `true`.

How would I translate that in plain English?

[1]: https://gohugo.io/templates/menu-templates/

kaushalmodi:
@bep Is there any objection in merging this PR? We can always rephrase `IsMenuCurrent` and `HasMenuCurrent` explanation in future. For now, I just have "something" rather than nothing :)

kaushalmodi:
Merged this commit after rebasing to latest master in https://github.com/gohugoio/hugoDocs/commit/c0430f8f971b1a2475e486fdb716b0c7202fa6f7.

PR #496, opened 2018-06-01
Remove language assuming knowledge of GitHub OAuth

deyton:
The old wording made it sound like everyone has seen this modal before. Beginners may be discouraged if this is the first time they've come across the GitHub authorization modal. My thoughts are this is similar to what is described here: https://css-tricks.com/words-avoid-educational-writing/#article-header-id-6


kaushalmodi:
Sounds good. Thanks!

PR #508, opened 2018-06-25
add openbsd howto install

wondratsch:


kaushalmodi:
I have not verified the instructions for the OpenBSD installation, but I'll assume that you have :)

Thanks.

Issue #471, opened 2018-05-02
Document new per-language contentDir (Hugo v0.38), and add a comprehensive tutorial on multilingual website


anthonyfok:
See https://gohugo.io/news/0.38-relnotes and https://gohugo.io/news/0.38.2-relnotes, for example.

Right now, multilingual related features seem to be (from my vague memory) separated on a few different pages that are not linked with one another.

Also investigate `defaultContentLanguageInSubdir` because I still don't fully understand it.

regisphilibert:
Hey there! Can't promise a delivery date but I'd love to document this feature.
I'll dive into it (using bep's comments in https://github.com/gohugoio/hugo/commit/eb42774e587816b1fbcafbcea59ed65df703882a as reference) and come back if some question pop up.

regisphilibert:
Just to clarify how Page Bundle are treated here:

`/content/english/about/index.md` > English
`/content/french/about/index.md` > French, linked with content file above

`/content/english/about/index.md` > English
`/content/french/about.md` > French, not inked with content file above

regisphilibert:
I just want to make sure I fully comprehend how this works. Most of it is explained by @bep in the commit comment, but the Page Bundles are left out. This is what I understand about how Page Bundles are being treated shall we opt for a "language per directory" content structure:

`/content/english/about/index.md` > English 
`/content/french/about/index.md` > French, **linked** with content file above 

`/content/english/about/index.md` > English 
`/content/french/about.md` > French, **not inked** with content file above (shouldn't it be linked?)

It appears in the templates, Page Bundles from both languages are treated as one big bundle, each linked page being able to access the other one's Resources. This is great :)
One exception is for files with the same name (relative path?) being present in both Bundles. In this case only the resource from the page being treated is returned by `.Resources`. 
This is also great, because in many cases, you will look for say a "header" file which may or may not be different depending on your language.

Also:
> Also investigate defaultContentLanguageInSubdir because I still don't fully understand it.

From what I understand, you have to set it to true if your default language directory is not at the root of your project. (as in the `content/english`, `content/norwegian` exemple). 
Why do we need this set when the language `contentDir` is defined in the language param, I don't know... Maybe @bep can explain so we can mention this in the doc, this seems important.

I'll post this comment and hopefully won't need any more clarification.

bep:
Your first question:

* For bundles, the content files (*.md, *.html etc) are per language. You will not inherit a content file in another language.
[...]

PR #509, opened 2018-06-25
Mutlilingual: Document "content directory per language" system

regisphilibert:
Following https://github.com/gohugoio/hugoDocs/issues/471

I grouped the three first paragrahs under "Configure Languages" considering all three dealt with configuration matters and improved them slightly.

Most of my additions are grouped underneath "Translate Your Content".
We know make a distinction between two language management/translation systems: 

- Filename (about.fr.md)
- Placement (content/french/about.md)

The naming of the second one may be questioned, if so, feel free to jump in with suggestions.
I tried among those systems to make a clear distinction between "language assignment" and "page linking".

I also tried and uniform the examples as English and French (Only cause I can to handle the latter without google translate, and french country code and basic words are more widespreadly understood I think).

/cc @kaushalmodi @budparr (I don't want to hassle you with review requests but, I could add you as reviewers if you have some input to nudge in there)


budparr:
About the naming, shouldn't `norwegian` be a country code?

regisphilibert:
No you set the language directory in the config, in our case it is `content/french` (that norwegian part was a leftover from early edits). I learned a lot while doing this :)

PR #512, opened 2018-07-03
Include "summaryLength" config option

brennanangel:
*Updated to include "summaryLength" configuration option from 
https://github.com/gohugoio/hugo/pull/3924
*Took line about HTML loading in summary and updated it to clarify the user has options to show html or not instead of implying there are no options
*Fixed typos


content/en/content-management/summaries.md
```diff
@@ -21,25 +21,26 @@ With the use of the `.Summary` [page variable][pagevariables], Hugo generates su
...
+The Hugo-defined summaries are set to use word count calculated by splitting the text by one or more consecutive whitespace characters. If you are creating content in a `CJK` language and want to use Hugo's automatic summary splitting, set `hasCJKLanguage` to `true` in your [site configuration](/getting-started/configuration/).
 {{% /note %}}
 
-### User-defined: Manual Summary Splitting
+### Manual Summary Splitting
 
-Alternatively, you may add the <code>&#60;&#33;&#45;&#45;more&#45;&#45;&#62;</code> summary divider where you want to split the article. For [org content][org], use `# more` where you want to split the article. Content that comes before the summary divider will be used as that content's summary and stored in the `.Summary` page variable with all HTML formatting intact.
+In addition, you may add the <code>&#60;&#33;&#45;&#45;more&#45;&#45;&#62;</code> summary divider where you want to split the article. For [org content][org], use `# more` where you want to split the article. Content that comes before the summary divider will be used as that content's summary and stored in the `.Summary` page variable with all HTML formatting intact.
```
rdwatters: This is an alternative and not an additional option, correct? As in, you can't do _both_ the "automatic" and "manual"  summaries for a given piece of content. Is this a gratuitous correction, @brennanangel?

brennanangel: Thank you for the feedback @rdwatters! 
I believe it's an important clarification because in my testing you are able to do both.   

You can set the default summary length at a global level and use the manual summary splitting for individual articles.   If you couldn’t do both automatic and manual, “Alternatively,” would be more clear. 

I was hoping this change would clarify that both are an option. 

Do you think it's clear that a user can do both with "Alternatively"? If so, yes, this line may be an unnecessary change.

kaushalmodi:
@brennanangel Can you rebase this PR and force push to the PR branch? Thanks.

kaushalmodi:
@brennanangel That doesn't look good :) You need to rebase, **not** merge.

kaushalmodi:
@brennanangel If rebasing doesn't work out, you might just close this PR and create a new one, with the commit based off the latest master.

PR #518, opened 2018-07-08
clarify wording

RichMorin:


kaushalmodi:
I'm not sure if this clarification is required.. the adjoining screenshot makes it pretty clear.

PR #519, opened 2018-07-08
Showcase : "Cocoon Arolla showcase"

divinerites:
@bep  I wrote a showcase about using Hugo for setting up a web site in 24h.

Hope this suits your standards. May be you'll have to edit my bad english in some place I'm afraid.

And again big thanks for your amazing job !!!


divinerites:
I don't know if my pull request is correct and if there is enough information and if it suits your standards. Do you need something more from me ?

Just let me know if i have to do something more/differently.

divinerites:
Thanks a lot for publishing it.

Issue #436, opened 2018-04-02
Incorrect post location in example


Zebradil:
Is it a mistake in the docs or I'm wrong?
I followed the steps from [quick start guide](https://gohugo.io/getting-started/quick-start/), but I couldn't see any posts. It seems like there is a mistake in [«add some content» step](https://gohugo.io/getting-started/quick-start/#step-4-add-some-content):
```
hugo new posts/my-first-post.md
```

Because if I change the directory name to `post` it works as expected.

bep:
I wrote that and tested it several times, so it should be correct. Did you test this with the Ananke theme? If not, that my explain. Other themes may have hardcoded the "post" section in their page query.

kaushalmodi:
@Zebradil And I just tested it out on the latest hugo version (again).

```
km²~/temp/:> hugo new site quickstart                                                                                       04/02 3:22pm
Congratulations! Your new Hugo site is created in /home/kmodi/temp/quickstart.

Just a few more steps and you're ready to go:

1. Download a theme into the same-named folder.
   Choose a theme from https://themes.gohugo.io/, or
   create your own with the "hugo new theme <THEMENAME>" command.
2. Perhaps you want to add some content. You can add single files
   with "hugo new <SECTIONNAME>/<FILENAME>.<FORMAT>".
3. Start the built-in live server via "hugo server".

Visit https://gohugo.io/ for quickstart guide and full documentation.
km²~/temp/:> cd quickstart/                                                                                                 04/02 3:22pm
km²~/temp/:quickstart> git init                                                                                             04/02 3:22pm
Initialized empty Git repository in /home/kmodi/temp/quickstart/.git/
km²~/temp/:quickstart> git submodule add https://github.com/budparr/gohugo-theme-ananke.git themes/ananke                   04/02 3:22pm
Cloning into '/home/kmodi/temp/quickstart/themes/ananke'...
remote: Counting objects: 919, done.
remote: Total 919 (delta 0), reused 0 (delta 0), pack-reused 919
Receiving objects: 100% (919/919), 2.42 MiB | 3.01 MiB/s, done.
Resolving deltas: 100% (482/482), done.
km²~/temp/:quickstart> echo 'theme = "ananke"' >> config.toml                                                               04/02 3:23pm
km²~/temp/:quickstart> cat config.toml                                                                                      04/02 3:23pm
baseURL = "http://example.org/"
languageCode = "en-us"
title = "My New Hugo Site"
theme = "ananke"
[...]

PR #520, opened 2018-07-11
Align new post creation with singular post form

pgilad:


kaushalmodi:
The proposed change is not needed. See https://github.com/gohugoio/hugoDocs/issues/436.

pgilad:
@kaushalmodi Then https://gohugo.io/content-management/types/#create-a-type-layout-directory should be edited as it explicitly says use the *singular* form

kaushalmodi:
@pgilad That link is about **layouts**.. `hugo new posts/foo.md` creates a file in **content**. Though.. the wording in that link you referenced might need changing. I'll open another issue for that.

PR #550, opened 2018-07-30
add note on "extended" channel for snap package

dwalkr:


anthonyfok:
See also #543

bep:
Fixed in #543

PR #554, opened 2018-08-03
Remove unused metaDataFormat setting

brennanangel:
I believe we stopped using the metadata format in config value and it now pulls from archetype files:

In https://github.com/gohugoio/hugo/issues/3706
On July 16, 2017, @bep said
"In short, you will now get the metadata format of your archetype files. See #3632 -- but we should do something with this metaDataFormat config value we're not using anymore. Not sure what."

### DIFF
--- a/content/en/getting-started/configuration.md
+++ b/content/en/getting-started/configuration.md
@@ -138,9 +138,6 @@ logFile ("")
 menu
 : See [Add Non-content Entries to a Menu](/content-management/menus/#add-non-content-entries-to-a-menu).
 
-metaDataFormat ("toml")
-: Front matter meta-data format. Valid values: `"toml"`, `"yaml"`, or `"json"`.
-
 newContentEditor ("")
 : The editor to use when creating new content.
 


blueyed:
There are more occurences:
```
docs/config.toml
8:metaDataFormat = "yaml"

create/content_test.go
83:	v.Set("metaDataFormat", "toml")

hugolib/config.go
416:	v.SetDefault("metaDataFormat", "toml")

docs/content/en/templates/introduction.md
418:Hugo provides the option of passing values to your template layer through your [site configuration][config] (i.e. for site-wide values) or through the metadata of each specific piece of content (i.e. the [front matter][]). You can define any values of any type and use them however you want in your templates, as long as the values are supported by the front matter format specified via `metaDataFormat` in your configuration file.

docs/content/en/getting-started/configuration.md
144:metaDataFormat ("toml")
```

blueyed:
Oh, sorry - I've grepped `hugo`.
Should get fixed there then also I assume?

kaushalmodi:
@bep Is metadataformat config really unused? I see it being used [here][0]:

```go
	var separator string
	if allSettings["metadataformat"] == "toml" {
		separator = " = "
	} else {
		separator = ": "
	}
```

[0]: https://github.com/gohugoio/hugo/blob/37d6463479952f7dfba59d899eed38b41e223283/commands/config.go#L56-L60

blueyed:
Yeah, that appears to control the output of `hugo config` - I've missed it in my grepping because of the casing (is case important for config keys?).

PR #579, opened 2018-08-25
Add Docs for Nested Taxonomies

onedrawingperday:
This PR is about Nesting Taxonomies under a Hugo Section. 

This undocumented feature has been available since Hugo 0.31 and in particular since commit https://github.com/gohugoio/hugo/commit/d9a78b61adefe8e1803529f4774185874af85148

Also here is a link to a repo with a Hugo test project that makes use of Nested Taxonomies:
https://github.com/onedrawingperday/hugo_nested_taxonomies

P.S. This is the first time that I am sending such an extensive PR to the Hugo Docs so feel free to add, remove or re-word whatever you think is necessary to make it better.

P.S.2 This technique for Nested Taxonomies can be used in a slightly different way to generate Blog Archives with Hugo. If this PR is accepted I will send a separate PR for the Blog Archives functionality.


onedrawingperday:
Closing because this PR since Hugo is in the process of being upgraded to Go 1.11 and conditionals using `.Scratch` will have to be refactored.

PR #582, opened 2018-08-28
Tweak 'name: weight' to 'name: date' in example

peterkappus:
Per https://github.com/gohugoio/hugoDocs/commit/19e900a17e75cc6c8202d22dafeb4257ee9d6303 I've fixed the demo yaml for the default indices to include `name: date` rather than `name: weight`. Thanks!


regisphilibert:
Thanks a ton @peterkappus for this set of eyes ;)

PR #584, opened 2018-08-30
Add Docs for Nested Taxonomies

onedrawingperday:
This PR is about Nesting Taxonomies under a Hugo Section.

This undocumented feature has been available since Hugo 0.31 and in particular since commit gohugoio/hugo@d9a78b6

Also here is a link to a repo with a Hugo test project that makes use of Nested Taxonomies:
https://github.com/onedrawingperday/hugo_nested_taxonomies

PS. This is the first time that I am sending such an extensive PR to the Hugo Docs so feel free to add, remove or re-word whatever you think is necessary to make it better.

PS.2 This technique for Nested Taxonomies can be used in a slightly different way to generate Blog Archives with Hugo. If this PR is accepted I will send a separate PR for the Blog Archives functionality.

PS.3 Despite the variable overwrites that Go 1.11 offers `.Scratch` is very much needed in this use case.

### DIFF
--- a/content/en/content-management/taxonomies.md
+++ b/content/en/content-management/taxonomies.md
@@ -205,6 +205,74 @@ If you need to add custom metadata to your taxonomy terms, you will need to crea
 
 You can later use your custom metadata as shown in the [Taxonomy Terms Templates documentation](/templates/taxonomy-templates/#displaying-custom-metadata-in-taxonomy-terms-templates).
 
+## Nested Taxonomies
+
+If you need to nest a Taxonomy under a Section of your Hugo project, you can do so with something like this:
+
+{{< code file="/config.toml" >}}
+disableKinds = ["taxonomyTerm"]
+
+[taxonomies]
+  topic = "topics"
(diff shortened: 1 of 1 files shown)


regisphilibert:
This is a great on informing people how they can achieve this. 
But as this is not a "feature" proper, I wonder if the place for this is not rather in a linked blog post or a Discourse's __Tips & Tricks__?

onedrawingperday:
@regisphilibert 
Thanks for the feedback. 

This is a simple technique that would help many people, so I thought that perhaps it belongs in the Docs. 

My currently unmaintained blog is only for my drawings (no dev articles there) and the forum **Tips and Tricks** do get buried under the support tickets.

I'll keep this PR open for a few more days,  to give time to the others to review it if they want.

Otherwise I'll eventually close it (I don't like having PRs open forever) and then I'll see how to go about it.

One idea would be to syndicate this article on your blog if you want, since you already have a series of blog posts about Hugo.

PR #594, opened 2018-09-05
Make sure `disableKinds` example is wrapped in `taxonomies` section

brandur:
The doc page has an example on how to disable default taxonomies using `disableKinds`, but it seems this key must be wrapped in the `taxonomies` section of the config or it's ignored by Hugo. This patch adds the section.

### DIFF
--- a/content/en/content-management/taxonomies.md
+++ b/content/en/content-management/taxonomies.md
@@ -97,7 +97,8 @@ Without adding a single line to your [site config][config] file, Hugo will autom
 If you do not want Hugo to create any taxonomies, set `disableKinds` in your [site config][config] to the following:
 
 {{< code-toggle copy="false" >}}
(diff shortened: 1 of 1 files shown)


brandur:
Hm, although this has the right effect in eliminating those taxonomies, I'm not sure whether that result is accidental or not. I'm seeing this line in my verbose output which is a little suspicious:

```
INFO 2018/09/05 08:08:47 found taxonomies: map[string]string{"disablekinds":""}
```

It looks like this in a default installation:

```
INFO 2018/09/05 07:54:07 found taxonomies: map[string]string{"tag":"tags", "category":"categories"}
```

I can say though that the docs as written don't seem to work.

brandur:
Nevermind. I think this is wrong — I was confused by the log message, and because the old taxonomy artifacts were not cleaned from my repo when I disabled them.

Issue #580, opened 2018-08-26
Installation instructions for OpenBSD inaccurate, package does not exist.


Seep1959:
https://gohugo.io/getting-started/installing/
> OpenBSD provides a package for Hugo via pkg_add:
> `doas pkg_add hugo`

The package does not exist in OpenBSD 6.1, 6.2, or 6.3 repos.

wondratsch:
It is in OpenBSD -current. Please wait for OBSD 6.4.

wondratsch:
OBSD 6.4 has arrived. This Issue should be closed.

PR #599, opened 2018-09-11
Updated installing.md for the OpenBSD

neerajpal09:
Changes in the file installing.md for the Hugo installation on OpenBSD.
As OpenBSD doesn't provide any package for Hugo and mage.


kaushalmodi:
The Hugo build instructions are explained in https://github.com/gohugoio/hugo#fetch-from-github. You can open a new PR with just a link pointing to that.

wondratsch:
See https://github.com/gohugoio/hugoDocs/issues/580#issuecomment-423648220

PR #603, opened 2018-09-14
Fix broken link to https://www.wercker.com

coliff:
(requires the www)


kaushalmodi:
That's a bad design; they should have redirected wercker.com to www.wercker.com.

In any case, thanks for this fix.

PR #630, opened 2018-10-07
Add draft statement to FAQ

RickCogley:
Questions re "my content is invisible" are coming up a lot in the Discourse. Add a point about draft mode.

### DIFF
--- a/content/en/troubleshooting/faq.md
+++ b/content/en/troubleshooting/faq.md
@@ -17,6 +17,10 @@ aliases: [/faq/]
 **Note:** The answers/solutions presented below are short, and may not be note be enough to solve your problem. Visit [Hugo Discourse](https://discourse.gohugo.io/) and use the search. It that does not help, start a new topic and ask your questions.
 {{% /note %}}
 
+## I can't see my content!
+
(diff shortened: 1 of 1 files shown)


bep:
Yes, this belongs there. But a note for future "I want to add to the FAQ section" people: There is a reason why the FAQ is short. It should be.

RickCogley:
Thanks


-- 
*This email message and any attachments are confidential, and are meant to 
be read by the intended recipient only. If you are not the intended 
recipient, we would request you delete this message or any attachments 
without reading or distributing, and kindly advise eSolia Inc. (contacts: 
esolia.com/about <http://esolia.com/about>). Thank you in advance.*

StephenBrown2:
But if it is in fact a frequently asked question, it should be in the FAQ. I prefer first searching a long FAQ page that gives links to further documentation than trying to dig through the docs on my own.

bep:
>But if it is in fact a frequently asked question, it should be in the FAQ. 

Yes. But I just want us to be a little bit critical to what we put there. The goal is to make people stop asking it: Maybe it is better to improve the original section about the topic.

PR #634, opened 2018-10-14
Replace javascript:void(0) docs menu toggle links

thth:
For the headings in the docs menu, instead of `javascript:void(0)`, providing links disabled in javascript to respective overview pages would:
1. retain current dropdown toggle functionality
2. provide a link that can be interacted with by non-left-click actions and interfaces, such as middle mouse click to open a new tab or the right-click context menu
3. as a positive side-effect, allow for the docs menu to be navigable by those with javascript disabled.

I also removed an unnecessary whitespace and built `app.bundle.js` with `npm run build:production` (the current iteration of the file wasn't)


thth:
To better illustrate the change, here's some images:

before | after|
---|---
 | 

(the link would be disabled by javascript, so a regular left-click would still bring out the dropdown)

PR #646, opened 2018-10-26
Attempting to make this example complete

cottrell:
Probably wrong suggestions but example is not a complete diff of a viable change AFAIK. I have attempted to include discussion points that would have enough information for the diff.


davidsneighbour:
This should be an issue, not a pull request. It's merely a list of discussion points. 
Closing this?

PR #647, opened 2018-10-28
git command to update submodule to latest

funkydan2:
This is how I update my themes. It took a while to find the commands that work, so I think it would be good to have them here for other users.

### DIFF
--- a/content/en/hosting-and-deployment/hosting-on-netlify.md
+++ b/content/en/hosting-and-deployment/hosting-on-netlify.md
@@ -128,6 +128,12 @@ You can checkout a specific version as follows:
 git checkout tags/<version-name>
 ```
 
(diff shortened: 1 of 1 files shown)


funkydan2:
Hmm, not sure why the Netlify deploy fails on this PR. The change is just a few lines of markdown.

funkydan2:
@hanzei Thanks. I've rebased my fork, but I'm not sure if the checks have been re-run.

bep:
I merged it, not sure what's wrong with Netlify. A quick note: I use `git submodule update --remote --merge` for this same thing. The difference is `merge` vs `rebase` -- it kind of depends on the workflow, but I suspect `merge` is more common, not sure ...

funkydan2:
I'm no git guru. That's the command I found on a stackoverflow discussion and it seems to work for me. Very happy for the docs to reflect a better practice.

PR #649, opened 2018-10-30
Update & Fix headers template

TomPichaud:
Allow use with latest version of Hugo.
Also Remove unnecessary line */ (trigger a bug on Netlify).


bep:
This is an old "news article" -- they are not part of the documentation, and, like the release notes, represent a snapshot in time.

PR #662, opened 2018-11-08
Add `/` to taxonomy urls

lislis:
Hey people!
Thank you for writing these excellent docs!

While c/p an example from the taxonomy page I noticed that slashes are missing from the urls.
So here they are :)

Have a great day!


bep:
Hey, and thanks for this. I'm pretty sure that `absURL` will add slashes; are you sure about this?

PR #676, opened 2018-11-30
Update Warning for Theme Links

onedrawingperday:
The advice in this page is actually the opposite. Relative URLs output the wrong URL for theme assets on the Hugo website.

See relevant advice in the Theme's repository's [README](https://github.com/gohugoio/hugoThemes#common-permalink-issues)

I hadn't noticed this page (if I had noticed it I would have sent this PR earlier).

The reason I am merging this PR right away is due to [this](https://github.com/gohugoio/hugoThemes/issues/504#issuecomment-443253163) comment by a confused theme author. 

If you feel that this could be improved feel free to reword and amend as you see fit but the bottom line is that for themes that are going to be published on Hugo Themes authors should create absolute paths in their templates.

cc: @bep @digitalcraftsman @kaushalmodi

### DIFF
--- a/content/en/themes/creating.md
+++ b/content/en/themes/creating.md
@@ -19,8 +19,8 @@ toc: true
 wip: true
 ---
 
-{{% warning "Use Relative Links" %}}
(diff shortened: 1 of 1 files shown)


bep:
If you find a bug, don't hide it in some "docs comment". Create an issue about it here:

https://github.com/gohugoio/hugo/issues

PR #680, opened 2018-12-04
Document more os.Stat fields

zwbetz-gh:
Add os.Stat examples for:
- `.Mode`
- `.ModTime`
- `.IsDir`

Remove duplicate link.

Fix grammar in param shortcode example.


bep:
Why is this needed? I say this because we have to set the line somewhere. These fields are documented elsewhere and we have a link to that documentation. That and considering that this is a specialist funciton to start with (it has some very rare and specialist use cases), that should be enough. We cannot maintain a mirror of Go's docs.

zwbetz-gh:
Fair enough. I'll close this out.

PR #682, opened 2018-12-07
Arguments in example

juh2:
The example with the abba only works if the arguments are swapped.


kaushalmodi:
@moorereason The implementation looks like a bug.

The correct syntax is `{{ strings.TrimLeft "a" "abba" }}`. But with that, we cannot do `{{ "abba" | strings.TrimLeft "a" }}`.

Should I open an issue on the hugo repo? Should this implementation be fixed?

The same swapped args apply to `strings.TrimRight` too (I haven't checked other functions).

/cc @bep

PR #692, opened 2018-12-20
The path to the first post should be under 'post' not 'posts'

bcl:


kaushalmodi:
"posts" is fine, see https://github.com/gohugoio/hugoDocs/issues/720.

PR #700, opened 2018-12-29
Remove command continuation

clemens-tolboom:
These looks confusing enough to remove them. Let's learn people one command at a time ;-)


bep:
They were added for a purpose -- which they still serves: If you want to test the "quick start" quickly, having a copy-paste friendly version is gold. There are plenty of people coming to the forum saying "the quick start does not work for me". We could possibly keep to version of this, but that would be double book-keeping and not worth it.

clemens-tolboom:
I assumed people are new to command line so they just learned this. That's why I added a new line. But I should have added another to make it work :-(

I preferred a little discussion instead of getting closed so fast :-(

clemens-tolboom:
What about:

```bash
cd quickstart

# We want to pull in the code using git so first initiallise git
git init

# Pull in the theme code
git submodule add https://github.com/budparr/gohugo-theme-ananke.git themes/ananke

# Edit your config.toml configuration file
# and add the Ananke theme.
echo 'theme = "ananke"' >> config.toml
```

PR #708, opened 2019-01-03
Link site hierarchy schema to larger version

davidsneighbour:
The image is quite small inside of the text flow. Linking it to a larger version seems accessible.


davidsneighbour:
ping. let's help people by making it visible.

PR #716, opened 2019-01-08
Add more info about complementation of date vars

salim-b:


bep:
I'm not sure that this addition makes this more clear, and I'm not a fan of having links to discussion threads in the documentation.

salim-b:
> I'm not sure that this addition makes this more clear, and I'm not a fan of having links to discussion threads in the documentation.

Well, you're probably right. I just thought I'd add this here as a kind of shortcut for future people investigating how hugo is handling `date` and `publishDate` – it's far from obvious from the current documentation I think.

I briefly thought about writing a self-contained paragraph about the complementation of date variables, but then I didn't know where to put that. Would you

1. welcome such a section in the doc at all?

2. tell me where in the doc to put it?

salim-b:
@bep Would you be so kind as to answer my open questions?

PR #722, opened 2019-01-17
Fix typo in `paginate` description

timmc:


bep:
I don't think that is a typo. It looks odd, but I think it's right.

timmc:
I *guess* I can see what you're saying—if a paginated listing is seen as a group of documents/pages, then you could call it a page of pages... but I had to read it a couple times to make sure I was understanding the setting correctly.

Maybe there's some other wording. Feel free to close or change the PR, either way.

bep:
>if a paginated listing is seen as a group of documents/pages, then you could call it a page of pages... 

Yes, that was the intention. But then I spent some brain cells about the names when I created the pagination API (there are plenty of implementations floating around, but the names used varies wildly). 

That said,

* I see that I use element myself in the method names (`NumberOfElements`)
* A paginated page can contain either a list of either `Page` or `PagesGroup`, and we may get others in the future.

So, you are right -- but it wasn't a typo, it was just me who could not decide ...

PR #727, opened 2019-01-20
update the help output to the current 0.53 version

CoolGoose:
This is especially useful when you're searching for the html minify that's from the 0.47 version onward.


bep:
Sorry for the delay, but I think I have updated this after the recent additions?

CoolGoose:
@bep yup, seems updated. Can be closed.

PR #731, opened 2019-01-27
Improve Algolia Search Display Styling

budparr:
This commit restores original styling to the Algolia search box and adds a Hugo color to headings.


rdwatters:
Regrettably, this seems to break viewport on mobile. Tested on iPhone 8.

PR #746, opened 2019-02-24
Move 2nd part of step 3 to it's own code block

hoguej:
I missed this step the first time I went through the tutorial because it was at the bottom of a large comment block. Starting a new block may make it more obvious for other newbs.


prophoss:
I would like to add that I had the same trouble, and when I typed echo 'theme = "ananke"'>> config.toml in the powershell treminal, using vscode,  it did not add "theme = ananke" correctly to the config.toml file. 
instead it added t�h�e�m�e� �=� �"�a�n�a�n�k�e�"�

�

�

davidsneighbour:
You are adding an empty code block to achieve a documentation layout. That's not solving the issue at hand. Please open a issue to fix this.

PR #747, opened 2019-02-28
improve dates i18n workaround

gdeshors:
The existing one worked only when you had one language. This proposal works with several languages.

### DIFF
--- a/content/en/content-management/multilingual.md
+++ b/content/en/content-management/multilingual.md
@@ -352,32 +352,37 @@ And then in the template:
 
 ## Customize Dates
 
-At the time of this writing, Go does not yet have support for internationalized locales, but if you do some work, you can simulate it. For example, if you want to use French month names, you can add a data file like ``data/mois.yaml`` with this content:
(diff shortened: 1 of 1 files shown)


gdeshors:
(in reply to the pull request being stalled) I like my proposal because I spent quite some time figuring out how to do it for myself, hence I'd like to share it with future readers.

davidsneighbour:
This pull request is a solution that better lives in [the Tips and Tricks section](https://discourse.gohugo.io/c/tips-tricks/) of the Hugo Forum. You replace a one-language solution with a multi-language solution and by the way remove the `time` tag.

gdeshors:
You're right about me removing the time tag, this is an error, thanks for pointing it out. Do I have a way to edit my pull request ?
Apart from that, a multi-language solution still seems better to me than a one-language one, and if it was my need, I may not be alone.

davidsneighbour:
Any commits to the branch you merge in this pull request will be added to the pull request.

PR #751, opened 2019-03-07
Updated Hugo convert to clarify for and not ..

tim-bradley:
.. for configuration files.


kaushalmodi:
Looks good. Only the CLA needs to be signed.

PR #752, opened 2019-03-10
Correct Google Analytics config key casing

phanan:
The current config key `googleAnalytics` doesn't work (at least for me). `GoogleAnalytics` (with a capital G) works fine.


davidsneighbour:
Please add more information. To me this seems to be more an issue than worthy a pull request. `googleAnalytics` works well for me in an environment where I am not using a specific theme. I think your issue might be theme related, not a Hugo issue.

bep:
The naming standard used in documentation/examples are camel case (staring with lower case).

PR #757, opened 2019-03-13
Clarify: summaryLength uses words not characters

Phlow:
I had to look up if characters or words… Just a little improvement to the docs.


bep:
I think it's actually runes (UTF8 characters) for CJK languages, but your edit it probably the best compromise for the docs...

Phlow:
Thanks @bep – Greets to Sweden!

bep:
I'm from Norway, but it's close ...

PR #761, opened 2019-03-14
update link to external mmark documentation

Trewaters:
I am learning to use Hugo and I saw the original repo for mmark was deprecated. I thought it might be helpful if you had the updated link since it seems to point to an old location. 

Hope this helps. I am enjoying the docs so far.


bep:
> I thought it might be helpful if you had the updated link since it seems to point to an old location.

Thanks.

The old location may be old, but it is the correct location re. the version Hugo uses.

PR #770, opened 2019-03-24
Describe how to set a `datetime` with a post's `date`

jamietanna:
This doesn't seem to be documented anywhere explicitly aside from in
posts such as https://discourse.gohugo.io/t/clarity-on-front-matter-date-format/1794

Changes can be seen at https://deploy-preview-770--gohugoio.netlify.com/content-management/front-matter/


jamietanna:
This is quite useful for folks to know, so I'd appreciate if this could get merged!

davidsneighbour:
This commit does not explain _HOW_ to set to an ISO 8601 string. The documentation as of now says quite right that it's set to the frontmatter type's date format?

PR #774, opened 2019-03-26
Fix search box text being invisible in Safari

tinymachine:
Here's how the search box is rendered in Safari (12.0.3 on macOS and on iOS 12.1.4), with white text ('test') on a white background:

Screen shot showing white text 'test' on white search box background

This is virtually unusable.

For some reason (Purgecss misfire?) the CSS class `.bg-transparent` does not appear in the built CSS file, so the assignment of the class to the `input` element is meaningless.

The inline style declared on the `input` element using the `background` shortcut declaration lacks a background-color. Chrome and Firefox interpret that as a transparent background, but Safari just ignores it, falling back to its default white background. This means text typed into the search field renders as white on white, effectively invisible. 

This PR explicitly declares a transparent background color so the search box should render correctly in all three browsers (white text on black background).

### DIFF
--- a/themes/gohugoioTheme/layouts/partials/site-search.html
+++ b/themes/gohugoioTheme/layouts/partials/site-search.html
@@ -1,6 +1,6 @@
 <form id="site-search-form" action="" role="search">
   <fieldset class="bn ma0 pa0">
     <label class="clip" for="search-input">Search</label>
-    <input type="search" id="search-input" class="needs-js bg-left bg-transparent bn f5 input-reset lh-solid mt3 mt0-ns pl4 pv2 w5 white" placeholder="Search the Docs" name="search-input" value="" style="background: url('/images/icon-search.png') no-repeat 0 8px /16px 16px;;">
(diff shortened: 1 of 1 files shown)


bep:
I will check it on the docs site then merge it to the theme if it works. Thanks!

tinymachine:
You're welcome!

>

budparr:
I removed purge, for now, because it's proven to be problematic.

tinymachine:
> I removed purge, for now, because it's proven to be problematic.

@budparr For what it's worth, from what I can tell I don't think the issue here (`.bg-transparent` not appearing in the CSS even though the class was present on the `input` element) is the result of a purge bug. Looking at the commit history, it looks like `.bg-transparent` has been added and removed a bit from that `input`, and that might be the only place in the whole codebase where that class was present, so if it wasn't there the last time the CSS was built, it would have (appropriately) been purged.

`.bg-transparent`was added back in with 49809a038b2691637bab7f3f2e385dde654a88b8 on Jan 31, but it looks like, until today, app.css hadn't been regenerated since Jan 27 with e6777153b5aea72cbe7094501d3c1fe0eb93615c ([see app.css's history](https://github.com/gohugoio/hugoDocs/commits/master/themes/gohugoioTheme/assets/output/css/app.css)). 

I might be totally off, but maybe the issue with this file was more of a workflow issue? (Perhaps running the build script could be tied to a git commit hook or something? 🤔)

PR #775, opened 2019-03-26
Actuall show how to generate pages in quickstart

baryluk:
The development server isn't really that important for quick start.

Yes it is useful but makes things complex.

The most important part is really static site generation, and it is actually hard to find how to do it in your tutorials!

Same for `hugo new`. It is automagic, and anti-thesis of things being simple. This is why I added section to make it less magic, and simply allow content to be done manually.


bep:
@baryluk can you push the "CLA sign" button.

baryluk:
> @baryluk can you push the "CLA sign" button.

Done

PR #779, opened 2019-04-03
wrong use of git parameters to install all themes

nonumeros:
correcting wrong use of git parameters to clone all themes

there's no such thing as `depth` but `depth=` and neither `--recursive` but `--recurse-submodules`  

read https://git-scm.com/docs/git-clone
read https://git-scm.com/book/en/v2/Git-Tools-Submodules


nonumeros:
moved config back to _default didn't realize it

PR #780, opened 2019-04-03
wrong use of git parameters to install all themes

nonumeros:
correcting wrong use of git parameters to clone all themes

there's no such thing as `--depth` but `--depth=` and neither `--recursive` but `--recurse-submodules`  

read https://git-scm.com/docs/git-clone
read https://git-scm.com/book/en/v2/Git-Tools-Submodules


nonumeros:
What's the problem, if any, with this pull request?

If you're not going to merge it, at least fix the incorrect usage of git parameters on the documentation.

the only relevant part of the documentation dates as far back as when @spf13  was involved.

PR #791, opened 2019-04-14
Update summaries.md

jcheng1203:
Clarified that you need to call .Params.summary to access a summary defined in front matter, and that calling .Summary won't work.


bep:
>and that calling .Summary won't work.

It does, provided you don't have a summary delitmter.

jcheng1203:
Hmm well it's not working for me...I just created a simple test site with just an _index.md:
code

And here's how it renders:
output

Note how {{ .Summary }} pulls the summary from the actual content, even with no summary delimiter in the markdown file. Am I missing something here?

FYI I'm using Hugo Static Site Generator v0.53/extended windows/amd64 BuildDate: unknown. 
Also, the _default/baseof.html is empty, and the config file looks like this:
config

bep:
I'm going to close this. This works in my tests. If it, for some odd reason, does not work, it's a bug and need to be createdd as an issue in `hugo` repo.

PR #795, opened 2019-04-19
add warning about .Hugo variable being deprecated

inwardmovement:
[#5835](https://github.com/gohugoio/hugo/issues/5835)

### DIFF
--- a/content/en/variables/hugo.md
+++ b/content/en/variables/hugo.md
@@ -19,6 +19,11 @@ toc: false
 wip: false
 ---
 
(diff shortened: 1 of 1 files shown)


danieljsummers:
I'll 2nd this one; finding this helped me get rid of the warning.

_(Assuming the `hugo` API is the same as the `.Hugo` variable, this is a really good interim note. I'd think that future edits could rework the entire page to reference `hugo` instead of `.Hugo`. I also wouldn't mind making them, if it's just translation work.)_

PR #798, opened 2019-04-24
Add `preserveTaxonomyNames` as a documented option

jasikpark:
This is an important feature for myself that I only found at https://www.jakewiesler.com/blog/hugo-taxonomies/ since I want my tags to be lowercase or upper case if I want them to be. 

Thanks!!


kaushalmodi:
Hello, thanks for the PR. Though, `preserveTaxonomyNames` has been removed.

See my comment in the forum: https://discourse.gohugo.io/t/do-not-titlecase-tags-in-taxonomy-pages/11815/15?u=kaushalmodi

So, closing this PR.

PR #802, opened 2019-04-25
Add showcase 12bit.vn

tatthien:


bep:
Thanks for this.

Your site looks great, but I think that personal blogs fall outside the scope of the showcase. I wouldn't consider adding my own blow to the showcase, even if I probably could have written something interesting about it.

Again, thanks, but I will close this now.

tatthien:
Thank you @bep

PR #807, opened 2019-04-27
no -> fr

danieljsummers:
consistency in examples (I suspect it used to use Norse, but now uses French)


bep:
Norse isn't a language code. `no` is Norwegian (bokmål),  which I suspect comes from my pen. Why does the example need to be ... French?

danieljsummers:
I don't suppose the example needs to be French, but it's French throughout the document except in this one spot, where it currently says that configuring your site for French will produce a `no` directory. :)

PR #811, opened 2019-05-02
Added Pharmaseal case study

jonoroboto:
Added PHARMASEAL case study to the showcase pages. Included some relevant information about the tech stack as well as why we've built it this way and how we aim to disrupt the pharmaceutical CTMS market.

If you would like anything adding to it, feel free to message jono@roboto.studio as we can discuss in depth the tech used and the process for building.


bep:
This looks great! (so does the site itself!) Thanks.

PR #815, opened 2019-05-07
Added related sha function into the See Also

biodranik:


content/en/functions/md5.md
```diff
@@ -14,7 +14,7 @@ keywords: []
 signature: ["md5 INPUT"]
 workson: []
 hugoversion:
-relatedfuncs: []
+relatedfuncs: [sha]
```
chalin, line 17: @bep - Because of this change, is `sha` supposed to show up in the "See Also" section of the [md5](https://gohugo.io/functions/md5/) page?

If so, it isn't there; in fact, there is not "See Also" section.

If not, what effect is this supposed to have on the docs?

(I'm asking because I'd like to submit an addition to the "See Also" section for another doc page, but I'm unsure how to do so. I'm reading up on it, but figured I ask the questions above none-the-less.)

PR #817, opened 2019-05-11
hugo-extended no longer the default

yarko:
It appears hugo-extended is no longer the default build.  I'm on Linux / Ubuntu, so would appreciate if someone check that it is also no longer the default via `brew install` (I assumed the change was consistently throughout, based on github changes).


bep:
> check that it is also no longer the default via brew install 

Do you use brew on Ubuntu?

yarko:
Sorry for delay (traveling).
No - didn't even know "brew" was a thing on Ubuntu.... Just `git pull`, and `go install` which lead to finding that (for Academic theme, for example) I needed to use `go install --tags extended`.

peaceiris:
@yarko @bep 

I use Hugo via `brew` on macOS. The Hugo seems to be extended version by default.

```
$ hugo version
Hugo Static Site Generator v0.55.5/extended darwin/amd64 BuildDate: unknown
```

PR #825, opened 2019-05-24
Update output-formats.md

lpar:
"suffix" is now "suffixes" and takes an array.

>Error: MediaType.Suffix is removed. Before Hugo 0.44 this was used both to set a custom file suffix and as way to augment the mediatype definition (what you see after the "+", e.g. "image/svg+xml").
>
>This had its limitations. For one, it was only possible with one file extension per MIME type.
>
>Now you can specify multiple file suffixes using "suffixes", but you need to specify the full MIME type
identifier:


bep:
Sorry for the delay, the changes look good, but could you push the CLA sign button above?

PR #830, opened 2019-06-01
Adding required XML tag to sample templates

davidsneighbour:
Since Hugo 0.55.0 the XML declaration on XML and RSS files is not added automatically anymore. The samples should reflect that.


davidsneighbour:
@bep you might have overlooked this one. I added two similar pull requests. one for the sitemaps doc (merged yesterday) and one for RSS templates (this one). Should have put those into one single pull request, but Github somehow made me create two.

PR #837, opened 2019-06-10
Correct typo "RFC339" to "RFC1123Z"

genderquery:
I believe the example under the section "Hugo Date and Time Templating Reference" should be "[RFC3339](https://tools.ietf.org/html/rfc3339)", which is "Date and Time on the Internet: Timestamps" instead of "[RFC339](https://tools.ietf.org/html/rfc339)", which is "MLTNET - A 'MULTI-TELNET' SUBSYSTEM FOR TENEX".

### DIFF
--- a/content/en/functions/format.md
+++ b/content/en/functions/format.md
@@ -87,7 +87,7 @@ date: 2017-03-03T14:15:59-06:00
 `"Mon, 02 Jan 2006 15:04:05 MST"` (RFC1123)
 : **Returns**: `Fri, 03 Mar 2017 14:15:59 CST`
 
-`"Mon, 02 Jan 2006 15:04:05 -0700"` (RFC339)
+`"Mon, 02 Jan 2006 15:04:05 -0700"` (RFC3339)
 : **Returns**: `Fri, 03 Mar 2017 14:15:59 -0600`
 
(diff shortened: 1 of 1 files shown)


kaushalmodi:
Interestingly while that could be a typo, RFC3339 does not represent date/time like that.

See: https://tools.ietf.org/html/rfc3339#section-5.8

Here's a sample RFC3339 date/time stamp (the one Hugo uses in the Time type front-matter: *1937-01-01T12:00:27.87+00:20*.


May be that reference to RFC339/3339 should be taken out altogether from that example. @bep?

genderquery:
Ah, you are correct. It appears the format is supposed to be RFC1123Z. RFC3339 would be `"2006-01-02T15:04:05Z07:00"`. The Go documentation for the `time` package lists constants for different common formats: https://golang.org/pkg/time/#ANSIC. Perhaps they should be referenced, instead?

kaushalmodi:
> The Go documentation for the time package lists constants for different common formats: https://golang.org/pkg/time/#ANSIC. Perhaps they should be referenced, instead?

Yes please. Can you make the necessary edits?

kaushalmodi:
Thanks!

PR #842, opened 2019-06-17
add docs for page variable:  .IsSection

ianre657:
This page variable can be found at:
[page variable](https://github.com/gohugoio/hugo/blob/75467cd7852852305549a6c71ac503bb4a57e716/resources/page/page.go#L173)

### DIFF
--- a/content/en/variables/page.md
+++ b/content/en/variables/page.md
@@ -72,6 +72,9 @@ See [`.Scratch`](/functions/scratch/) for page-scoped, writable variables.
 .IsPage
 : always `true` for regular content pages.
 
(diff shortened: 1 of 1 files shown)


bep:
Thanks, but I don't think your definition is correct.

`IsSection` returns true when a `Page` is of kind `section` (which may or may not have an `_index.md` file).

ianre657:
I'm a newbie. 
can't see what would be the case that there is a `Page` with kind `section` but is not a file with `_index.md`.
How about using your definition above and with some examples attached for other developers.
It will be more comprehensive, I believe.

PR #845, opened 2019-06-18
Use actual width in image processing examples

### LINKED ISSUE #812: Make the image processing images non-responsive
https://gohugo.io/content-management/image-processing/

They are very small (90x120 etc.)  and when stretched to the full width of the container, they look very ugly.

rhcarvalho:
The image processing examples are rendered within containers with a
'nested-img' parent CSS class, which by default makes images responsive
by resizing them to the size of their containers.

That doesn't work well for the examples as the small images end up
stretched.

Since there can be other images in the same 'nested-img' context, we
keep that class in the parent and set the example images width to auto,
such that they will render with their actual size.

On small screens, larger images will still be sized down via the
max-width property.

## Testing

Tested manually in Chrome 75.0.3770.90 and Firefox 67.0

Affected URLs:
- https://gohugo.io/content-management/image-processing/
- https://gohugo.io/about/new-in-032/
- https://gohugo.io/content-management/organization/
- https://gohugo.io/news/lets-celebrate-hugos-5th-birthday/

### Before

image

### After

image

---

Fixes #812

### DIFF
--- a/layouts/shortcodes/imgproc.html
+++ b/layouts/shortcodes/imgproc.html
@@ -12,7 +12,7 @@
 {{ end }}
 {{ $image := .Scratch.Get "image" }}
 <figure style="padding: 0.25rem; margin: 2rem 0; background-color: #cccc">
-	<img style="max-width: 100%; height: auto;" src="{{ $image.RelPermalink }}" width="{{ $image.Width }}" height="{{ $image.Height }}">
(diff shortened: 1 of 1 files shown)


rhcarvalho:
I have considered also resizing the `figure` container to avoid the extra gray area, but making it fit the images would cause unnecessary line wrapping on the captions:

image

bep:
Thanks a lot -- it has been on my list of things to fix for a little too long...

PR #854, opened 2019-07-09
Fix statement about version installed with apt-get

alexearnshaw:
When I use `sudo apt-get install hugo` on Ubuntu this installs the non-extended version, not the extended version as this page states. After installing this way the version I have is:
```
hugo version
Hugo Static Site Generator v0.40.1 linux/amd64 BuildDate: 2018-04-25T17:16:11Z
```

### DIFF
--- a/content/en/getting-started/installing.md
+++ b/content/en/getting-started/installing.md
@@ -456,7 +456,7 @@ Hugo installed via Snap can write only inside the user’s `$HOME` directory---a
 
     sudo apt-get install hugo
 
-This installs the "extended" Sass/SCSS version.
+This installs the non-extended version without Sass/SCSS support.
 
 ### Arch Linux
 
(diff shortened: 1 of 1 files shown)


martignoni:
@alexearnshaw I'm afraid this is only correct on Ubuntu Bionic.

On Ubuntu disco (Ubunto 19.04), on the contrary, we get:

```bash
nicolas@xxxx:~# uname -a
Linux xxxx 5.0.0-25-generic #26-Ubuntu SMP Thu Aug 1 12:04:58 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
nicolas@xxxx:~# hugo version
Hugo Static Site Generator v0.52/extended linux/amd64 BuildDate: 2018-12-21T15:59:29Z
```

If amending the doc is needed, I suggest your change your PR to mention the two cases.

alexearnshaw:
@martignoni Thanks I wasn't aware of that. I've updated the PR as suggested.

PR #857, opened 2019-07-13
Update Forestry CMS features

DirtyF:
😗


bep:
Good thinking to put Hugo first in that list ...

PR #865, opened 2019-07-26
Fix .RSSLink typo in Release Notes

coliff:


bep:
Thanks, but that is a typo in a commit message. We might fix it, but I should have done that when I released, no we need to edit in two places, and then it should be more important than a typo.

PR #867, opened 2019-07-26
HTTPS link to www.oreilly.com

coliff:
safer and faster


kaushalmodi:
@bep the deploy preview is failing for a different reason, so merging this.

PR #883, opened 2019-08-08
Mention that drafts are not deployed by default

nakamin:
I was following the tutorial, and was a bit confused as to why there were no posts in the production version of the site.

### DIFF
--- a/content/en/getting-started/quick-start.md
+++ b/content/en/getting-started/quick-start.md
@@ -182,5 +182,5 @@ hugo
 Output will be in `./public/` directory by default (`-d`/`--destination` flag to change it, or set `publishdir` in the config file).
 
 {{% note %}}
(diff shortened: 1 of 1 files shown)


kaushalmodi:
Can you remove the literal "Note:" text in bold? It's already inside a note shortcode.

content/en/getting-started/quick-start.md
```diff
@@ -180,3 +180,7 @@ hugo
 ```
 
 Output will be in `./public/` directory by default (`-d`/`--destination` flag to change it, or set `publishdir` in the config file).
+
+{{% note %}}
+Drafts do not get deployed; once you finish a post, update the header of the post to say `draft: false`. More info [here](/getting-started/usage/#draft-future-and-expired-content)
```
kaushalmodi: Now it's just missing the ending period/fullstop:

```diff
- More info [here](/getting-started/usage/#draft-future-and-expired-content)
+ More info [here](/getting-started/usage/#draft-future-and-expired-content).
```

PR #886, opened 2019-08-12
Update data-templates.md

XhmikosR:
Fix anchor to `#call-the-functions-with-a-url`

### DIFF
--- a/content/en/templates/data-templates.md
+++ b/content/en/templates/data-templates.md
@@ -217,7 +217,7 @@ Currently, you can only use those authentication methods that can be put into an
 
 To load local files with `getJSON` and `getCSV`, the source files must reside within Hugo's working directory. The file extension does not matter, but the content does.
 
-It applies the same output logic as above in [Calling the Functions with a URL](#calling-the-functions-with-a-url).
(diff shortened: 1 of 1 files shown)


XhmikosR:
BTW wouldn't this be better if there was a shortcode or something so that build would fail?

kaushalmodi:
> BTW wouldn't this be better if there was a shortcode or something so that build would fail?

I think only running a link checker would help here. We do not have link checking enabled in the CI yet. May be this deserves a discussion in a separate hugoDocs issue.

XhmikosR:
I was referring to using `ref``relref` to link to internal pages/anchors.

kaushalmodi:
I tried that but that doesn't error out on broken anchors.

XhmikosR:
True. Not sure if this could be implemented.

XhmikosR:
BTW you could give [linkinator](https://github.com/JustinBeckwith/linkinator) a go. We are using it on Bootstrap, it's pretty fast, but I doubt it checks anchors either.

kaushalmodi:
I have been using https://github.com/wjdp/htmltest (again, don't think that checks anchors either).

XhmikosR:
It seems htmltest does support anchors

>a: Whether internal hashes work.

PR #898, opened 2019-08-26
fix languageCode format with https://www.ietf.org/rfc/rfc4646.txt

h0wardch3ng:
fix languageCode format with https://www.ietf.org/rfc/rfc4646.txt


davidsneighbour:
I believe that Hugo is not using rfc4646 for language codes. I recently read somewhere that @bep wrote that the language code can contain now "any string". I'll try to find where I read this, but I would suggest to close this.

PR #901, opened 2019-08-28
clarify config directory structure. fixes #893

### LINKED ISSUE #893: Directory structure for config.toml
On https://github.com/gohugoio/hugoDocs/blob/master/content/en/getting-started/configuration.md it displays correctly:

image

But on https://gohugo.io/getting-started/configuration/#configuration-directory the last (production) part is wrong:

image

bmackinney:
I ran `tree examplesite` from my theme and then deleted/renamed config files. 
image

Entries for `menus.en.toml` revised to `menu.en.toml` to match config docs.

### DIFF
--- a/content/en/getting-started/configuration.md
+++ b/content/en/getting-started/configuration.md
@@ -54,10 +54,9 @@ In addition to using a single site config file, one can use the `configDir` dire
 │   ├── _default
 │   │   ├── config.toml
 │   │   ├── languages.toml
-│   │   ├── mediaTypes.toml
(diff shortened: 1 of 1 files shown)


bmackinney:
Related to the change I made and then removed, `menus.en.toml`->`menu.en.toml`, I need to test and then will open a separate issue based on what I find.

bep:
>menus.en.toml->menu.en.toml

There is an alias so both works, but we decided to make the "plural one" the official variant.

PR #907, opened 2019-09-01
Migrations.md update: Another tool for migration from Medium platform

RaVbaker:
I discovered another tool for migration from Medium to Hugo which his one stop shop. Written in Go, so available as a single binary on any platform and very fast + also imports appropriate tags and images from original posts. 

I tested it on my blog and worked flawlessly. Highly recommended!


lfbittencourt:
Can we accept this one? It's a very nice addition.

Ping @nantipov

nantipov:
@lfbittencourt Yes, looks good.
@bep Sorry for disturbing, what do you think if it would be possible to go through the open pull-requests with simple documentation changes (like this one) and accept/reject them?
I would be glad to help, but do not have write access.

PR #913, opened 2019-09-07
Borrow slogan from the inter web 😄

DirtyF:
Originally posted by @bep on TND Slack:

image


bep:
I assume you know where I got it from? :-)

PR #915, opened 2019-09-13
Add image actions

bep:
See https://github.com/gohugoio/hugoThemesSite/issues/64

### DIFF
--- a/.github/calibre/image-actions.yml
+++ b/.github/calibre/image-actions.yml
@@ -1,0 +1,6 @@
+jpeg:
+  quality: 80
+png:
+  quality: 80
+ignorePaths:
+  - "resources/**"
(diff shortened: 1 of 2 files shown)


coliff:
Nice!  I'd suggest **losslessly** optimizing the images though (the default method). I think setting 'quality' to 80 will make some images look worse.

bep:
OK, I guess I need to merge this first before anything can ...happen.

bep:
Not sure what you mean, I just copied the default config and changed the ignore path:

https://github.com/marketplace/actions/image-actions#configuration

bep:
I will do a test PR and see what happens ...

coliff:
Yeah I think removing this part would be best:
```
jpeg:
  quality: 80
png:
  quality: 80
```
That way the config will be just set to ignore that path you added.

PR #916, opened 2019-09-13
Test image opt

bep:


bep:
I get an error, yet it succeeds:

```bash
(node:1) UnhandledPromiseRejectionWarning: HttpError: Resource not accessible by integration
    at response.text.then.message (/usr/local/src/image-actions/node_modules/@octokit/request/lib/request.js:56:27)
    at processTicksAndRejections (internal/process/task_queues.js:86:5)
(node:1) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:1) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
 Complete job
```

/cc @coliff I will try to read up on the docs later, I assume there is some permission issue somewhere.

coliff:
I wonder if you need to sign up for the beta for it to work... https://github.com/features/actions

bep:
I have and we're already in.

bep:
The same setup works on one of my test sites (https://github.com/bep/hugotest/pull/3), so I assume it's something with this org.

bep:
I'll try another PR, I'm guessing there is a timing issue between this PR and me adding the workflow files.

coliff:
ah, I think it only runs on PRs with new/modified images - so it doesn't run against all the images in the existing repo. Maybe I should just open a PR with the optimised images and then hopefully any newly added images will be taken care of the image-actions you've setup already.

bep:
> so it doesn't run against all the images in the existing repo. 

It does. See the exact same PR in my account.

coliff:
oh right. The error in the GitHub Action looks like the same error as this here:
https://github.com/calibreapp/image-actions/issues/13

PR #918, opened 2019-09-13
Test image action

bep:


bep:
@coliff I've sent a support ticket to GitHub about this. This works with the same repo and setup on my account, but not here. I checked that both have Action beta access. I assume it's a permission thing, but I fail to see what.

PR #931, opened 2019-10-03
Fleshed out explanations in rsync deployment

napcs:
Added some additional explanations and split up a couple command blocks to make things more clear to people less experienced with this kind of deployment.


bep:
This is great; thank!

napcs:
@bep wonderful! I'm in the process of writing a Hugo book for PragProg and it's almost done - want to do my part to give back to Hugo's official docs, so I'll keep going. Thanks for accepting.

PR #954, opened 2019-10-23
Update 404.md

gfrcsd:
Update link to Netlify docs regarding 404 redirection and explain how it is now automatically handled


davidsneighbour:
While you are at it, wouldn't you want to add the info or a note that setup for specific needs is available via _headers file (translations, people who don't want 404.html in the root)?

Issue #862, opened 2019-07-21
quickstart guide: final step doesn't produce static files for post, because post is still draft


liasomething:
As per title: the quick start guide https://gohugo.io/getting-started/quick-start/ generates static html with no posts, which is not useful for users.

To fix: either 

1. in Step 7, hugo needs to be invoked as `hugo --buildDrafts`, or
2. in Step 4, the user should be told to set `draft` to `false`

clemens-tolboom:
That will not help as the suggested theme *Anaka* uses **/post/**. See https://github.com/budparr/gohugo-theme-ananke/tree/master/exampleSite/content

The **asciicast** is blocking me to create a PR.

rdela:
Hi Clemens, thanks for commenting, but I think that is a different issue than the one LIA opened this issue about, that drafts did not get built in the old version of the quick start.

clemens-tolboom:
I did the Quickstart again today and did not get any content with `hugo --buildDrafts` until I changed the directory name from `posts` into `post` (which anaka theme uses) so even your ref to #883 was not enough.

PR #958, opened 2019-10-29
Anaka theme uses /post/ and NOT /posts/

clemens-tolboom:
This should solve #862 

Not sure about the **asciicast** though :-(

### DIFF
--- a/content/en/getting-started/quick-start.md
+++ b/content/en/getting-started/quick-start.md
@@ -89,7 +89,7 @@ echo 'theme = "ananke"' >> config.toml
 You can manually create content files (for example as `content/<CATEGORY>/<FILE>.<FORMAT>`) and provide metadata in them, however you can use the `new` command to do few things for you (like add title and date):
 
 ```
-hugo new posts/my-first-post.md
+hugo new post/my-first-post.md
 ```
 
 {{< asciicast eUojYCfRTZvkEiqc52fUsJRBR >}}


davidsneighbour:
This is the documentation for Hugo, not for the Anaka theme. Please re-open where ever this applies to.

clemens-tolboom:
Ehm ... https://gohugo.io/getting-started/quick-start/#step-3-add-a-theme **adds** the theme then the page will never be displayed

davidsneighbour:
Hmm, so the consensus seems to be that the folder should be the plural version of the post type but both versions (singular/plural) work. I would keep it consistent and maybe change the used sample theme for this or change it in Anaka?

@bep?

clemens-tolboom:
The themes I checked (trying to use) did not have *concensus* neither on structure nor plural form :-(

- https://github.com/budparr/gohugo-theme-ananke/tree/master/exampleSite/content
- https://github.com/victoriadrake/hugo-theme-introduction/tree/master/exampleSite/content/en
- https://github.com/digitalcraftsman/hugo-artists-theme/tree/master/exampleSite
- https://github.com/okkur/syna/tree/master/exampleSite/content

So I guess we should allow this PR

BUT what about this **asciicast** ... can we get rid of that or is that auto generated?

budparr:
When Hugo was created I recall the docs being specific about the folder being singular. I'll see if I can find that.

clemens-tolboom:
@budparr would changing anaka not impact all anaka users in an unsuspected way (renaming folder(s)?

This PR is only to make the quickstart work properly ;-)

budparr:
I don't think it would be a good idea to change Ananke because there are a fair number of users out there.

I can't find much of a trace of Steve's way of naming folders. It was formerly singular (as I recall), but now I see people are using plural (some changes from this year). Personally I think `site.com/post/post-name/` makes the most sense.

At the end of the day I think the docs outside of this particular how-to should be unopinionated about folder naming conventions, so it seems reasonable that themes would handle it differently

bep:
So, what the main section(s) is named in a theme will wary; post/posts/blog/foo... So if you're doing a `where .Site.RegularPages ...` on the home page it's recommended to query by the `mainSections` param:

https://gohugo.io/functions/where/#mainsections

A user can set this in the config if he/she wants, but Hugo will do some page counting to determine what's considered "the main".

bep:
So, I will not merge this. #862 is about something else, I think.

clemens-tolboom:
I guess something went wrong ... the quick start does not funciton as **/posts/** are not 'rendered'

I should have filed a better summary :-( and refer to mentioned issue. My bad.

- https://gohugo.io/getting-started/quick-start/#step-4-add-some-content suggests to run `hugo new posts/my-first-post.md`
- https://gohugo.io/getting-started/quick-start/#step-5-start-the-hugo-server serves `anaka` based site which does not support `/posts/`

Thus this PR.

PR #970, opened 2019-11-26
Update installing.md

luucamay:
observing that the Linux package for Hugo is out of date

### DIFF
--- a/content/en/getting-started/installing.md
+++ b/content/en/getting-started/installing.md
@@ -463,7 +463,7 @@ Hugo installed via Snap can write only inside the user’s `$HOME` directory---a
 
 This installs the "extended" Sass/SCSS version.
 
-This option is not reccommended because the Hugo in Linux package managers for Debian and Ubuntu is out of date. This issue is described [here](https://github.com/gcushen/hugo-academic/issues/703)
+This option is not recommended because the Hugo in Linux package managers for Debian and Ubuntu is usually a few versions behind as described [here](https://github.com/gcushen/hugo-academic/issues/703)
 
 ### Arch Linux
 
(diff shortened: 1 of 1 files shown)


content/en/getting-started/installing.md
```diff
@@ -463,6 +463,8 @@ Hugo installed via Snap can write only inside the user’s `$HOME` directory---a
 
 This installs the "extended" Sass/SCSS version.
 
+This option is not reccommended because the Hugo in Linux package managers for Debian and Ubuntu is broken. This issue is described [here](https://github.com/gcushen/hugo-academic/issues/703)
```
bep: Please use "... Ubuntu is out of date". It's not broken.

content/en/getting-started/installing.md
```diff
@@ -463,7 +463,7 @@ Hugo installed via Snap can write only inside the user’s `$HOME` directory---a
 
 This installs the "extended" Sass/SCSS version.
 
-This option is not reccommended because the Hugo in Linux package managers for Debian and Ubuntu is broken. This issue is described [here](https://github.com/gcushen/hugo-academic/issues/703)
+This option is not reccommended because the Hugo in Linux package managers for Debian and Ubuntu is out of date. This issue is described [here](https://github.com/gcushen/hugo-academic/issues/703)
```
vsopvsop: This option is not reccommended because the Hugo in Linux package managers for Debian and Ubuntu is usually few versions behind.

vsopvsop: This option is not recommended because the Hugo in Linux package managers for Debian and Ubuntu is usually a few versions behind.

luucamay: Thanks for the suggestion, I already changed the file like that

PR #973, opened 2019-12-07
Fix link in rss.md

TixieBorg:
Hi,
There is currently a problem with the html rendering. 

A doc link in `bla bla bla (there is a http://example.com).` would redirect to ` http://example.com)`.

I saw this in https://gohugo.io/templates/rss/#section-rss 

So this just a little fix while I don't know how to fix the md->html rendering in itself.

### DIFF
--- a/content/en/templates/rss.md
+++ b/content/en/templates/rss.md
@@ -29,7 +29,7 @@ RSS pages are of the type `Page` and have all the [page variables](/variables/pa
 
 ### Section RSS
 
-A [section’s][section] RSS will be rendered at `/<SECTION>/index.xml` (e.g., https://spf13.com/project/index.xml).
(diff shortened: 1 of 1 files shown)


bep:
This is fixed in Goldmark (which Hugo uses), I will do a release soon and this will fix itself.

kaushalmodi:
I checked today on https://gohugo.io/templates/rss/#section-rss that that link gets auto-hyperlinked:

image

kaushalmodi:
I confirm the issue.. 

```md
A section’s RSS will be rendered at /<SECTION>/index.xml (e.g., https://spf13.com/project/index.xml).
```

is creating a link to `https://spf13.com/project/index.xml)`, with that closing bracket.

---

This should technically be an issue on the https://github.com/gohugoio/hugo repo.

davidsneighbour:
I would think this is a Goldmark bug. A workaround would be to change into a full markdown link instead of hoping that the bug will be fixed :)

```
[https://spf13.com/project/index.xml](https://spf13.com/project/index.xml)
```

We could merge and close the PR then.

PR #976, opened 2019-12-12
Format URL

maiki:
Per https://discourse.gohugo.io/t/typo-error-on-templates-rss-section-rss/22267?u=maiki, the ending parentheses renders the URL incorrectly.


maiki:
Nevermind, should have checked the queue, first. ^_^

PR #980, opened 2019-12-14
fixed command for adding the public submodule

PascalMayr:
Me and other fellows had some problems adding the submodule. I don't know for sure if this is the right thing to do. It worked for me.


kaushalmodi:
The git command is correct.. the repo name is `<USERNAME>.github.io`, not `<USERNAME>.io`.

See https://github.com/gohugoio/hugoDocs/blame/d060b1a1d44d8c20e8fcd4597d5b94b485a2fd16/content/en/hosting-and-deployment/hosting-on-github.md#L47

PR #981, opened 2019-12-16
Change incorrect MIME type from image/jpg to image/jpeg

zaitseff:
According to multiple sources, both official ([IANA](https://www.iana.org/assignments/media-types/media-types.xhtml#image) and [RFC2046](https://tools.ietf.org/html/rfc2046)) and otherwise (eg, [Wikipedia](https://en.wikipedia.org/wiki/JPEG) and [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)), the official MIME type for JPEG images is `image/jpeg`, not `image/jpg`.  Change Hugo docs to match.

Note: Hugo itself may need to change: I did a quick grep of Hugo's source code, but did not find anything that looked relevant.

### DIFF
--- a/content/en/content-management/page-resources.md
+++ b/content/en/content-management/page-resources.md
@@ -17,7 +17,7 @@ menu:
 ## Properties
 
 ResourceType
-: The main type of the resource. For example, a file of MIME type `image/jpg` has the ResourceType `image`.
+: The main type of the resource. For example, a file of MIME type `image/jpeg` has the ResourceType `image`.
 
 Name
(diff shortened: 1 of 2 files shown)


bep:
>Note: Hugo itself may need to change: I did a quick grep of Hugo's source code, but did not find anything that looked relevant.

https://github.com/gohugoio/hugo/blob/master/media/mediaType.go#L143

We might as well do both code and docs changes in that repo as it needs to be synced.

PR #982, opened 2019-12-20
docs: link to Régis Full Partials series

DirtyF:
@regisphilibert writes excellent teaching material that helps dig Hugo internals 🙏 

props @jygastaud


regisphilibert:
Thanks @DirtyF! I would only reference the `partialCached` article (Part 1) rather than the whole series though.

PR #990, opened 2019-12-23
Add external links example to Render Hook Templates

inwardmovement:


bep:
I don't think your example is correct, but we can take that discussion on the forum.

PR #1000, opened 2019-12-31
show 'description' when hovered on function nav link

### LINKED ISSUE #999: (Docs) show 'description' when hovered on nav link
### Summary
We may show description of the item from nav links, on which the user is hovering.

### Pros
It would enable users to gain an understanding of where they are headed, since, sometimes name of (say) function or variable may not be self explanatory.

### Cons
No major cons.

SanmayJoshi:
### Changes
Added `title` attribute with function's description as its value, so that user hovering an entry of a function in sidebar (nav links) sees the function's short description. Attribute is added to `<a>` tag, alternatively it could be added to `<li>` tag too; though it wouldn't make any major difference. Closes #999.
### Previously
```
line 13
            <a href="{{.URL}}" class="db link hover-bg-gray hover-white pl3 pr2 pv2 {{if $currentPage.IsMenuCurrent "docs" . }}primary-color {{ else }}black {{end}}">
```
### Now
```
line 13
            <a href="{{.URL}}" class="db link hover-bg-gray hover-white pl3 pr2 pv2 {{if $currentPage.IsMenuCurrent "docs" . }}primary-color {{ else }}black {{end}}"{{if .Page.Description}} title="{{.Page.Description}}"{{end}}>
```

### DIFF
--- a/_vendor/github.com/gohugoio/gohugoioTheme/layouts/partials/nav-links-docs.html
+++ b/_vendor/github.com/gohugoio/gohugoioTheme/layouts/partials/nav-links-docs.html
@@ -10,7 +10,7 @@
         <ul class="{{ .Identifier }} desktopmenu animated fadeIn list pl0 bg-light-gray{{if $currentPage.HasMenuCurrent "docs" . }} db{{ else }} dn{{ end }}">
           {{- range .Children }}
           <li class="f6 fw4">
-            <a href="{{.URL}}" class="db link hover-bg-gray hover-white pl3 pr2 pv2 {{if $currentPage.IsMenuCurrent "docs" . }}primary-color {{ else }}black {{end}}">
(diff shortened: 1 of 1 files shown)


davidsneighbour:
I have requested @bep to review. I am 50/50 about this. The question is if the description is properly added everywhere and also if the maybe another method for "tooltips" might be used. The title attribute is interpreted differently in different browsers (string length) and also accessibility is a point to look at.

SanmayJoshi:
@davidsneighbour thanks for the input!
> if the description is properly added everywhere

Does including the `title` attribute in `if block` solves the problem?

> also if maybe another method for "tooltips" might be used

'tooltips' might be a great idea! They generally require additional JavaScript though.

> The title attribute is interpreted differently in different browsers (string length)

As per MDN, *The title global attribute contains text representing advisory information related to the element it belongs to.*  Also, as per [this question on Stack Overflow](https://stackoverflow.com/questions/8516235/max-length-of-title-attribute), maximum rendered length is 512 chars by IE. Chrome has 1024 chars. May be it seems sufficient. Thoughts?

PR #1016, opened 2020-01-17
Fixed description of 0.62.2 release notes

frerich:
This version improves on 0.62.1.


bep:
That description is autogenerated. For patch releases I have an assumption that most fixes is fixes for the recent main release (the 0 version). So it may not always be correct, but it isn't a typo.

PR #1017, opened 2020-01-19
Update .GroupByParamDate to reflect limitations

adrian5:
I ran into this, as [others](https://discourse.gohugo.io/t/solved-groupbyparamdate-produces-runtime-error-index-out-of-range/10668) did [before](https://discourse.gohugo.io/t/sorting-by-a-date-param/20755) me.

I've updated the documentation to point out that only a fully qualified ISO 8601 date will work. Since we're talking about custom parameters here, this wasn't apparent to me. Some users might e.g. only store `yyyy-mm-dd` or a time without second precision.

### DIFF
--- a/content/en/templates/lists.md
+++ b/content/en/templates/lists.md
@@ -457,7 +457,7 @@ In the above example, you may want `{{.Title}}` to point the `title` field you h
 
 ### By Page Parameter in Date Format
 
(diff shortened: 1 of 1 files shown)


adrian5:
Actually, the issue runs deeper than that: https://github.com/gohugoio/hugo/issues/3983

I will close the pull request and create a new one. It's important to notify the reader of this limitation!

PR #1022, opened 2020-01-27
Add note to List Templates about .GroupByParamDate not working

adrian5:
This should prevent users from [wasting](https://discourse.gohugo.io/t/solved-groupbyparamdate-produces-runtime-error-index-out-of-range/10668) their [time](https://discourse.gohugo.io/t/sorting-by-a-date-param/20755) on an unimplemented feature. I didn't get to work with TOML frontmatter either, but that's beside the point as YAML frontmatter is far more common on the web.

### DIFF
--- a/content/en/templates/lists.md
+++ b/content/en/templates/lists.md
@@ -457,6 +457,10 @@ In the above example, you may want `{{.Title}}` to point the `title` field you h
 
 ### By Page Parameter in Date Format
 
+{{% note %}}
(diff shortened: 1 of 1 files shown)


adrian5:
Not fixed afaik. Please add a note or remove the feature from the notes until implemented.

adrian5:
I take this has been [resolved](https://github.com/gohugoio/hugo/issues/3983] now? I cannot verify right now, but if so, you can close this PR.

bep:
It has been fixed.

PR #1023, opened 2020-01-27
Adding an issue link to the issue creation workflow

davidsneighbour:
This config enables us to add external links to the process of creating new issues. I propose to add a friendly link to the community, just in case the issue opener is not aware of this forum. This might be useful for the gohugoio/hugo repo too, not sure if it's possible to cherry-pick from different repos?

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,4 @@
+contact_links:
+    - name: Ask the Hugo community
+      url: https://discourse.gohugo.io/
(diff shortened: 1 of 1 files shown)


davidsneighbour:
This was a bit lost somehow. I'll merge it now even though it's my own PR, because I think this addition makes much sense to decrease the amount of opened issues here. This might be useful too for the main repo.

davidsneighbour:
This did not go as intended. I reverted the PR and will open a new one with issue templates. Only adding a link to the community without defining at least one issue template will hide the issue form to the uninitiated.

Selection_014

I'll add a new PR with templates for "Bug" and "Enhancement".

Sorry for the mess. I don't see a way to test this properly.

PR #1031, opened 2020-02-03
Explain that Ace/Amber support has been removed since Hugo 0.62

peterkappus:


peterkappus:
Thanks @bep for the speedy PR merge. I hope support for Ace (or Pug or similar) returns to Hugo someday 😢 ...meanwhile, I'll make friends with angle brackets again 😉 Thanks again.

PR #1041, opened 2020-02-19
Fix typo in config docs

adrian5:


bep:
@coliff you do some great work with these docs contributions. If you don't mind, I could add you as a collaborator so you get commit access. You can still create PRs for bigger stuff, but then you wouldn't have to go that distance just to do obvious copy-edits/typo fixes. Let me know.

coliff:
thanks Bep! Sure, yes, I'd be happy to help out with this some more.

PR #1054, opened 2020-03-11
Fix typo in page-bundles.md

9ft:
Fix typo in page-bundles.md


bep:
Thanks, but since this is a copy-paste from the Unix tree command, I would hardly call it a typo. I'll admit your way looks prettier, but we cannot maintain the docs on this detail level.

PR #1058, opened 2020-03-16
Remove broken link on "where" page

Alexander01998:
The link "[wherefunction]" was not defined, so the square brackets where visible on the rendered page.
image

It seems that this link would have just gone back to the top of the page, so I removed it instead of fixing it.


bep:
Me too ...

PR #1059, opened 2020-03-16
Update configuration-markup.md (security) 

mbvsj:
Links with `target="_blank"` should also have `rel="noopener noreferrer"` attribute to keep users safe. I think it's good to include this because it seems like there are many Hugo sites who use this code snippet and they don't seem to be aware of this vulnerability.

More info: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/


coliff:
I agree with adding `rel="noopener"` - but `rel="noreferrer"` can have a negative impact on analytics for outbound links so I'd advise against including that here.

REF: https://developers.google.com/web/tools/lighthouse/audits/noopener

PR #1061, opened 2020-03-18
No need for type="text/javascript"

schalkneethling:
This is a small change to the script tag in the example. There is no longer a need to include the `type` attribute on the `script` tag. Saves a couple of bytes and is generally good practice these days.

### DIFF
--- a/content/en/hugo-pipes/fingerprint.md
+++ b/content/en/hugo-pipes/fingerprint.md
@@ -25,5 +25,5 @@ Any so processed asset will bear a `.Data.Integrity` property containing an inte
 ```go-html-template
 {{ $js := resources.Get "js/global.js" }}
 {{ $secureJS := $js | resources.Fingerprint "sha512" }}
-<script type="text/javascript" src="{{ $secureJS.Permalink }}" integrity="{{ $secureJS.Data.Integrity }}"></script>
+<script src="{{ $secureJS.Permalink }}" integrity="{{ $secureJS.Data.Integrity }}"></script>
 ```


content/en/hugo-pipes/fingerprint.md
```diff
@@ -25,5 +25,5 @@ Any so processed asset will bear a `.Data.Integrity` property containing an inte
 ```go-html-template
 {{ $js := resources.Get "js/global.js" }}
 {{ $secureJS := $js | resources.Fingerprint "sha512" }}
-<script type="text/javascript" src="{{ $secureJS.Permalink }}" integrity="{{ $secureJS.Data.Integrity }}"></script>
+<script src="{{ $secureJS.Permalink }}" integrity="{{ $secureJS.Data.Integrity }}"></script>
```
schalkneethling: I also noticed that I got a cors error in Chrome Canary(might be in other places as well). Should this also perhaps container `crossorigin="anonymous"`?

davidsneighbour: I am against this. You are right, that the type attribute is not neccessary, but _only in HTML5_. You can not assume, that everyone and every theme is using HTML5 and "modern browsers". As long as that is the case you will need to give fallback options like the type tag. 

About your CORS issue: it's a notice, not an error, right? If the javascript file is in the local installation (relative to the file with the script tag and NOT on another domain (CDN)), then the CORS notice can be ignored. Anyway, that would be the topic of an issue, not a PR.

schalkneethling: > As long as that is the case you will need to give fallback options like the type attribute.

Fair enough. I wonder whether a comment to that effect can be added as well as perhaps a note to nudge people in the direction of `<!doctype html>`

> About your CORS issue: it's a notice, not an error, right? If the javascript file is in the local installation (relative to the file with the script tag and NOT on another domain (CDN)), then the CORS notice can be ignored. Anyway, that would be the topic of an issue, not a PR.

It has been a while 😄 Will have to try this again and see whether it was an error or warning. Here also, I wonder whether adding a comment would assist users?

Thank you for your feedback.

bep:
I will close this, as the type attribute has several (useful) meanings

PR #1065, opened 2020-03-25
Add note on setting baseURL

joelostblom:
This note [already exists for the GitHub pages setup section](https://gohugo.io/hosting-and-deployment/hosting-on-github/#github-project-pages), and it appears that it is needed for Hugo themes to display correctly on GitLab as well.


joelostblom:
I think adding this note is important because it is needed for Hugo themes to display correctly on GitLab.

joelostblom:
Thanks for merging!

PR #1068, opened 2020-03-31
HTTPS links to external sites

coliff:


coliff:
I'm not really a fan of Calbire's Image Actions because the compression is lossy. Have those images previously been optimised? If so, they'll look worse than they did before. I'm in favour of adding https://imgbot.net instead (100% lossless)

coliff:
@bep - I think I'll revert that commit from the calibre-action and then merge if that's ok with you.

PR #1074, opened 2020-04-07
Add roller2hugo to migrations.md

jlevon:


jlevon:
anyone?

jlevon:
ping?

coliff:
That project has no stars and the README warns that it hasn't been thoroughly tested so maybe it night be better to not add it to this page yet?

jlevon:
If there was an alternative to link to I'd agree 100%. Surely something is better than nothing?

coliff:
Yep, maybe :-) . Let's see what @bep says.

bep:
I think that people who really need this can Google for it.

jlevon:
Then what's the point of the page at all?
BTW you tried googling for it?

PR #1082, opened 2020-04-12
Update documentation for Ugly URLs

MauriceBrg:
Update documentation to replace the CLI-Argument with an environment variable as the CLI-argument no longer exists. 

References:
- gohugoio/hugo#4343
- https://gohugo.io/getting-started/configuration/#configure-with-environment-variables


MauriceBrg:
@davidsneighbour 
Done :)

PR #1086, opened 2020-04-16
Mentioning a range is equivalent to foreach

daveredfern:
I’m new to Go and really struggled to find this. I added this sentence so it would be searchable for other people to find. At the moment, searching for a foreach doesn’t return range and IMHO it should.

I suspect there are other places that could benefit from this.

Another idea, further down the line would be a table showing the equivalents in other templating languages such as liquid. This would make transition from Jekyll much simpler for me.

### DIFF
--- a/content/en/functions/range.md
+++ b/content/en/functions/range.md
@@ -20,6 +20,6 @@ draft: false
 aliases: []
 ---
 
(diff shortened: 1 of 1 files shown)


daveredfern:
@coliff thanks for spotting that. I’ve fixed it.

PR #1087, opened 2020-04-17
Add Showcase - KeyCDN

jonas-krummenacher:
We've written a showcase as we're using Hugo throughout our website and we're very happy with it.


bep:
I merged this manually, should be live in a few minutes -- thanks for this, much appreciated!

PR #1091, opened 2020-04-24
Update internal.md

GiovanniMet:
the variable disqusShortname in code and also in example have this name DisqusShortname


bep:
The old value was correct.

PR #1094, opened 2020-05-01
Submitting Digital.gov to the Hugo Showcase

jeremyzilar:
**Digital.gov** helps people in government deliver better, more accessible digital services through publishing essential guidance, resources, tools, and online events that make it easier for people to design, build, and deliver essential services for the public.

Learn more at https://digital.gov/


bep:
Thanks, this is really appreciated!

PR #1099, opened 2020-05-08
Link to the index function fixed

dsolemorera:
The index function was pointing to https://gohugo.io/functions/index wich is the "Functions Quick Reference" webpage.


jmooring:
This was resolved with https://github.com/gohugoio/hugoDocs/pull/1133.

PR #1101, opened 2020-05-09
Disambiguate global and page resources

tjamet:
There are 2 different kind of resource, one that is attached
to the page, matching all non-rendered files, and the global
one matching all files in `/assets`. Although they have similar
behaviour, the way to access both of them are slightly different.

Explicitly express differences between both and clarify when a page
is attached resources and when not.


bep:
This is great, thanks!

PR #1103, opened 2020-05-10
Update configuration-markup.md (add rel="noopener")

h-enk:
For security /performance reasons, add `rel="noreferrer noopener"` to code example for how the render-link.html template could look. See also: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target

### DIFF
--- a/content/en/getting-started/configuration-markup.md
+++ b/content/en/getting-started/configuration-markup.md
@@ -134,7 +134,7 @@ PlainText
 Here is a code example for how the render-link.html template could look:
 
 {{< code file="layouts/_default/_markup/render-link.html" >}}
-<a href="{{ .Destination | safeURL }}"{{ with .Title}} title="{{ . }}"{{ end }}{{ if strings.HasPrefix .Destination "http" }} target="_blank" rel="noreferrer noopener"{{ end }}>{{ .Text | safeHTML }}</a>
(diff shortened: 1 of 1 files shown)


coliff:
Agree with `noopener`, but `noreferrer` can have a negative effect on analytics data (by not providing the reffing website on external links). So i suggest to just use `noopener`.

h-enk:
@coliff, if you want other websites to see traffic from your website as 'Referral traffic', then simply do not open external links in a new tab. See also: https://www.reliablesoft.net/noreferrer-noopener/

Using `target="_blank"` without `rel="noreferrer"` and `rel="noopener"` makes the website vulnerable to `window.opener` API exploitation attacks ([vulnerability description](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)). See also: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Security_and_privacy

coliff:
Firefox does support `noopener` now (REF: https://caniuse.com/#feat=rel-noopener) so I think just that should be fine.

h-enk:
Yes, so? Firefox also does support `noreferrer`. Your're missing the (security) point here.

coliff:
sorry, maybe my last message wasn't clear - I meant that I think just adding `rel="noopener"` is sufficient. There isn't a need to add `rel="noreferrer"` too since all major browsers support `rel="noopener"`.

h-enk:
Yes, I see now /I agree. I will remove the `rel="noreferrer"` part.

PR #1105, opened 2020-05-14
Hugo Docs CSS Visual Glitch

alexandre1985:
gohugo.io visual css bug. The code box was expanding all the width by a mouse hover. This bug caused visual problems by the document-seeker person and it make gohugo.io docs look a bit less professional.

### DIFF
--- a/_vendor/github.com/gohugoio/gohugoioTheme/assets/css/_code.css
+++ b/_vendor/github.com/gohugoio/gohugoioTheme/assets/css/_code.css
@@ -61,25 +61,6 @@ pre {
   content: "Copied"
 }
 
-@media (--breakpoint-large) {
-  .full-width, pre.expand:hover
-  {
-  /*width: 100vw;
-  position: relative;
-  left: 50%;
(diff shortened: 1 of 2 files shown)


davidsneighbour:
I would not term that a glitch. The codebox is supposed to expand to show the full code example. The problem is within the `100vw` used, that use more space than is available (sidebar is not counted in). So moving to `80vw` might make the codebox not hover over the sidebar anymore. Or, another approach, the codeboxes background color could be set instead of being transparent, so it sits on top of the sidebar navigation.

I wouldn't just remove it.

Also: your editing tool is "fixing" a lot of spacing and makes it hard to see where your changes really are.

alexandre1985:
Sorry for the identation "fix" :disappointed: .
Hopefully someone can correct what I want to correct

alexandre1985:
This can be deleted. replaced my https://github.com/gohugoio/hugoDocs/pull/1155 because too much lines were affected by my text editor identation "fix" (or unfix)

PR #1107, opened 2020-05-15
Use-modules: More help how to get started importing a theme

siedentop:
It took me too long to figure out how to do this simple thing. How can I import a theme using the new hugo-module concept.

Let's update the documentation so that it is easier for others.

### DIFF
--- a/content/en/hugo-modules/use-modules.md
+++ b/content/en/hugo-modules/use-modules.md
@@ -30,8 +30,21 @@ Use `hugo mod init` to initialize a new Hugo Module. If it fails to guess the mo
 hugo mod init github.com/gohugoio/myShortcodes
 ```
 
-
 Also see the [CLI Doc](/commands/hugo_mod_init/).
+
+## Use a Module for a Theme
+The easiest way to use a for a theme is to import it in the config. 
+
+1. Initialize the hugo module system: `hugo mod init github.com/<your_user>/<your_project>`
+2. Import the theme in your `config.toml`: 
+
+```toml
+[module]
+  [[module.imports]]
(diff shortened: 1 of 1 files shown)


content/en/hugo-modules/use-modules.md
```diff
@@ -30,9 +30,22 @@ Use `hugo mod init` to initialize a new Hugo Module. If it fails to guess the mo
...
+## Use a Module for a Theme
+The easiest way to use a for a theme is to import it in the config. 
+
+1. Initialize the hugo module system: `hugo mod init github.com/<your_user>/<your_project>`
+2. Import the theme in your `config.toml`: 
+
+```toml
+[module]
+  [[module.imports]]
+    path = "github.com/dzello/reveal-hugo"
```
bep, line 44: Thanks. It would be great, though, if you could pick a "more neutral" theme as an example, e.g. https://github.com/spf13/hyde/ (ideally we should have a simple theme in the gohugoio org, but spf13 did start this project, so might as well give him some PR).

content/en/hugo-modules/use-modules.md
```diff
@@ -30,9 +30,22 @@ Use `hugo mod init` to initialize a new Hugo Module. If it fails to guess the mo
...
+
+theme = "reveal-hugo"
```
bep, line 46: This can be removed. You should not use both, and it also gets shadowed by the TOML map above.

siedentop:
@bep Thanks for the review. I’ll apply your suggestions later today or tomorrow.

siedentop:
@bep please review again and merge.

Issue #1106, opened 2020-05-14
Please document the Extended version


rgov:
I've been using Hugo for a while but just learned about the existence of "Hugo Extended" that seems to have SCSS support. However the differences between the versions are not really highlighted explicitly in the documentation so it's not really clear.

1. Why would I ever _not_ want to use the extended version?

2. The Install page explains how to get it in certain circumstances. But it is not mentioned for macOS, even though in the Releases page there's both a regular and extended version. It turns out that `brew` installs the extended version by default.

3. The only real documentation seems to be a minor "note" in the [0.43 release notes](https://gohugo.io/news/0.43-relnotes/).

My true preference though would be to eliminate this distinction.

rgov:
Thanks for moving it.

I don't presume to know all use cases of Hugo out there, but I am struggling to picture a need to build a static site in an environment that must be so minimal it cannot even include libc. Is this really a constraint? Well anyway, that can be another discussion beyond just documenting the differences.

bep:
>, but I am struggling to picture a need to build a static site in an environment that must be so minimal it cannot even include libc. 

You asked for some arguments. The main argument is the build chain. I spent 2 long days filled with gray hairs to get the extended build up and running on Circle Ci for Linux, MacOs and Windows. Extending that to the other platforms was not in my time budget ...

davidsneighbour:
See #1152 where we will continue and finish the discussion.

PR #1124, opened 2020-05-27
Added notes on Hugo Versions and Sass Support

faraixyz:
I tried to summarize the differences between the versions of Hugo (#1106). I also made a few notes on SASS's limitations.

### DIFF
--- a/content/en/hugo-pipes/scss-sass.md
+++ b/content/en/hugo-pipes/scss-sass.md
@@ -1,6 +1,6 @@
 ---
 title: SASS / SCSS
-description: Hugo Pipes allows the processing of SASS and SCSS files.
+description: Hugo Pipes allows the processing of SASS and SCSS files. Only works in Hugo Extended.
 date: 2018-07-14
 publishdate: 2018-07-14
 lastmod: 2018-07-14
@@ -15,6 +15,9 @@ sections_weight: 02
 draft: false
 ---
 
+{{% note %}}
+Built-in SASS/SCSS only works on the extended version of Hugo.
+{{% /note %}}
 
 Any SASS or SCSS file can be transformed into a CSS file using `resources.ToCSS` which takes two arguments, the resource object and a map of options listed below.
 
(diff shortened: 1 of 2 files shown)


content/en/hugo-pipes/scss-sass.md
```diff
@@ -23,6 +26,12 @@ Any SASS or SCSS file can be transformed into a CSS file using `resources.ToCSS`
 {{ $style := $sass | resources.ToCSS }}
 ```
 
+### Limitations
+
+Asides from only working on the extended version of Hugo, some newer SASS features might not be present, like the @use rule. If you look through [Sass's documentation](https://sass-lang.com/blog/the-module-system-is-launched), any features which aren't supported by `libsass` won't work in Hugo Extended.
+
+To get newer SASS features, you will have to add it manually, as shown by the [Victor Hugo](https://github.com/netlify/victor-hugo) starter kit.
```
bep, line 33: I would rather leave this "newer SASS features" section out.

content/en/troubleshooting/faq.md
```diff
@@ -54,3 +54,15 @@ error: failed to transform resource: TOCSS: failed to transform "scss/main.scss"
 We release two set of binaries for technical reasons. The extended version is not what you get by default for some installation methods. On the [release page](https://github.com/gohugoio/hugo/releases), look for archives with `extended` in the name. To build `hugo-extended`, use `go install --tags extended`
 
 To confirm, run `hugo version` and look for the word `extended`.
+
+## What Is The Difference Between Hugo Regular and Hugo Extended and Which Version Should I Choose?
+
+The biggest difference is that Hugo Extended includes [built-in SCSS/SASS support]({{< ref "scss-sass">}}). This comes at the expense of some performance[^1].
```
bep, line 60: It is the ONLY difference. And the extended part does not come with any performance overhead. I would just drop this perfomance debate. A native Go implementation of SASS would be slightly faster, but a C++ binding version is plenty fast -- and since you don't have anything to comare it with, it's not slower than anything, either.

content/en/troubleshooting/faq.md
```diff
@@ -54,3 +54,15 @@ error: failed to transform resource: TOCSS: failed to transform "scss/main.scss"
...
+
+Some good reasons to stick to regular Hugo are:
+
+* If you don't need SASS/SCSS.
+* You want to use certain [SASS features which aren't supported on Hugo Extended]({{< relref "scss-sass#limitations" >}}), like `@use`.
+* You want to run Hugo in a resource constrained environment[^1], like a minimal docker container.
+
+[^1]: The extended version includes [libsass](https://sass-lang.com/libsass), a C dependency which comes with overhead, reduced portability (since C is needed in addition to Go) and a complicated build chain. Given Hugo's emphasis on performance, [C dependencies aren't allowed in Hugo's source code](https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#code-contribution) and this is the only exception.
```
bep, line 68: Again, the regular user do not need to know about all of this. I would suggest that we just remove the section above.

content/en/hugo-pipes/scss-sass.md
faraixyz: I’ll just mention that some features aren’t supported by the built in pipeline which are noted by not being compatible in Sass’s documentation.

content/en/troubleshooting/faq.md
faraixyz: I’ll shorten this section to just the Sass Support.

PR #1125, opened 2020-05-31
Remove JustComment

Leon0824:
JustComment is shutting down.


coliff:
Reference: https://just-comments.com/blog/2020-03-23-hyvor-talk-alternative-system-supports-justcomments.html

PR #1127, opened 2020-06-01
Add an example of using front-matter

abitrolly:
Example by @rindolphus from https://discourse.gohugo.io/t/how-to-display-a-custome-page-variable/1557/10

I am evaluating Hugo and don't have anything generated yet. Please check that it
1. actually works and
2. is the best way

Also, is it true that the `pdf` can not be referenced from markdown itself? I can reference the var from template only, right?

### DIFF
--- a/content/en/content-management/front-matter.md
+++ b/content/en/content-management/front-matter.md
@@ -27,12 +27,12 @@ toc: true
 If a variable is defined in front matter of `report.md`, as
 ```
 --- 
-pdf: "energy.pdf"
+energydoc: "energy.pdf"
 ---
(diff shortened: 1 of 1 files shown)


abitrolly:
@jmooring should be better now. PTAL.

bep:
Thanks for this. This looks good, but I don't think it adds enough new information, so I'm not going to merge this right now. But your contribution is very much appreciated!

abitrolly:
@bep it adds example, which drastically cuts down the time that is needed to understand the concept. What makes you think that examples are not useful?

PR #1136, opened 2020-06-03
fix new post command directory

vrechson:
The quickstart says that hugo new posts/my-first-post.md is the command to create new posts but it wasn't working. I could see another person with the same problem here: https://stackoverflow.com/questions/51974423/hugo-posts-not-displaying. It seems that by default hugo is looking for posts in content/posts/ not posts/


davidsneighbour:
The command `hugo new posts/test.md` creates a new item in `content/posts/test.md`. all locations are _relative_ to the content directory in this command. If you have an issue with creating new content with this command at the right location please feel free to open a conversation in discourse.gohugo.io. I will close this pull request.

PR #1137, opened 2020-06-03
update live reload location in docs to head of template

willduncanphoto:
small change, just noticed it was wrong in the docs, still had the /body location.


bep:
>I believe this whole sentence can be dropped, because obviously something changed in the way the livereload is injected into the HTML.

It now first looks for a `<head>` ... and the reason is Turbolinks (and similar libraries), which replaces the body on navigation.

PR #1140, opened 2020-06-06
Dynamic menu localisation with menu Identifier attribute

sergioisidoro:
I have struggled with menu localisation a bit. I don't want to configure the menu per language due to large number of languages. Also I want the translations of the menu to be in the `i18n` folder so it can be translated using common translation management tools.

Therefore I added this documentation that explains how to make a simple dynamic localisation using menu `identifiers` and `i18n`. 

PS: Sorry my english. Not my native language.

Screenshot 2020-06-06 at 20 49 52


sergioisidoro:
Can anyone review this?

PR #1145, opened 2020-06-16
Updated to include a new menu template example

kenmorse:
Updated to include a template example for when the menu is defined in site config file.


kenmorse:
Just tried to resolve some documentation bugs; not sure why they are not being pulled in.

bep:
I'm sorry for the late follow up. Your example looks correct, but it looks a little clumsy. If that is how it need to be implemented, I prefer to fix Hugo instead of adding less than optimal examples to the docs.

PR #1146, opened 2020-06-17
Warn about zero weights in front matter.

OmarEmaraDev:


jmooring:
I don't think weights need to start at 1. Negative numbers seem to work fine as well. Also, please describe the negative side effects of using zero as a weight. Is the sort order indeterminate? Finally, must the weight be an integer, or is a float value acceptable?

OmarEmaraDev:
@jmooring  See this [post](https://discourse.gohugo.io/t/menu-weight-and-other-oddity/5653/4) on the Hugo forum. There are no negative side effects of using zero as weight, it is just that it will not work as the user might expect, as I wrote in the patch, zero is considered as an unset weight, while the user might consider it a weight that will give the page the highest priority. And yes, by skimming over the code, it should be an integer.

bep:
> it is just that it will not work as the user might expect

I agree, but since having no weight (as in either not setting it, or setting it to 0) is a perfectly fine use case:

OmarEmaraDev:
@bep Sounds good. I clarified the description.

jmooring:
@OmarEmaraDev Thanks for this. For better or worse I spent a fair amount of time in Drupal land, where negative, positive, and zero (sorted between negative and positive) weights are common. This clarification should be helpful for others with similar experience.

PR #1153, opened 2020-06-18
Add note GroupByParamDate only working with TOML frontmatter.

finnito:
There has been extended discussions on the forum and in the issues about GroupByParamDate not always working and it would be good to be upfront about this in the documentation.

- https://github.com/gohugoio/hugo/issues/3983
- https://discourse.gohugo.io/t/is-groupbyparamdate-broken/22922
- https://discourse.gohugo.io/t/solved-groupbyparamdate-produces-runtime-error-index-out-of-range/10668


jmooring:
The underlying issue has been resolved with 0.73.

PR #1156, opened 2020-06-19
more loving into last commit -> css fix of expanding code box in docs

alexandre1985:
on previous commit the max-width was set to 50vw, as a fix. With some viewport width this is not the perfect solution.
so I'm fixing this, so that in all viewport width the code box that expands on hover (.expand class) never overlaps the right-side navigation column.
my solution is this commit; setting `.expand` max-width to 48vw

### DIFF
--- a/_vendor/github.com/gohugoio/gohugoioTheme/assets/output/css/app.css
+++ b/_vendor/github.com/gohugoio/gohugoioTheme/assets/output/css/app.css
@@ -4828,7 +4828,7 @@ pre {
   right: 50%;*/
   /*margin-left: -30vw;*/
   margin-right: -30vw;
-  max-width: 50vw;
+  max-width: 48vw;
   }
 }
 .code-block .line-numbers-rows {


alexandre1985:
this is a continuation of https://github.com/gohugoio/hugoDocs/pull/1155

davidsneighbour:
What exactly was the problem that is fixed with going 2vw lower? My guess is the display width on some mobile devices?

alexandre1985:
Current state:
bug-1
bug-2

With this pull request merge:
fix-1
fix-2

davidsneighbour:
It still looks aweful near the sidebar with 48vw. Maybe we should use a percentage or a calculation of "100% - sidebarwidth - 2rem" where the sidebarwidth is the known sidebar width and the 2rem the padding that is used between left (navigation) column and content column to make it more consistent? The last PR looked well on my devices. No overlapping.

alexandre1985:
Yes, we could. Is right navbar column width fixed (widthin the programmed media query)?

alexandre1985:
Also, for this I suggest adding css varables for sidebar padding and margin. Or else that calculation modification will not be permanent, as I presume, your wish according to my interpretation of your suggestion reply

davidsneighbour:
There is no assets folder on this repo, just a finished compiled css file... Not sure how to proceed. Let's open an issue about this and wait for it. I would like to have the hugo documentation have it's own asset pipe... we will be more flexible then.

alexandre1985:
Good proposal

alexandre1985:
To make things clear @davidsneighbour , I was talking about CSS variables. CSS variables are **native** to CSS, does not require SASS or preprocessing.
You may google CSS variables if you are not familiar. Is quite a new feature of CSS

davidsneighbour:
I am merging this now. If the design is re-done we will find a nice solution.

alexandre1985:
If the re-design is done, hope the codeboxes don't have bugs, after the re-design

PR #1158, opened 2020-06-25
Fix page kinds shortcode not rendering on Taxonomies page

jessicahuynh:
A small fix; I noticed that the table for page kinds in the taxonomies page was showing Markdown instead of rendering appropriately. I also removed a trailing space for `term` in the table.


coliff:
thanks for the fix. 👍

PR #1176, opened 2020-07-20
dateformat docs

TamaMcGlinn:
Added more examples and explicitly state the dateformat layout string has a fixed date.

The information was already available, but you had to go through the 'more info' link. Such fundamentals should be stated immediately. Having three examples also really helps because then you can see which parameter is the format and which is the date being passed - the parameters passed could easily be confused as they may legally have the same value.

### DIFF
--- a/content/en/functions/dateformat.md
+++ b/content/en/functions/dateformat.md
@@ -17,10 +17,12 @@ relatedfuncs: [Format,now,Unix,time]
 deprecated: false
 ---
 
-`dateFormat` converts the textual representation of the `datetime` into the specified format or returns it as a Go `time.Time` type value. These are formatted with the layout string.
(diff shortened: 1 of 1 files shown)


jmooring:
1. The word "the" is duplicated.
2. It would be good to include one example that has the time (with offset) as well as the date.
3. This is wrong:

   > dateFormat converts the textual representation of the datetime into the specified format or returns it as a Go time.Time type value. 

   Maybe this instead?

   > dateFormat returns a string representation of a `datetime` value using the specified layout string. The `datetime` must be a [time.Time](https://golang.org/pkg/time/) value, or a valid string representation of a time.Time value.

4. This is also wrong:

   > Converts the textual representation of the datetime into the specified format.

   Maybe this instead?

    > Returns a string representation of a `datetime` value using the specified format.

bep:
Sorry for the delay. This function has since then gotten a new name structure etc. which makes this PR hard to merge, so I'm closing it.

PR #1187, opened 2020-08-10
Improved slice.md

jwarner112:
- Added more examples
- Marked `ITEM` as optional in signature
- Minor refactor of markdown footnotes


bep:
Thanks for this, and sorry for the delay. This looks good, but it uses a lot of terms/API that's not Hugo (e.g {{ if .IsUser }}) and as such will not work, so I don't think we can add this.

PR #1190, opened 2020-08-20
Add Victor Hugo Kontent boilerplate

makma:
Add Victor Hugo Kontent boilerplate into starter kits.


makma:
> This issue has been automatically marked as stale because it has not had recent activity. The resources of the Hugo team are limited, and so we are asking for your help.
> If you still think this is important, please tell us why.
> This issue will automatically be closed in the near future if no further activity occurs. Thank you for all your contributions.

I believe it's beneficial for the new Hugo users and developers to get in the loop as fast as possible, so they are able to find all available information about CMS alternatives and sample code starters.

bep:
Thanks for this, but we need to reduce the number of tools in this section and do a general clean up, so we're not adding more for now.

PR #1201, opened 2020-08-27
Revise content-management/cross-references

### LINKED ISSUE #1199: Remove documentation related to hashed unique heading anchors
Hugo generates _hashed_ unique heading anchors under the following conditions:

```toml
[markup]
defaultMarkdownHandler = "blackfriday"  # Default is goldmark
[markup.blackFriday]
plainIDAnchors = false                  # Default is true
```

Neither of these are the default settings, and Blackfriday is slated for deprecation (https://github.com/gohugoio/hugo/issues/6487). 

The current documentation suggests that the default behavior is to generate _hashed_ unique heading anchors. It is not. By default, we generate unique identifiers on a _page_ by incrementing with an integer.

See <https://gohugo.io/content-management/cross-references/#hugo-heading-anchors>

PR to follow...

jmooring:
This could use more eyeballs. 

Closes #1199

### DIFF
--- a/content/en/content-management/cross-references.md
+++ b/content/en/content-management/cross-references.md
@@ -15,31 +15,39 @@ aliases: [/extras/crossreferences/]
 toc: true
 ---
 
-
-The `ref` and `relref` shortcode resolves the absolute or relative permalink given a path to a document.
+The `ref` and `relref` shortcodes display the absolute and relative permalinks to a document, respectively.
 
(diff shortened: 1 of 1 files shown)


bep:
This looks good. An added related comment is that the relRef feature was added before we got link render hooks. I would probably not use relRefs myself as they are unportable and ... ugly.

See https://portable-hugo-links.netlify.app/ for another way.

PR #1206, opened 2020-08-29
Add Bip hosting guide

joelkennedy:
Hello!

This PR adds a hosting guide on how to deploy a Hugo website to Bip.

Bip contains automatic framework detection for Hugo, so this is reflected in the guide. Further information on this and the up to date list of supported frameworks can be found [here](https://bip.sh/integrations). Regarding this page, I couldn't find documentation on logo usage guidelines, so please let me know if including the logo here isn't appropriate.


bep:
Thanks for this, but we need to reduce the number of tools in this section and do a general clean up, so we're not adding more for now.

PR #1207, opened 2020-08-30
Clarify treatment of zero weights

amdw:


content/en/templates/taxonomy-templates.md
```diff
@@ -158,7 +158,11 @@ Hugo uses both `date` and `weight` to order content within taxonomies.
 
 Each piece of content in Hugo can optionally be assigned a date. It can also be assigned a weight for each taxonomy it is assigned to.
 
-When iterating over content within taxonomies, the default sort is the same as that used for [section and list pages]() first by weight then by date. This means that if the weights for two pieces of content are the same, than the more recent content will be displayed first. The default weight for any piece of content is 0.
+When iterating over content within taxonomies, the default sort is the same as that used for [section and list pages]() first by weight then by date. This means that if the weights for two pieces of content are the same, than the more recent content will be displayed first.
```
coliff, line 161: I think:
than the more recent content will be displayed first
should be:
**then** the more recent content will be displayed first

amdw: Agreed! That problem was there in the old version too, but I'm happy to fix it here. Done!

coliff: Cool. I can't review this from a technical perspective (I don't know the logic), but It looks good I think - though it looks like that  `[section and list pages]()` is an error (empty link).
Again - not your fault as it was there before, but maybe we should fix in this same PR. :-)

amdw: Thanks. I'm not sure what page that link was supposed to refer to, so I've removed the link altogether.

For the record, the zero-weight special-casing logic I'm referring to here is:

https://github.com/gohugoio/hugo/blob/cdfd1c99baa22d69e865294dfcd783811f96c880/resources/page/pages_sort.go#L113

coliff (approved):

PR #1208, opened 2020-08-31
Clarify pagination documentation

amdw:
It is not necessary that the pages passed to .Paginate be a subset of .Pages.
Also, clarify that more than one call to .Paginator or .Paginate will give unexpected results.
See https://github.com/gohugoio/hugo/issues/7609

### DIFF
--- a/content/en/templates/pagination.md
+++ b/content/en/templates/pagination.md
@@ -43,9 +43,24 @@ Setting `Paginate` to a positive value will split the list pages for the homepag
 There are two ways to configure and use a `.Paginator`:
 
 1. The simplest way is just to call `.Paginator.Pages` from a template. It will contain the pages for *that page*.
-2. Select a subset of the pages with the available template functions and ordering options, and pass the slice to `.Paginate`, e.g. `{{ range (.Paginate ( first 50 .Pages.ByTitle )).Pages }}`.
+2. Select another set of pages with the available template functions and ordering options, and pass the slice to `.Paginate`, e.g.
+  * `{{ range (.Paginate ( first 50 .Pages.ByTitle )).Pages }}` or
+  * `{{ range (.Paginate .RegularPagesRecursive).Pages }}`.
 
 For a given **Page**, it's one of the options above. The `.Paginator` is static and cannot change once created.
+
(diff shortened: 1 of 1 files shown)


content/en/templates/pagination.md
```diff
@@ -43,10 +43,25 @@ Setting `Paginate` to a positive value will split the list pages for the homepag
...
+```
+$paginator := ""
+{{ if c }}
+    {{ $paginator = .Paginator }}
+{{ else }}
+    {{ $paginator = .Paginate x }}
+{{ end }}
+```
+
+and **not** `$paginator := cond c .Paginator (.Paginate x)`, because this will cause both `.Paginator` and `.Paginate x` to be evaluated.
```
davidsneighbour, line 63: I am not sure if there should be a sample of "don't use this code". It is noted in the documentation for `cond` that always both conditions are executed. https://gohugo.io/functions/cond/#readout

Maybe just write that once .Paginator or .Paginate ran they need to be reset (not sure if that functionality is available, if not, that might be a feature request) or that multiple calls can't run subsequently per page build.

bep: >It's essential that there be **at most one call** to either 

That statement is not correct. The side effect is that the result of the first invocation to either is cached, which allows you to call it thousands of times. The paginator is static, so there is only 1 for a given page

amdw: Thanks for the clarification. I've reworded the statement to be more accurate. Please could you have another look?

To be honest, I like having the example of what you should not do in there, because it helps clarify what can go wrong.

I knew in theory that `cond` is just a function with eagerly evaluated arguments, but it's not common that you have to think about such things, because side effects are rare in Hugo templates. Since we are dealing with quite a surprising kind of side effect here, I think the example is useful. But if you really feel strongly that it shouldn't be there, I will take it out.

amdw: Hi! Did my latest changes address your concerns, or is there anything more I need to do to get this merged? Thanks!

davidsneighbour: Looks good to me. Going to merge now.

davidsneighbour (approved):

PR #1209, opened 2020-08-31
js.Build: Add SourceMap flag into js.Build opts which can turn on sou…

richtera:
Add documentation for sourceMap opt in js.Build. Comment on future external support.
Temporarily tagged with 0.75-DEV


bep:
We can just keep this here, but it is easier to manage this if documentation for unreleased features is added to 

https://github.com/gohugoio/hugo/tree/master/docs

If you want, create a new PR there and use the "expected" tag for the shortcode -- which would be "0.75.0".

jmooring:
Superseded by e50a28fbc77caeeeb785a8340168699c55d56072

PR #1210, opened 2020-09-02
Added link to hugo-embed-pdf-shortcode project.

anvithks:
The hugo-embed-pdf-shortcode project was created to support embedding of PDF files in Hugo pages.
The simple shortcode  was written to embed local files and supports remote files when CORS is enabled. 
Uses PDF.js library for the PDF rendering.


anvithks:
You can check out the shortcode in action [here](https://hugo-embed-pdf.netlify.app/)

bep:
Hey and thanks for this,

I do, however, not think that this reaches the bar of general Hugo relevance/usefulness to be listed on the Hugo site.

I do understand that some of the tools already listed there would already fall into that category, and we should clean that up. The Hugo site is a very popular site, and having all of these links is probably great for **their** SEO, but probably not optimal for the average Hugo user looking for tools he/she can use.

@davidsneighbour we should set some clearer criteria for this. I'm not sure, but for a tool should be listed it should 1) Be a tool built for Hugo and not some general "web tool", it should be relevant to many and be fairly easy to use. For products such as Netlify and Algolia etc. we need a link to some good guide about "how to use it with Hugo" for it to be listed.

anvithks:
@bep Sure thing. 
As you pointed ou,  I just saw the page listed similar shortcodes and hence raised a PR.
I was looking to embed PDFs in a Hugo site and realised there is no shortcode to do that and wrote one for the site I was working on. Thought it might come in handy for others as well.

Will ask to be listed here [https://github.com/theNewDynamic/awesome-hugo](https://github.com/theNewDynamic/awesome-hugo).

Cheers :beers:

davidsneighbour:
Using awesome-hugo for shortcodes is the way to go :+1:

PR #1225, opened 2020-09-16
module import path remove slash at end

yaythomas:
When using import path as given in the docs
```toml
path = "github.com/spf13/hyde/"
```

Hugo fails with : "Error: module "github.com/spf13/hyde/" not found; either add it as a Hugo Module or store it in "mypathhere/themes".: module does not exist"

Remove the trailing slash /, works as expected. 

Hugo version: Hugo Static Site Generator v0.75.1/extended darwin/amd64 BuildDate: unknown


davidsneighbour (approved):
correct. can be merged.

PR #1226, opened 2020-09-17
Doc: Fix typo in hugo command

enguerran:


coliff:
`Hugo serve` is a valid command (alias of Hugo server) so this isn’t a typo as such, but we do use `hugo server` elsewhere so i approve of this change.

coliff (approved):

PR #1228, opened 2020-09-22
Update multilingual.md

sedacrivity:
Just getting into HUGO so reading the docs. 

I noticed that the referenced example for 

If `.ReadingTime.Count` in the context has value is 1

might be wrong.

### DIFF
--- a/content/en/content-management/multilingual.md
+++ b/content/en/content-management/multilingual.md
@@ -383,7 +383,7 @@ Assume `.ReadingTime.Count` in the context has value is 525600. The result will 
 If `.ReadingTime.Count` in the context has value is 1. The result is:
 
 ```
-One minutes to read
+One minute to read
 ```
 
 In case you need to pass a custom data: (`"(dict Count" 25)` is minimum requirment)


davidsneighbour:
Did you test that and it resulted in `1 minutes` or do you think that it should be `minute` for a value of 1 instead of `minutes` as written in the docs? The latter case is reason to open an issue to change it. The first is... just how it is then and the docs are right. I did not test this.

sedacrivity:
See 
https://gohugo.io/content-management/multilingual/#query-a-singularplural-translation

The doc indicates that the I18n file contains a specific sentence with "One minute to read" in case of count 1.

[readingTime]
one = "One minute to read"
other = "{{.Count}} minutes to read"


Thus the so called result of count 1 should match the above i18n.  
I did not test this but imo the doc displays the wrong result.

tomkonidas:
Yes, I ran into the same conclusion!

Either way you look at it, its wrong,

If it was to match on other, then it should print: `1 minutes to read`, however we are seeing `One minutes to read`, This is a combination of `one` and `other`, which I don't believe is possible; It is one or the other.

bep:
This documentation patch is certainly correct. If it behaves differently, which I doubt, then that is another bug.

jmooring:
> If it behaves differently, which I doubt, then that is another bug.

It does not behave differently.

PR #1229, opened 2020-09-22
Missing Translations section update

antonvasilev52:
I suggest to clarify that there is always a "default default language" when speaking about missing translations. A link to the **Configure Languages** section could also be useful.

### DIFF
--- a/content/en/content-management/multilingual.md
+++ b/content/en/content-management/multilingual.md
@@ -465,7 +465,7 @@ The rendering of the main navigation works as usual. `.Site.Menus` will just con
 
 ## Missing Translations
 
-If a string does not have a translation for the current language, Hugo will use the value from the default language. If no default value is set, an empty string will be shown.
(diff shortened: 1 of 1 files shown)


bep:
Isn't there a setting to insert some "missing translation" string? Might be relevant to add.

antonvasilev52:
> Isn't there a setting to insert some "missing translation" string? Might be relevant to add.

 If you mean `enableMissingTranslationPlaceholders = true`, this option is mentioned just a bit below in the same chapter, so this point is clear in the docs.

davidsneighbour (approved):

antonvasilev52:
@bep @davidsneighbour I am sorry if I don't quite understand how it works, but what else must be done to actually merge this change?  
You know the docs are usually 1) relatively easy to change than the code and 2) loose value when are outdated.

jmooring:
See https://github.com/gohugoio/hugo/issues/9216. Neither the existing or proposed documentation matches the current behavior.

bep:
I'm closing this for now, but thanks for the contribution.

Issue #1237, opened 2020-09-25
Always add Hugo version number to title


davidsneighbour:
The release notes on gohugo.io sometimes (often) include a version number (like "Hugo 0.75.1: A couple of Bug Fixes"). Sometimes they don't (like in https://gohugo.io/news/0.75.0-relnotes/). Is there a programmatical way to force @bep ;) to always include the version name in the title? 

I think this could be solved by a Github action that prepares the release news page automatically with links to changesets and pull requests. I don't have a full understanding of how these news are created to attempt this on my own.

This is a usability issue for people that try to find changes between specific versions.

bep:
So, there is an auto-generated title which always includes the version number ... But for main releases I edit the title. And I don't think there is a way to ... force me to to do sensible things.

coliff:
hey @bep - on the subject of the Release Notes, I noticed that the "thanks for the updates to the Documentation site" paragraph is always duplicated each time.

> And a big thanks to @digitalcraftsman for his relentless work on keeping the themes site in pristine condition and to @davidsneighbour, @coliff and @kaushalmodi for all the great work on the documentation site.
> 
> Many have also been busy writing and fixing the documentation in hugoDocs, which has received 24 contributions by 15 contributors. A special thanks to @jmooring, @bep, @jornane, and @inwardmovement for their work on the documentation site.

Maybe you can fix that for the next release. :-)

kaushalmodi:
I agree. I haven't been an active contributor in recent months. My name should be removed from there.

bep:
You could always _contribute_ by creating a PR that fixes that text.

coliff:
Ok, opened a PR to fix that in the 0.75 release notes which I guess you copy and paste each time.

jmooring:
Closing per https://github.com/gohugoio/hugoDocs/pull/1240#issuecomment-699849388, new issue is https://github.com/gohugoio/hugo/issues/8360.

PR #1240, opened 2020-09-28
Fix duplicated 'thanks for documentation contributions' in release notes

coliff:
and removed @kaushalmodi  - as per his request: https://github.com/gohugoio/hugoDocs/issues/1237#issuecomment-699014937


bep:
The duplication comes from autogeneration, so I'd rather fix this in the code.

coliff:
ah, ok - where is that code for us tp update then?

bep:
So, the logic is in this area:

jmooring:
Closing per #1240 (comment), new issue is gohugoio/hugo#8360.

Issue #1227, opened 2020-09-21
gcc is missing from requirements list ?


fvant:
On page https://gohugo.io/getting-started/installing/ , these steps https://gohugo.io/getting-started/installing/#fetch-from-github fail on `go install --tags extended` with an error that `# github.com/bep/golibsass/internal/libsass
exec: "gcc": executable file not found in %PATH%`.

The requirements list in that section only mention Git and Go.

ngdangtu-vn:
Can you tell me what OS you are using? Should I assume it is Windows? (Considering the `%PATH%` pattern, it may be a window) In that case, you must install `gcc` command in order to compile Hugo, because Windows doesn't have it out of the box. Here is a website with [recommended binaries](https://gcc.gnu.org/install/binaries.html). And remember to add the path contains `gcc` command to `%PATH%` so the go compiler could see it.

ngdangtu-vn:
Hello, I've added a note for Windows' case: https://github.com/ngdangtu-vn/hugo-docs/commit/19a96fd906bd3eae345eb11ddb9f122fccb89bd7

Please checkout and lemme know if it is good for PR. @fvant @bep

fvant:
gcc is only required when you use  `--tags extended`, a  `go install` works fine without. Should we recommend gcc for all Windows users ?

davidsneighbour:
Little hair-splitting comment: Is gcc really missing or is it just not in the path?

fvant:
I had not installed it; should be mentioned in requirements anyway imo

ngdangtu-vn:
@davidsneighbour 
`gcc` is like `go`, they are compilers so we must install it manually. The msg error isn't sure if our pc has cpp compiler or not installed properly.

ngdangtu-vn:
I've added few more text: https://github.com/ngdangtu-vn/hugo-docs/commit/507997706ec9e2c68316e0dece5fa49162256b89
I hope it could clear up the some misunderstanding. Please rechecked it again and tell me how you think @fvant @davidsneighbour 

So far, I still wonder should I add `MinGW` as recommendation. On my machine, it worked ok. But does that make the doc turns to commercial product?

PR #1241, opened 2020-09-29
PR for Issue#1227

ngdangtu-vn:
This PR is about:
* [x] additional info for installation guide as request in https://github.com/gohugoio/hugoDocs/issues/1227.
* [x] correct some typos from my previous PR about multilingual
* [ ] investigate the myth: does i18n only support `toml` format in hugo? (One thing I'm sure that it doesn't work with `yaml`, but haven't found a doc about this, still working on it.)

### DIFF
--- a/content/en/content-management/multilingual.md
+++ b/content/en/content-management/multilingual.md
@@ -360,13 +360,13 @@ This article has 101 words.
 
 ### Query a singular/plural translation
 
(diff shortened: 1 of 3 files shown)


bep:
Merged in ed48563aa290d95c5278179c26b824c522875227

Issue #1178, opened 2020-07-24
Document js.Build's externals + test out shimming


bep:
See https://github.com/evanw/esbuild/issues/283

/cc @regisphilibert

UtkarshVerma:
@bep I've tested out shimming and it works. I tried my best with the documentation, but since I'm new to Node.js, there might be some mistakes. I hope this PR  at least serves as a starting point for this issue, and helps out other developers who might be wanting to load external dependencies, like I was.

PR #1246, opened 2020-10-02
Add docs for shimming JS libraries

UtkarshVerma:
This PR is aimed at #1178 . I've tested out shimming with `algoliasearch` and `instantsearch.js`. Since I'm really new with Node.js stuff, so there might be some mistakes with the terminologies here and there.


bep:
Thanks for this, this is very much appreciated! I have been meaning to test this out, but ... time!

PR #1252, opened 2020-10-08
Intend sub levels in table of content

davidsneighbour:
Right now the table of contents is a list of headings each below each other without any visible structure of sub headings. This PR adds a 1rem intendation for sub lists. I am using no `>` like the lines above my change because this way every ul that is inside another ul will receive intending.


coliff:
Great idea. I do a similar thing for a different Hugo-powered documentation site I work on.

coliff (approved):

davidsneighbour:
It's not working, probably due to the `>` rules above. I will rewrite.

coliff:
There's more we can do to improve the Table of Contents styling. The hover effect with the different color background and text changing to white isn't ideal. I'd prefer it if hover text color was darker but background color doesn't change... we could tackle that in a different PR

davidsneighbour:
I am closing this for now. This has to be fixed in gohugoio/gohugoioTheme (the style in static does not seem to be used at all). I can't get compilation in gohugoioTheme running and opened an issue there (gohugoio/gohugoioTheme#176) and will add a pull request there when I get it going.

PR #1257, opened 2020-10-11
docs: Replaced the broken link

saintmalik:
### Description

Easy navigation and easy reading for others.


davidsneighbour (commented):
I advise against this. The name might be the same, but the original by petraszd was a Python App, the new one is a .Net app that works only on Windows. The link should be evaluated as NEW tool that needs verification.

bep:
I'm closing.

PR #1258, opened 2020-10-11
Clarify where the replace command has to go in go.mod

davidsneighbour:
It appears that the `replace` command has to come before the `require` command in a go.mod to work properly.
See https://thewebivore.com/using-replace-in-go-mod-to-point-to-your-local-module/

### DIFF
--- a/content/en/hugo-modules/use-modules.md
+++ b/content/en/hugo-modules/use-modules.md
@@ -81,7 +81,7 @@ Also see the [CLI Doc](/commands/hugo_mod_get/).
 
 ## Make and test changes in a module
 
-One way to do local development of a module imported in a project is to add a replace directive to a local directory with the source in `go.mod`:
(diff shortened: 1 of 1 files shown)


bep:
I'm pretty sure that is not correct.

davidsneighbour:
I am not sure about it either, because samples in the comments to the linked article place the line behind, but the replacement works when I add them before, they don't work when placed after. 

go version go1.15.2 linux/amd64
Hugo Static Site Generator v0.76.2-207913F3/extended linux/amd64 BuildDate: 2020-10-07T09:09:05Z

bep:
So, from memory:

* If `require` is on 1 line, replace comes first
* else: last

Not totally sure about the above, but I'm pretty sure that this PR isn't correct.

davidsneighbour:
That would explain why it didn't work in the beginning. Let's close this then.

PR #1272, opened 2020-11-06
don't hardcode https://gohugo.io/ to site name

agaengel:
don't hardcode https://gohugo.io/ to site name use / to get it working with hugo serve or hugo --gc --minify -b url


bep:
As this theme is used on both themes.gohugo.io and gohugo.io, I don't think this fix will work as intened.

agaengel:
ok I see is it possible to injetct the  baseurl in this case

bep:
@agaengel I don't see how that would work, either.

coliff:
A possible workaround solution (at least so it works on localhost) could be something like:

```html
{{- if not .Site.IsServer -}}
<a href="https://gohugo.io/" class="link white">
{{ else }}
<a href="/" class="link white">
{{ end }}
```

PR #1281, opened 2020-11-17
markdownify: add note about render-hooks and .RenderString

chalin:
Motivation: I still have a hard time remembering the name of `.RenderString`. Ideally, `.RenderString` would appear in the "See Also" section, but that section seems limited to 5 entries. This note at least also mentions why one might consider `.RenderString` as an alternative.

Preview: https://deploy-preview-1281--gohugoio.netlify.app/functions/markdownify/

### DIFF
--- a/content/en/functions/markdownify.md
+++ b/content/en/functions/markdownify.md
@@ -23,3 +23,8 @@ aliases: []
 ```
 {{ .Title | markdownify }}
 ```
+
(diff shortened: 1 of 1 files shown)


chalin:
Any thoughts on this change @jmooring?

chalin:
cc @nate-double-u @zacharysarah

jmooring:
@chalin Thank you. I will commit this change, but I intend to make further edits in the near future---both this and the RenderString page need some more work.

PR #1287, opened 2020-11-25
Clarify that "with" blocks do not render with empty values

amdw:


bep:
I'll merge this even if it's not totally accurate. The true/false is determined by this function:

https://github.com/gohugoio/hugo/blob/master/common/hreflect/helpers.go#L82

Its only difference from the Go implementation is that it also checks for the `IsZero` method.

PR #1288, opened 2020-11-27
Improve substr examples

jmooring:
Partially addresses https://github.com/gohugoio/hugo/issues/7993


jeremielp:
I would particularly like to have the following example in the doc, which is the most common case. 

To extract the N characters starting from the right of the string, use `{{ substr "BatMan" -1 N }}` :
{{ substr "BatMan" -1 3 }} → "Man"
{{ substr "BatMan" -1 4 }} → "tMan"

PR #1292, opened 2020-12-06
Update multilingual.md

robert-jelenic:
Seems like “minutes” should be singular here based on the above logic.


bep:
Sorry for the delay, and you are right, but I'm not able to merge this for some tech reason, so I'll have to close it.

PR #1301, opened 2020-12-28
Update bundling.md

kirillaristov:
Added an extended instruction


bep:
Thanks for this, but since we added the mounts configuration, that is the way we want people to configure asset dirs etc.

PR #1309, opened 2021-01-11
add rel="noopener noreferrer" to insecure links

lpmi-13:
links to external domains using target="_blank" are vulnerable to clickjacking, and `noopener` addresses this. Using `noreferrer` is an addition to address older browsers that don't support `noopener`.


coliff:
thanks for the PR, however this is no longer needed.
Safari and Firefox already imply `noopener` to `_blank` links and Chrome/Edge does too as of v88. (https://www.chromestatus.com/feature/6140064063029248)

lpmi-13:
ah, good to know! Awesome site, btw.

PR #1310, opened 2021-01-17
pipes: babel and js

richtera:
Add sourceMaps to babel docs
Add sourceMap = external to js docs


bep:
Please move these changes into /docs in the hugo repo. That way they get released in synch.

PR #1312, opened 2021-01-23
content/en/functions: remove mention of a file size limit for readFile.

avdva:
The limit itself was removed in [this PR](https://github.com/gohugoio/hugo/pull/8172).


bep:
Could you move this PR to the /docs folder of the Hugo repo? -- then we know it gets published at the right time.

avdva:
here it is: https://github.com/gohugoio/hugo/pull/8173

PR #1315, opened 2021-01-26
Sitemap Protocol 0.9 to 1.0

karabaras:


coliff:
We need to check, but doesn't Hugo ship with Sitemap Protocol 0.9?
https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates.autogen.go

The Hugo site itself has sitemap 0.9.
https://gohugo.io/sitemap.xml

coliff:
The official sitemap protocol is currently 0.9 so we'll close this for now.
https://www.sitemaps.org/protocol.html

PR #1318, opened 2021-01-27
Correct sitemap version

NicoHood:
As correctly stated at the beginning of the docs, sitemap protocol uses version 0.9 currently. Dont confuse with xml version 1.0


coliff (approved):

coliff:
nice - thanks!

PR #1321, opened 2021-02-04
Update page-bundles.md

chmgui:
In the first table under Branch Bundle, the Allowed Resources are "Only non-page (like images, pdf, etc.) types".  
But is this true?    In the example at the bottom of the page: "branch-bundle-1
This branch bundle has the _index.md, two other content Markdown files and two image files.".  Thank you.


bep:
The docs is correct, but I understand your confusion.

* By _bundling_ we mean those files that end up in `.Resources`
* Pages (content files) in a branch bundle end up in `.Pages` and `.RegularPages`.

jmooring:
Closing per https://github.com/gohugoio/hugoDocs/pull/1321#issuecomment-773119327.

PR #1329, opened 2021-02-08
fix deploy-on-github workflows yaml

ziyitony:
There are always problems with the `deploy-on-github` workflow yaml, this PR will fix it.

1. github default branch name is `master` now, not `main`.
2. the `peaceiris/actions-gh-pages` workflow does not have a `latest` tag, `v3` is recommended now.
3. fix some errors on the bottom of the yaml content.

The updated yaml file is tested and works fine.

### DIFF
--- a/content/en/hosting-and-deployment/hosting-on-github.md
+++ b/content/en/hosting-and-deployment/hosting-on-github.md
@@ -55,7 +55,7 @@ name: github pages
 on:
   push:
     branches:
-      - main  # Set a branch to deploy
+      - master  # Set a branch to deploy
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


ziyitony:
For the second point, this is the related issue: https://github.com/peaceiris/actions-gh-pages/issues/566

parsiya:
The `actions-hugo` action also does not have a `latest` tag and we should use `v2`, too.

```
      - name: Setup Hugo
        uses: peaceiris/actions-hugo@latest <--- Should v2 instead of latest
        with:
          hugo-version: 'latest'
          # extended: true
```

I got this error:

```
Unable to resolve action `peaceiris/actions-hugo@latest`, unable to find version `latest`
```

content/en/hosting-and-deployment/hosting-on-github.md
```diff
@@ -55,7 +55,7 @@ name: github pages
 on:
   push:
     branches:
-      - main  # Set a branch to deploy
+      - master  # Set a branch to deploy
```
coliff, line 58: HugoDocs uses 'master', but the default new branch name on GitHub is 'main'.
Main is quickly becoming the standard so I think we should keep it as that.

coliff:
Fixed now with #1332 - thanks !

PR #1331, opened 2021-02-10
Update 404 docs: Netlify auto-detects 404.html

supernrm:
Added information for deploying on Netlify: "This doesn’t require any redirect rules. If you add a 404.html page to your site, it will be picked up and displayed automatically for any failed paths" info from: https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling


bep:
Sorry for the delay, looks like I cannot merge this because the upstream branch has been deleted.

PR #1332, opened 2021-02-11
Fix the GitHub workflow

gagarine:
I tested the documentation on a sample project. Those was the change I had to make in the workflow to make it run. 

- The `@latest` doesn't seem to work and made a fatal error.
- They was twice the "Setup Hugo" action, certainly a bad copy/past.


coliff (approved):

coliff:
Thanks!

PR #1337, opened 2021-02-16
Clean documentation summary on CLI commands page

sebastiaan-lampo:
Fixes https://github.com/gohugoio/hugo/issues/7608

Note that I recommend closing the issue without applying 
this patch as per argument in the issue.

Lack of 'description' frontmatter generated by spf13/cobra
resulted in ugly summary pages. This patch fixes that by
inserting itself before the default summary use with a
custom code snippet extracting the synopsis as summary.

There is no direct access to the content DOM so
string matching is used to identify the start and end
of the synopsis section.

This commit overrides a partial in the gohugoioTheme to 
achieve the intended outcome. As it solves a non-theme  
issue I felt this is the better place.

### DIFF
--- a/content/en/content-management/page-bundles.md
+++ b/content/en/content-management/page-bundles.md
@@ -64,26 +64,27 @@ content/
 In the above example `content/` directory, there are four leaf
 bundles:
 
-about
+`about`
 : This leaf bundle is at the root level (directly under
     `content` directory) and has only the `index.md`.
 
-my-post
+`my-post`
 : This leaf bundle has the `index.md`, two other content
     Markdown files and two image files.
 
-image1
-: This image is a page resource of `my-post`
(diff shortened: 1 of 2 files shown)


jmooring (changes requested):
These changes are not related to the stated problem. Not sure what you did here...

sebastiaan-lampo:
Apparently as a git newbie I incorrectly squashed commits so that my branch for #1336 and this one are identical and I've lost the work for this PR. As it's been several weeks I'd have to go back and redevelop the additional partial that I inserted to fix this. With the planned work on completely overhauling the docs + the fact that this solution was a hack in my opinion I'll close this PR.

I still stand by my recommendation to close the original issue as "won't fix" as well but can't show you the code behind that reasoning.

PR #1338, opened 2021-02-18
Hugo pipes: update asset details

mireille-raad:
- Clarify where assets are stored, for one second I thought /assets is at the root of the project, not theme.
- add link to the config file to make it faster to access

### DIFF
--- a/content/en/hugo-pipes/introduction.md
+++ b/content/en/hugo-pipes/introduction.md
@@ -18,7 +18,7 @@ aliases: [/assets/]
 
 ### Asset directory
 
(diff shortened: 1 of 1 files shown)


coliff:
I'm not sure if this is correct. The default **is** `assets/` in the root of the project.

bep:
The assetDir config option is replaced by mount configuration. It still works, but we should promote the latter.

And the default is `assets` both for themes and project.

mireille-raad:
Sorry for the `asset` vs `assets` typo in the PR 🤦‍♀️ I meant the latter.

I am closing this because you are right, it works in both locations. I don't know why yesterday it wasn't working for me in the root directory. It just worked now.

PR #1342, opened 2021-02-19
Describe the keepComments option for minify

jmooring:


davidsneighbour (commented):
I would use a "new-in" badge (`{{< new-in "0.66.0" >}}`) and use "release-neutral" language (remove the "with a new setting").

coliff:
If the keepComments option was added in Hugo v0.66 then it probably shouldn't be added to the v0.81 release notes.

davidsneighbour:
Nono, I just copypasted the shortcode from somewhere else for reference. The version has to be of course the one where it was added.

jmooring:
Why would we add a "new-in 0.81.0" badge to the release notes for 0.81.0? This is redundant.
Why use "release-neutral" language in release notes, that are by definition release-specific?

coliff:
We should also really mention the new JavaScript `keepVarNames` minification setting too then.
https://gohugo.io/getting-started/configuration/#configure-minify

jmooring:
@coliff There were a number of changes. See:
<https://github.com/gohugoio/hugo/commit/1b364b003f68df3adb1644769fe69810d85e3897>

I'm not sure if the change from "decimals" to "precision" is a breaking change or not.

I documented the one change that I actually knew something about.

Regardless, if you want to further improve the release notes for 0.81, I suggest opening another issue because this one is closed.

PR #1344, opened 2021-02-22
Function sort: fix example labels

chalin:
Preview: https://deploy-preview-1344--gohugoio.netlify.app/functions/sort/

### DIFF
--- a/content/en/functions/sort.md
+++ b/content/en/functions/sort.md
@@ -22,22 +22,22 @@ aliases: []
 A sorted array of map values will be returned with the keys eliminated. There are two optional arguments: `sortByField` and `sortAsc`. If left blank, sort will sort by keys (for maps) in ascending order as its default behavior.
 
 ```
-+++
-keywords: [ "tag3", "tag1", "tag2" ]
-+++
+---
+tags: ["tag3", "tag1", "tag2"]
+---
 
 // Site config
 +++
 [params.authors]
-  [params.authors.Derek]
-    "firstName"  = "Derek"
-    "lastName"   = "Perkins"
   [params.authors.Joe]
-    "firstName"  = "Joe"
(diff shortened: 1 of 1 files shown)


jmooring:
@chalin, the sample data on this page is wrong as well. Please update it to:

```
+++
tags = ["tag3", "tag1", "tag2"]
+++

// Site config
+++
[[params.authors]]
  id = "003"
  firstName = "Derek"
  lastName = "Perkins"
[[params.authors]]
  id = "001"
  firstName = "Joe"
  lastName = "Bergevin"
[[params.authors]]
  id = "002"
  firstName = "Tanner"
  lastName = "Linsley"
+++
```

Without these changes the sample code fails or provides the wrong results.

jmooring:
See <https://github.com/gohugoio/hugo/issues/8335>.

chalin:
@jmooring - thanks for your feedback. You beat me to adding the reference to issue https://github.com/gohugoio/hugo/issues/8335 :).

I fixed the list field name, changing it to `tags`.

As for `authors`, I believe that the example was meant to illustrate the sorting of a map. I've kept the map value, but swapped the first two entries.

Can we get this merged now and I'll revisit if/once https://github.com/gohugoio/hugo/issues/8335 is addressed?

chalin:
cc @nate-double-u @zacharysarah

jmooring:
Thoughts:

1. The documentation is currently wrong.
2. If we merge this the documentation will still be wrong, but a little less so.
3. gohugoio/hugo#8335 may never be addressed, in which case the documentation will remain wrong.

I'm OK merging this if you first create another issue in this repo that references gohugoio/hugo#8335. I don't want to forget that we have a documentation problem.

chalin:
Done: #1365

PR #1345, opened 2021-02-24
Update quick-start.md  to add explanation for -D where it is used

subsoap:
The inclusion for -D without explanation confused me as to why drafts were still being included in public. I may be dumb, but this has probably confused others / will continue to confuse more in the future without the extra note.

### DIFF
--- a/content/en/getting-started/quick-start.md
+++ b/content/en/getting-started/quick-start.md
@@ -175,5 +175,7 @@ It is simple. Just call:
 hugo -D
 ```
 
+> Note that -D will include posts marked as drafts when building and is not required to build. See: https://gohugo.io/commands/hugo/#options
+
 Output will be in `./public/` directory by default (`-d`/`--destination` flag to change it, or set `publishdir` in the config file).
 


jmooring:
I'm inclined to close this. The documentation already includes an explanation immediately before instructing the user to type `hugo -D`:

> Now, start the Hugo server with [drafts](https://gohugo.io/getting-started/usage/#draft-future-and-expired-content) enabled:

subsoap:
I would personally remove the -D from the step 7 to make it clearer to new users that it's not necessary. Otherwise I would still include a short explanation at that step if it must be included, such as linking to all of the possible options.

> Now, start the Hugo server with [drafts](https://gohugo.io/getting-started/usage/#draft-future-and-expired-content) enabled:

I would change this in step 5 to

> Now, start the Hugo server with [drafts](https://gohugo.io/getting-started/usage/#draft-future-and-expired-content) forced to be enabled (-D tells Huge to not ignore all posts marked as drafts):

jmooring:
With a new installation of Hugo, the value of "draft" in the default archetype is `true`. The Quick Start user may or may not change the value after creating the content, so `--buildDrafts` or `-D` (easier to type) must be included, because a tutorial like this should just work.

subsoap:
If it's necessary then why not take another chance to better explain it?

The tutorial as is confused me, it will confuse more people in the future too. It's not a math book, the tutorial should be as clear as possible with multiple points of reinforcement when necessary.

PR #1358, opened 2021-03-13
Update syntax-highlighting.md: add link of Chroma

CarsonSlovoka:
Added a link to [Chroma](https://github.com/alecthomas/chroma) to help more people realize this is a go development project  [#8326](https://github.com/gohugoio/hugo/issues/8326#issuecomment-798516268)

### DIFF
--- a/content/en/content-management/syntax-highlighting.md
+++ b/content/en/content-management/syntax-highlighting.md
@@ -125,7 +125,7 @@ The options are the same as in the [highlighting shortcode](/content-management/
 
 ## List of Chroma Highlighting Languages
 
-The full list of Chroma lexers and their aliases (which is the identifier used in the `highlight` template func or when doing highlighting in code fences):
+The full list of [Chroma](https://github.com/alecthomas/chroma) lexers and their aliases (which is the identifier used in the `highlight` template func or when doing highlighting in code fences):
 
 {{< chroma-lexers >}}
 


jmooring:
Thanks for this, but the first paragraph of this page already provides a link. 

> Hugo uses [Chroma](https://github.com/alecthomas/chroma) as its code highlighter; it is built in Go and is really, really fast -- and for the most important parts compatible with Pygments we used before.

I see no reason to link to the same source twice on the same page.

CarsonSlovoka:
Oh, I didn't realize that there was an introduction at the beginning😅

How about removing the [List of Chroma Highlighting Languages' shortcodes {{< chroma-lexers >}}](https://github.com/gohugoio/hugoDocs/blame/8e82c7ce1e6ca03573b060b35acf800a12908761/content/en/content-management/syntax-highlighting.md#L130) and replacing it with a link to Chroma's[supported languages]?

I feel it is smaller, more focused, and easier to read.

image


it's a great idea to create an alias, but when Chroma is updated (e.g., a new highlight is added), you must update the [data.json] again.

Is it possible to suggest directly in Chroma to add the contents of [supported languages] to the alias, then the link will be the same as Hugo's original intention?

If this is feasible, I can send an issue in Chroma to ask the author if he agrees to such a change before deciding on the subsequent behavior.

[data.json]: https://github.com/gohugoio/hugoDocs/blob/master/data/docs.json
[supported languages]: https://github.com/alecthomas/chroma#supported-languages

jmooring:
The data.json file is generated automatically when you build Hugo from source. See:
https://github.com/gohugoio/hugo/blob/master/helpers/docshelper.go

That means that the resulting list of lexers represents the version of Chroma in the _current_ Hugo build. We do _not_ want to list lexers that are in the current (HEAD) version of Chroma; it may contain items not yet supported by Hugo.

jmooring:
@CarsonSlovoka, I'm closing this. Open a separate PR if you want to update the lexer list styling at some point.

PR #1359, opened 2021-03-14
fix: base64-encoded inline image example

thiagovilla:


content/en/content-management/page-resources.md
```diff
@@ -38,7 +38,12 @@ RelPermalink
 : The relative URL to the resource. Resources of type `page` will have no value.
 
 Content
-: The content of the resource itself. For most resources, this returns a string with the contents of the file. This can be used to inline some resources, such as `<script>{{ (.Resources.GetMatch "myscript.js").Content | safeJS }}</script>` or `<img src="{{ (.Resources.GetMatch "mylogo.png").Content | base64Encode }}">`.
```
davidsneighbour, line 41: Is `` false? Your sample adds a data and base64 part in front of the encoded part, doesn't base64Encode add the data part itself?

If base64Encode does not do this then the documentation at https://gohugo.io/functions/base64/ needs rewriting too.

jmooring: > doesn't base64Encode add the data part itself?

No, it does not.

> If base64Encode does not do this then the documentation at https://gohugo.io/functions/base64/ needs rewriting too.

No, it does not.

jmooring (changes requested):
Please test your code before including it in the documentation; your sample code fails because you have not set the context for `.MediaType` or `.Content`. I suggest:

```
{{ $img := .Resources.GetMatch "mylogo.png" }}
<img src="data:{{ $img.MediaType }};base64,{{ $img.Content | base64Encode }}">
```
```

jmooring:
Superseded by #1587

PR #1380, opened 2021-03-24
Remove Blackfriday reference from shortcode desc

jmooring:


davidsneighbour (commented):
Can we please use a . instead of a semicolon? That's two different sentences. Easier to read.

PR #1382, opened 2021-03-25
Improve presentation of template lookup order

### LINKED ISSUE #1374: improve Template Lookup Order presentation
Below is a screenshot of how Template Lookup Order is displayed, which I find very difficulty to read and use. 

Options for improvement might include
- using HTML lists (eg. `<ol><li>`) to show `Template Lookup Order` items
- presenting these items in a `tree` format
- at least separating items with commas...

image
> https://gohugo.io/templates/lookup-order/#examples-layout-lookup-for-regular-pages

jmooring:
Closes #1374


jmooring:
This is what will look like:

PR #1388, opened 2021-03-29
Use code-toggle shortcode in `multilingual.md`

salim-b:


bep:
So, the netlify preview build fails indicating something wrong with that new code block -- but I cannot spot the error. Does this build locally for you?

salim-b:
I've only tried to build locally now. My findings:

- Initially the site doesn't build but errors with

   > Error: add site dependencies: create deps: 1 error(s) decoding:
   >
   > * 'Goldmark.Parser.Attribute' expected type 'bool', got unconvertible type 'map[string]interface {}'

   This is because I'm still on Hugo 0.80 (extended snap package) but the `Goldmark.Parser.Attribute` config changed with 0.81, i.e. it got split into subkeys.
   
   The relevant config entry is found in `config/_default/markup.toml`:

    ```toml
    [goldmark.parser.attribute]
    block = true
    title = true
    ```

    The specified `module.hugoVersion.min = "0.56.0"` in `./config.toml` is thus a bit misleading...
    
- As soon as I comment out the `goldmark.parser.attribute` config mentioned above, the site builds fine. My changes are rendered as expected :)

salim-b:
> So, the netlify preview build fails indicating something wrong with that new code block -- but I cannot spot the error. Does this build locally for you?

The Netlify preview for this PR [seems fine](https://deploy-preview-1388--gohugoio.netlify.app/), actually. I guess you meant to comment on https://github.com/gohugoio/hugoDocs/pull/1389 instead?

jmooring:
Yes, this is fine. Thank you.

PR #1393, opened 2021-03-31
Harmonize capitalization

salim-b:
Harmonize capitalization between `pagination.md` and `configuration.md`.

Hugo normalizes site configuration keys to lowercase internally, so technically, capitalization doesn't really matter. But it's less confusing for newbies if the capitalization in the documentation is consistent.


davidsneighbour (approved):
makes sense and in my project they are all starting lowercased, so it's working as changed.

PR #1399, opened 2021-04-01
Fix wrong example

salim-b:


coliff:
duplicate of #1228 - can you please take a look at that PR?

PR #1401, opened 2021-04-03
Fix `dateFormat` documentation

salim-b:
Please countercheck my changes – I _think_ `dateFormat` cannot return its `INPUT` "as a Go `time.Time` type value" as the documentation originally suggested. This is what `time` does.

### DIFF
--- a/content/en/functions/dateformat.md
+++ b/content/en/functions/dateformat.md
@@ -17,7 +17,7 @@ relatedfuncs: [Format,now,Unix,time]
 deprecated: false
 ---
 
-`dateFormat` converts an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp string `INPUT` into the format specified by the `LAYOUT` string.
+`dateFormat` converts a timestamp string `INPUT` into the format specified by the `LAYOUT` string.
 
 ```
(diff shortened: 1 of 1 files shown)


content/en/functions/dateformat.md
```diff
@@ -17,7 +17,7 @@ relatedfuncs: [Format,now,Unix,time]
 deprecated: false
 ---
 
-`dateFormat` converts the textual representation of the `datetime` into the specified format or returns it as a Go `time.Time` type value. These are formatted with the layout string.
+`dateFormat` converts an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp string `INPUT` into the format specified by the `LAYOUT` string.
```
bep, line 20: You are right about the return value (a string), but INPUT does not need to be ISO 8601 -- we support a bunch of different date/time formats, so if it looks like a date, it should work.

bep (changes requested):

content/en/functions/dateformat.md
salim-b: Perfect, thanks. Should be fine now.

coliff (approved):

PR #1403, opened 2021-04-04
Add native image lazy-loading example

NathanLovato:
Follow-up to #1402 

This shows how to get images' width and height in markdown render hooks, for example, for lazy loading, or could be to add a low-res placeholder image and later replace it with a high-res one (as seen on sites like Medium).

Depending on the case, you need to use either imageConfig or image resources.

I'm unsure this is the kind of contribution you're looking for in the docs, but if not, all good, just let me know.

### DIFF
--- a/content/en/getting-started/configuration-markup.md
+++ b/content/en/getting-started/configuration-markup.md
@@ -216,6 +216,49 @@ Here is a code example for how the render-image.html template could look:
 </p>
 {{< /code >}}
 
+#### Image with native lazy-loading Markdown example:
+
+Here is a code example to use [browser-native image lazy-loading](https://css-tricks.com/native-lazy-loading/).
+
+When using this feature, you want to set your images' `width` and `height` attributes to avoid layout shifts while loading.
+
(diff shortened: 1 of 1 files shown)


coliff:
This looks like it could be a useful example. It's good to always provide the width and height attributes on images to prevent CLS (content layout shifting). However your example doesn't work if any of the images are hotlinked to a 3rd party site. It'd be neat if it would also attempt to add width and height attributes for locally hosted images.

coliff:
This worked for me, though I'm certain there is a tidier way of doing it. (not sure how to write if string does **not** have prefix http)

```
<img src="{{ .Destination | safeURL }}" alt="{{ .Text }}" decoding="async" loading="lazy"
  {{ if strings.HasPrefix .Destination "http" }}{{else}} {{ with imageConfig ( printf "static/%s" .Destination ) }}
  width={{ .Width }} height="{{ .Height }}" {{ end }}{{ end }}>
```

jmooring:
Closing. Documentation for markdown render hooks was moved to another page quite a while ago, and the proposed examples have issues.

PR #1405, opened 2021-04-06
Rework robots.txt page

### LINKED ISSUE #1404: Description of robots.txt override is incorrect
According to the documentation:

<https://github.com/gohugoio/hugoDocs/blob/b66567de20dafe690640f1d5c308371175164539/content/en/templates/robots.md#robotstxt-template-lookup-order>

> The lookup order for the robots.txt template is as follows:
> 
> * /layouts/robots.txt
> * /themes/<THEME>/layouts/robots.txt
> > If you do not want Hugo to create a default robots.txt or leverage the robots.txt template, you can hand code your own and place the file in static. Remember that everything in the static directory is copied over as-is when Hugo builds your site. 

But this is not happening:

```
> ls public
ls: public: No such file or directory
> cat static/robots.txt 
User-agent: *
Disallow:

Sitemap: /sitemap.xml
> cat themes/anubis/layouts/robots.txt 
User-agent: *
# robotstxt.org - if ENV production variable is false robots will be disallowed.
{{ if eq (getenv "HUGO_ENV") "production" | or (eq .Site.Params.env "production")  }}
  Disallow:
{{ else }}
  Disallow: /
{{ end }}
> hugo
Start building sites … 

                   | EN   
-------------------+------
  Pages            | 560  
  Paginator pages  |  64  
  Non-page files   |   0  
  Static files     | 126  
  Processed images |   0  
  Aliases          | 167  
  Sitemaps         |   1  
  Cleaned          |   0  

Total in 665 ms
> cat public/robots.txt 
User-agent: *
# robotstxt.org - if ENV production variable is false robots will be disallowed.

  Disallow: /
```

Hugo is copying the wrong robots.txt file.

### What version of Hugo are you using (`hugo version`)?

<pre>
> hugo version
hugo v0.81.0-59D15C97 darwin/amd64 BuildDate=2021-02-19T17:07:12Z VendorInfo=gohugoio
</pre>

### Does this issue reproduce with the latest release?

yes

jmooring:
Fixes #1404

### DIFF
--- a/content/en/templates/robots.md
+++ b/content/en/templates/robots.md
@@ -18,37 +18,49 @@ aliases: [/extras/robots-txt/]
 toc: false
 ---
 
-To create your robots.txt as a template, first set the `enableRobotsTXT` value to `true` in your [configuration file][config]. By default, this option generates a robots.txt with the following content, which tells search engines that they are allowed to crawl everything:
+To generate a robots.txt file from a template, change the [site configuration][config]:
 
-```
+{{< code-toggle file="config">}}
+enableRobotsTXT = true
+{{< /code-toggle >}}
+
+By default, Hugo generates robots.txt using an [internal template][internal].
+
+```text
 User-agent: *
 ```
+
+Search engines that honor the Robots Exclusion Protocol will interpret this as permission to crawl everything on the site.
 
 ## Robots.txt Template Lookup Order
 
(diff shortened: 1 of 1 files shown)


davidsneighbour:
for some reasons I don't receive review requests... but I approve ;)

PR #1407, opened 2021-04-07
Update internal.md

thenktor:
Info about automatically used opengraph image metadata added. See source in https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/opengraph.html

### DIFF
--- a/content/en/templates/internal.md
+++ b/content/en/templates/internal.md
@@ -152,6 +152,7 @@ tags = []
 
 Hugo uses the page title and description for the title and description metadata.
 The first 6 URLs from the `images` array are used for image metadata.
+Furthermore, if [page bundles](/content-management/page-bundles/) are used and no `images` array is defined, images with filenames containing `*feature*` or `*cover*,*thumbnail*` are automatically used for image metadata.
 
 Various optional metadata can also be set:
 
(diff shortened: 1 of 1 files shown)


davidsneighbour:
Is this true? I don't think so. You will have to use an internal template in your templates to make that happen and many themes do not use internal templates.

thenktor:
> 
> 
> Is this true? I don't think so. You will have to use an internal template in your templates to make that happen and many themes do not use internal templates.

But that is what the doc page is about: the internal templates. If you use them, it works that way.

content/en/templates/internal.md
```diff
@@ -152,6 +152,7 @@ tags = []
 
 Hugo uses the page title and description for the title and description metadata.
 The first 6 URLs from the `images` array are used for image metadata.
+Furthermore, if [page bundles](/content-management/page-bundles/) are used and no `images` array is defined, images with filenames containing `*feature*` or `*cover*,*thumbnail*` are automatically used for image metadata.
```
jmooring, line 155:
```suggestion
If [page bundles](/content-management/page-bundles/) are used and the `images` array is empty or undefined, images with filenames matching `*feature*` or `*cover*,*thumbnail*` are used for image metadata.
```

PR #1413, opened 2021-04-11
Added one extension to the list

andreseduardop:
Added Hugo Snippets to the list of Sublime Text plug-ins.


davidsneighbour (commented):
why does TOC change from false to true?

andreseduardop:
> why does TOC change from false to true?

Sorry. I was wrong on this.
I created a new PR, without this error.

PR #1423, opened 2021-04-17
description: only list pages can have children

sjjh:
origin of change: https://discourse.gohugo.io/t/front-matter-for-nested-menu-does-not-work/32413/6


jmooring:
"Only list pages can have child pages." This is not true. I'm going to close this PR, then comment on the forum topic.

PR #1426, opened 2021-04-18
Add more Chinese translations of documentation

zoxoy:
I just translated some posts of documentation in this PR


coliff:
Hello, thanks for your PR and sorry for the long delay in replying!

As of January 1, 2023, we will no longer host translations of the Hugo documentation. Currently there is only one translation, and it is incomplete with almost no content.

Similar to many other open-source projects, if community members create their own translated sites, we may choose to include links to those sites from [gohugo.io](https://gohugo.io/).

Thank you for being an active member of the Hugo community.

PR #1427, opened 2021-04-20
Accessibility and SEO improvements

jcabak:
- Links have `title` attribute and are accessible to everyone


coliff:
Thanks for the PR but this doesn't look like a good idea to me. Adding the title of the link when the full title is already displayed doesn't help anyone and would be an annoyance to see the tooltip with the same text on it.

davidsneighbour:
It's not accessibility, because a screen reader would read the title AND the link text. It's not SEO either because here too both have the same content and no extra information is added.

In 2021 accessibility would mean to add `aria-label` to links that have no identifiable description (like icon links to social media). A link with text that is self described is enough.

PR #1428, opened 2021-04-21
Warning: do not use config dir in nested modules

yaaax:
"the config directory is only read for the main project. Themes/theme modules is config.toml etc. only."
@see https://discourse.gohugo.io/t/hugo-theme-module-with-nested-module/28199/6


jmooring:
This was resolved with https://github.com/gohugoio/hugo/pull/8664

PR #562, opened 2018-08-11
Add Hugo wrapper

khos2ow:
This PR is to add [hugo-wrapper](https://github.com/khos2ow/hugo-wrapper), a universal way to use Hugo binary anywhere:

- to `tools/starter-kits` page
- to the whole repository itself (as a showcase of how it works)

PR #1429, opened 2021-04-22
Create hosting-on-hostman

alena-ko:
Added a new deployment option for Hugo websites

### DIFF
--- a/content/en/hosting-and-deployment/hosting-on-hostman
+++ b/content/en/hosting-and-deployment/hosting-on-hostman
@@ -1,0 +1,85 @@
+---
+title: Host on Hostman
+linktitle: Host on Hostman
+description: Host your Hugo site for free with Hostman's global CDN, fully-managed SSL, SSH and auto deploys from your repository.
+date: 2021-04-22
+publishdate: 2021-04-22
+lastmod: 2021-04-22
+categories: [hosting and deployment]
+keywords: [hostman,hosting,deployment]
+menu:
+  docs:
+    parent: "hosting-and-deployment"
+    weight: 10
+weight: 10
+sections_weight: 10
+draft: false
+aliases: []
+toc: true
+---
+
+[Hostman](https://hostman.com/) is a cloud hosting provider for startups and new projects. It helps to get rid of most routine DevOps operations, saving time for developers and money for companies. Hostman employs a services concept to make it easier to develop complex architecture and scale it in one click.
+
+Deploy static websites **for free** with Hostman. All the following features included:
+
+- Hostman automates CI/CD, pulling, building and launching code as soon as you push a new commit to the repository.
+- Everything is transparent: you see the actual hardware your application is operating on and the actual load average, so you can assess them if something goes wrong. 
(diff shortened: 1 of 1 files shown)


coliff:
For reference:
Alexa rank: #562,946
https://www.alexa.com/siteinfo/hostman.com

@hugo-docs team: 
We already have a lot (too many?) hosting guides for companies and can't add guides for every hosting company. Perhaps we should define a minimum alexa rank for submissions....

jmooring (commented):
TLDR: I recommend closing this PR.

Findings:
- A [similar PR](https://github.com/jekyll/jekyll/pull/8652) was submitted to the Jekyll project today. Although these PR's may have been submitted in good faith in an effort to assist these communities, they smell a bit like requests for advertising.
- I searched the [Hugo forum](https://discourse.gohugo.io/) for "hostman" and found nothing, so it doesn't look like anyone has asked about this.
- I also searched for issues and PR's (both open and closed) related to "hostman" and found nothing.

This seems like a solution without a problem, where Hostman is the primary beneficiary, at least initially.

I have no data to support this, but I would not be surprised to learn that 90% of Hugo deployments are handled by fewer than 10 methods, one which is a generic deployment (rsync/ftp). If I am correct, let's make sure we document the other 9, and say no to everything else. If someone has a question about how to deploy a site to "my-super-free-fast-hosting.com" they can (a) contact the provider, or (b) create a forum topic.

bep:
Hey, and thanks for this.

I think I'm saying that we are currently not adding any more "deployment options/guides" to the documentation site unless it can provide an obvious and useful Hugo link. We need to take a step and look at the deployment options/guides we already have -- I'm starting to feel that we have become this marketing site for new host options. We should probably trim down what we have, but let's start by taking a pause with adding others.

alena-ko:
Hey guys,

great to get some feedback, even if it's not as nice as could be expected.

I'm Helen, the author of this PR and a part of Hostman team. I'll take a
bit of your time to clarify my purpose for adding this info and for
bothering you in general.

So Hostman is an application hosting provider that hosts static sites for
free. We've launched two years ago, but we still are a startup. Right now
we have more than 4K active clients and are growing by 10% per month (90%
of our clients have static websites). The reason why I reached you is that
we want developers to host and deploy their projects easily and want them
to have a nice choice of providers that would help to spread sites
developed on Hugo throughout the web faster. We understand that there are
already a good number of providers to choose from, but from our side, we
offer our users some cool goodies like free and fast deployment,
pull-request previews (that many competitors don't provide), out-of-the-box
CI/CD, permanent storages (that many competitors don't provide also) and so
on.

As you can see, Hostman is not another my-super-free-fast-hosting.com provider
as you may think and that PR was not a part of our advertisement. It can
rather be called the popularization of the Hugo framework and an effort to
make it more convenient for your users.

Anyway, the final choice is up to you.

With hopes to build a fruitful partnership with you,
Helen from Hostman

чт, 22 апр. 2021 г. в 19:29, Christian Oliff ***@***.***>:

PR #1435, opened 2021-04-28
Added dartsass install to build with Github Action

faraixyz:


bep:
Thanks for this, but I'm not sure we want to add this. It's useful, but as it is it would need some basic checksum validation to be safe(r), and I'm not sure we want to maintain that level of details in the docs.

Looking at the example GitHub action, it would probably be smarter/easier/better if it ran in a container with all of these tools (Hugo, Dart Sass) built in. E.g. @klakegg 's container.

faraixyz:
Fair enough. I'll see if I can add it to the Hugo installer used in the docs.

PR #1442, opened 2021-05-01
Use Hugo version badge shortcode

salim-b:


salim-b:
I've noticed only now that the shortcode is not rendered because the Markdown file containing it was included using the `readfiles` shortcode. Therefore I've submitted https://github.com/gohugoio/hugoDocs/pull/1444 that should properly fix this type of issue and allow use shortcodes in headless bundles, too.

PR #1444, opened 2021-05-01
Add `getcontent` shortcode

salim-b:
This is an alternative to the [`readfiles`](https://github.com/gohugoio/hugoDocs/blob/master/layouts/shortcodes/readfile.html) shortcode that is better suited to insert [headless](https://gohugo.io/content-management/page-bundles/#headless-bundle) page content since the inserted content can contain shortcodes itself. It's inspired by [this answer on Discourse](https://discourse.gohugo.io/t/using-hugo-shortcodes-in-readfile/20161/4).

This new shortcode is used to fix a regression introduced in https://github.com/gohugoio/hugoDocs/pull/1442 (`new-in` shortcode was not rendered as I intended).

If you agree, I can also update all other occurences of Markdown file snippet inclusion to use `getcontent` instead of `readfile`.

### DIFF
--- a/layouts/shortcodes/insertpages.html
+++ b/layouts/shortcodes/insertpages.html
@@ -4,7 +4,7 @@ Insert page `.Content` from a (headless) bundle; you can insert `.Content` from 
 Usage: {{< insertpages path="PATH/TO/MD_FILE" >}}
        {{< insertpages path="PATH/TO/BUNDLE" glob="*_PATTERN.md" >}}
 */}} -->
-{{ $path := .Get "path" -}}
+{{- $path := .Get "path" -}}
 {{ $glob := .Get "glob" -}}
 
 {{ $scratch := newScratch -}}
--- a/layouts/shortcodes/insertpages.html
+++ b/layouts/shortcodes/insertpages.html
@@ -1,0 +1,21 @@
+<!-- {{/*
+Insert page `.Content` from a (headless) bundle; you can insert `.Content` from multiple page resources of the same bundle by specifying `glob`
+
+Usage: {{< insertpages path="PATH/TO/MD_FILE" >}}
+       {{< insertpages path="PATH/TO/BUNDLE" glob="*_PATTERN.md" >}}
+*/}} -->
+{{ $path := .Get "path" -}}
+{{ $glob := .Get "glob" -}}
+
+{{ $scratch := newScratch -}}
+{{ with $glob -}}
+  {{ $bundle := site.GetPage (path.Join $path "index.md") -}}
+  {{ $scratch.Set "pages" ($bundle.Resources.Match $glob) -}}
+{{ else -}}
+  {{ $bundle := site.GetPage (path.Join (path.Dir $path) "index.md") -}}
+  {{ $scratch.Set "pages" ($bundle.Resources.Match (path.Base $path)) -}}
+{{ end -}}
+
+{{ range ($scratch.Get "pages") -}}
+  {{ .Content }}
+{{ end -}}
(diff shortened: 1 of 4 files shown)


layouts/shortcodes/insertpages.html
```diff
@@ -0,0 +1,21 @@
+<!-- {{/*
+Insert page `.Content` from a (headless) bundle; you can insert `.Content` from multiple page resources of the same bundle by specifying `glob`
+
+Usage: {{< insertpages path="PATH/TO/MD_FILE" >}}
+       {{< insertpages path="PATH/TO/BUNDLE" glob="*_PATTERN.md" >}}
+*/}} -->
```
bep, line 6: This looks mostly good, some comments:

*  You don't need to use `Scratch` for this, just declare your variable with `slice` outside of the with
* The best path to use in `site.GetPage` is the base path to the file/bundle without file suffix, e.g. "blog/mybundle" -- so no need to add "index.md"
* Could you rename the shortcode to `getpagecontent` ... I get that you support multiple matches, which is good, but "insertpages" is a little misleading.

bep (changes requested):

layouts/shortcodes/insertpages.html
salim-b: Thanks @bep for your comments!

- I've removed the `index.md` suffixes as suggested.

- I've renamed the shortcode to `getcontent`. I think it's better to not mention "page" since it allows to include the `.Content` from any page bundle resource, not only those of type `page`. But if you still prefer `getpagecontent`, just let me know...

- I can't really follow how I could replace `Scratch` with `slice`... can you post some code?

bep:
```html
{{ $resources := slice }}
{{ if true }}
{{ $resources = .Resources.Match "foo" }}
{{ else }}
// set resources to something else 
{{ end }}

{{ range $resources }}

{{ end }}
```

salim-b: Thank you! I simply assumed this wouldn't work... I obviously lack the necessary Golang knowledge, but out of curiosity: Is there a simple explanation why the scoping issue described in https://github.com/golang/go/issues/10608 doesn't apply here?

Anyway, I've updated the PR to avoid the scratch usage. Feel free to merge!

And another question: Do you think it would make sense to add this shortcode to [Hugo's built-in ones](https://gohugo.io/content-management/shortcodes/#use-hugos-built-in-shortcodes)? It's not specific to HugoDocs and personally I'd really like to have this shortcode available out-of-the-box in every Hugo theme...

bep: The issue you reference was created to fix the scoping issue in question. The above wasn't possible in older version of Go.

salim-b: I see, so the [`.Scratch` documentation](https://gohugo.io/functions/scratch) should be updated, I guess.

In order to do so, some questions that arise:

1. What are the remaining advantages of `.Scratch`, now that the original scoping issue in Go is fixed? Is it that `.Scratch`'s scope "spans partial and shortcode includes" and Go's regular template vars don't?

2. What's the minimal Hugo version of which the prebuilt binaries are compiled using a Go version that doesn't suffer the original scoping issue?

jmooring: The `=` assignment token was added to Go with https://github.com/golang/go/commit/28c1ad9d35f27b3b57afff4ee78faac746a8ed0a and available in Hugo [v0.48](https://github.com/gohugoio/hugo/releases/tag/v0.48).

This article includes example of usage:
<https://regisphilibert.com/blog/2017/04/hugo-scratch-explained-variable/>

salim-b: See https://github.com/gohugoio/hugoDocs/pull/1451 :)

PR #1450, opened 2021-05-03
repair menu position on mobile devices

lupsa:
When media screen is under 30rem, the menu and menu docs doesn't have padding on top and the top item from menu docs isn't visible


bep:
Any changes to the theme (what is in the _vendor folder) needs to be applied here: https://github.com/gohugoio/gohugoioTheme/

I may have accidently merged these changes in this repo, but those will be overwritten.

PR #1451, opened 2021-05-04
Overhaul scratch.md

salim-b:
I've overhauled the outdated `.Scratch` documentation based on [this short discussion](https://github.com/gohugoio/hugoDocs/pull/1444#discussion_r625086819) and @regisphilibert's enlightening [blog post](https://regisphilibert.com/blog/2017/04/hugo-scratch-explained-variable/) about the subject.

@regisphilibert: I've copy-pasted a small example from your post. Hope you're fine with that. Of course you're feedback on this PR is very welcome.

### DIFF
--- a/content/en/functions/scratch.md
+++ b/content/en/functions/scratch.md
@@ -20,7 +20,7 @@ draft: false
 aliases: [/extras/scratch/,/doc/scratch/]
 ---
 
-`.Scratch` is available as methods on `Page` and `Shortcode`.
+`.Scratch` is available as a method on `Page` and `Shortcode`.
 
 It was initially created as a workaround to fight a [Go template scoping limitation](https://github.com/golang/go/issues/10608) which prevented template variable overwrites. Starting with Hugo 0.48, this limitation was lifted and the use of `Scratch` isn't necessary anymore in those cases.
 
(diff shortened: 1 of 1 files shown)


regisphilibert:
Thanks for this! It's up for discussion, but as Scratch being attached to a particular page has always been a limitation to my sense, I now mostly use to efficiently manipulate unique data. For this I always create a `newScratch` scoped to the template file I'm working on (partials, returning partials etc...).

My point, is maybe the `newScratch` function should be introduced earlier to illustrate the fact that Scratch really does not have to be bound to a page. The only reason to do this, is to somehow store data to a page and use it later on the layouts but this can create problems depending in which orders pages are treated by Hugo. It also forces you to systematically ensure the Page context is available from any given partial so that `.Scratch` is available. 

So I think while you are overhauling the scratch doc, it would be a good opportunity to encourage users to move away from systematically using scratch as a Page method and get used the `newScratch` instead.

I would include in the introduction the fact that Scratch can either a native Page method or stored in a variable with explicit examples:

```
{{ $s := newScratch }}
{{ $s.Set "greeting" "bonjour" }}
VS
{{ .Scratch.Set "greeting" "bonjouer"
```

Then move `newScratch` at the top of the methods list

```
{{ $s := newScratch }}
```

And then use `$s.` in subsequent method examples. (or whatever variable names you see fit)

salim-b:
Thanks @regisphilibert for your feedback! I've made some changes based on your comment.

But regarding the following two points, I feel I lack the knowledge / competence to decide on what to write best:

1. Should we actually discourage the use of the `Page`/`Shortcode`-bound `Scratch`? And completely focus on locally scoped `newScratch` as you suggest?

2. Performance (build time): I have no clue here, i.a. differences between `Scratch` and normal template vars. Maybe you can add something about this in a follow-up PR (or if you have sufficient rights here, you can directly add commits to this PR)?

regisphilibert:
1.  I think both ways should be mentioned, but I would use the newScratch way for the examples after having introduced the notion of course.
2. I was just saying in order to manipulate maps, Scratch is faster than using `merge` or other workaround, no need to mention it in the doc.

salim-b:
I've changed all examples as you suggested in 1. And thanks for clearing up 2.

salim-b:
Thanks for your inputs @regisphilibert. I'm fine with your redactions. Although I'm not completely convinced that the history shouldn't be mentioned since all scratch instances I've encountered so far (apart from your blog post) seem to be due to the pre-Hugo-0.48 scoping issues with Go template vars – and a lot of people won't realize that they don't need to use Scratch anymore unless they actually read your blog post.

regisphilibert:
> Although I'm not completely convinced that the history shouldn't be mentioned since all scratch instances I've encountered so far

I'm not sure about the personal experiences of users landing on the .Scratch page from the doc. I personally see tech documentation as a handbook, where history or motivations of a certain feature seem superfluous. 

But I think you're right about certain users needing context. So we should keep the Go issue and the blog post references along a very short sentence that kind of lead the users you are mentioning to those helpful ressources without taking over the introduction of this page. 
Writing this short sentence with those references and clarifications contained in a `{{% note %}}` might not be easy, but I really think it's worth trying keeping this very short and outside of the main reader flow.

salim-b:
> (...) we should keep the Go issue and the blog post references along a very short sentence that kind of lead the users you are mentioning to those helpful ressources without taking over the introduction of this page

I tried my best :)

regisphilibert:
🎉

PR #1454, opened 2021-05-09
Update querify function description and examples

jmooring:
The querify function now accepts a slice per
<https://github.com/gohugoio/hugo/pull/8305>.


bep:
Good catch, but It would make it easier to manage if you added this to the `hugo` repo.

PR #1346, opened 2021-02-28
add more details on about gh-pages and baseURL on hosting-on-github.md

vinliao:
When I tried following the tutorial, I found out that these two small details are required for the whole thing to work, but it's not present in the docs.

P.S. I'm not sure whether to add my name to the `authors` at the top.

PR #1455, opened 2021-05-12
Hosting on GitHub: reflect configuration of used branch

va1entin:
Even though the GitHub Pages docs suggest using the `main` branch for personal or organization pages, the Hugo docs about Hosting on GitHub - correctly - suggest a GitHub workflow configuration, which pushes the site built by Hugo to a branch `gh-pages` and not the `main` branch (which makes sense because the Hugo page itself, not the built version, which you want to publish, is in `main`).
The action in question is `peaceiris/actions-gh-pages`.

At least for personal or organization pages the `gh-pages` branch needs to be explicitly set to be used by GitHub Pages  - otherwise it uses the `main` or `master` branch and the user might end up with a 404 or just the projects README (if it exists) being shown.

I stumbled upon the issue this week while migrating my own site running on GitHub Pages to Hugo and believe it stems from the GitHub Pages docs having been used when writing this document initially even though they might be specific to Jekyll, which is tightly integrated with GitHub Pages, in this particular aspect.

With the Jekyll version of my site the `main` branch was indeed the only one, which, combined with the docs, confused me.

This PR proposes adding a section about configuring the `gh-pages` branch to be the one used by GitHub Pages and modifying a section above to reflect this.

Looking forward to reading what you think about this issue.

### DIFF
--- a/content/en/hosting-and-deployment/hosting-on-github.md
+++ b/content/en/hosting-and-deployment/hosting-on-github.md
@@ -36,12 +36,7 @@ Please refer to the [GitHub Pages documentation][ghorgs] to decide which type of
 
 ## GitHub User or Organization Pages
 
-As mentioned in the [GitHub Pages documentation][ghorgs], you can host a user/organization page in addition to project pages. Here are the key differences in GitHub Pages websites for Users and Organizations:
-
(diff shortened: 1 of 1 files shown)


jmooring:
Superseded by #1346

PR #1462, opened 2021-05-18
Make render-heading example accessible by default

CarlSchwan:
Since this code will be copy-pasted on many projects, it's important to make sure it's accessible.

### DIFF
--- a/content/en/getting-started/configuration-markup.md
+++ b/content/en/getting-started/configuration-markup.md
@@ -221,7 +221,7 @@ Here is a code example for how the render-image.html template could look:
 Given this template file
 
 {{< code file="layouts/_default/_markup/render-heading.html" >}}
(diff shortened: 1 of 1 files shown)


coliff:
I'm not too sure about this, adding `aria-hidden` makes it not accessible.

Would it not be better to add `aria-label="Anchor` instead? That's what Anchor JS does.
https://www.bryanbraun.com/anchorjs/#options

ognarb:
yeah, that might be a better idea.

jmooring:
Thanks for this, but I think it would be better to publish this in the [forum](https://discourse.gohugo.io/) in the "tips & tricks" category with an "accessibility" tag. The examples should demonstrate, simply and without opinion, how to access the context passed to each render hook. I am not suggesting that accessibility is an opinionated objective, but I think accessible examples belong elsewhere.

PR #1471, opened 2021-06-06
live reload: add section about `--navigateToChanged`

lucasew:
After some hours of hacking I've discovered that Hugo do this with just one extra flag.

I think it should be more explicit, probably saving some hours of someone else xD.

Related: https://discourse.gohugo.io/t/how-to-get-the-changed-file-route-on-the-frontend-part/33260


angristan:
Thanks for this! I discovered the option thanks to you.

PR #1476, opened 2021-06-18
Pin hugo version

peaceiris:
The `latest` input will destroy the user pipeline silently. Hardcoding a tool version to a specific is better for the stable CI/CD workflow.

- Set `enableInlineShortcodes` to true
- Show a specific Hugo version using the inline shortcode


bep:
Thanks; we do need to revisit the whole GitHub actions thing, but we're not doing enableInlineShortcodes for this.

PR #1477, opened 2021-06-19
Mention that `deep` is the default for `_merge`

salim-b:
I think this helps spotting the **default** setting when briefly skimming this help section.

### DIFF
--- a/content/en/getting-started/configuration.md
+++ b/content/en/getting-started/configuration.md
@@ -94,7 +94,7 @@ shallow
 : Only add values for new keys.
 
 deep
-: Add values for new keys, merge existing.
+: Add values for new keys, merge existing. The default.
 
 Note that you don't need to be so verbose as in the default setup below; a `_merge` value higher up will be inherited if not set.
 


bep:
There is not a single default setting.

salim-b:
Ah damn, my fault, sorry. I didn't realize that the embedded `config.toml` isn't an example but a representation of _the_ default settings. 🙈

Complicated (though powerful) stuff!

I think the intro of the configuration merging documentation is quite confusing (for non-insiders). Being able to configure the configuration merging is a completely new thing (brought by 0.84), right?

bep:
>I think the intro of the configuration merging documentation is quite confusing (for non-insiders). Being able to configure the configuration merging is a completely new thing (brought by 0.84), right?

So, this started out by me fixing a bug in the OS env config overrides, not understanding the rather complicated upstream library used (Viper), taking me down a big rabbit hole, removing Viper. The new `_merge` was added to allow people to opt out of downstream (theme, modules) config ... But yes, it should be powerful, and certainly made the Hugo internal logic simpler to understand and explain to people.

PR #1483, opened 2021-06-27
Fix two typos

raoulb:
Well, seems there were two typos :-/


raoulb:
Hmm, actually both words `constitue` and `constitute`exist.

coliff (approved):

PR #1489, opened 2021-07-08
Fix link to theme components in 0.42 release notes

capsulecorplab:
Link to theme components should be https://gohugo.io/hugo-modules/theme-components/


coliff (approved):

coliff:
good spot - thanks!

PR #1492, opened 2021-07-11
Add documentation for babel sourceMap

richtera:
Add documentation for babel sourceMap.

### DIFF
--- a/content/en/hugo-pipes/babel.md
+++ b/content/en/hugo-pipes/babel.md
@@ -65,6 +65,10 @@ compact [bool]
 verbose [bool]
 : Log everything
 
+sourceMap [string]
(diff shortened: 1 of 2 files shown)


davidsneighbour:
This is ambiguous: Sourcemap settings inside of the babel.config.js are being ignored other than `inputSourceMap: true` which defaults to true. -- would `inputSourceMap: false` be ignored or not? if the value is irrelevant to the ignoring, let's remove the `: true` part of the sentence.

richtera:
Somehow the docs for the sourceMap "external" support in both js.Build and Babel didn't end up in the docs when the PR was merged. I might be adding the docs in the wrong place.

PR #1501, opened 2021-08-01
Document how isset can be used with shortcode parameters

ayoisaiah:
I was looking for a way to test for the presence of an optional shortcode parameter, but didn't find helpful information in the docs.

This PR documents how the it may be achieved through the `isset` function.


jmooring:
This is incorrect. With this shortcode:

```
{{ if isset .Params `lazy` }}
  <img src="" loading="lazy">
{{ else }}
  <img src="" loading="eager">
{{ end }}
```

This markdown:

```
{{< image lazy=false >}}
```

Will produce this HTML:

```
<img src="" loading="lazy">
```

PR #1502, opened 2021-08-01
Update hosting on GitHub with some modifications

vivekkj123:
This pull request aims to modify the Hugo documentation based on feedbacks from some beginners.
# Changes

1. Mention the need of secret token in documentation.
2. Change the variable `GITHUB_TOKEN` to `TOKEN` (Because GH doesn't allow to make a variable starting with `GITHUB_`)
3. Mention to change the source branch of GH pages to `gh-pages` instead of main branch


athul:
Just to chime in, GitHub actions already use a `GITHUB_TOKEN` variable to authenticate an action. That token can be used to authenticate the GitHub pages workflow. 
Reference : https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

vivekkj123:
Ohh. Thanks @athul , I was not aware of GITHUB_TOKEN. Closing this PR.

PR #1504, opened 2021-08-03
Fix `lang.FormatPercent` description

salim-b:


coliff (approved):

bep:
That JSON file is autogenerated from the Go source code.

I welcome PR with improvements here: https://github.com/gohugoio/hugo/blob/dfe54d32169ff6dae18023a4dada6009d48f8636/tpl/lang/lang.go#L74

(I will make sure to regenerate the JSON file when needed).

PR #1511, opened 2021-08-09
Add page for deploying with rclone

danielfdickinson:
Rclone is  a multi-protocol file copy and and sync tool that can make
deploying static websites a breeze. This commit adds a brief guide to
using Rclone to deploy sites generated with Hugo.

Hope you find this useful!

### DIFF
--- a/content/en/hosting-and-deployment/deployment-with-rclone.md
+++ b/content/en/hosting-and-deployment/deployment-with-rclone.md
@@ -1,0 +1,57 @@
+---
+title: Deployment with Rclone
+linktitle: Deployment with Rclone
+description: If you have access to your web host with SFTP/FTP/SSH/HTTP(DAV), you can use rclone to incrementally deploy your entire Hugo website.
+date: 2021-08-09
+publishdate: 2021-08-09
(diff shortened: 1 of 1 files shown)


danielfdickinson:
Any comments or issues, or just busy?

danielfdickinson:
Ping? I know the docs are being restructured, is that why this is on hold (it being based on an existing deploy method)?

jmooring:
I will merge this.

I think in general we are reticent to add generic deployment methods to the documentation. For example, "Deploy with ftp" or "Deploy with FileZilla" are better documented elsewhere. However, there is similar precedent in our documentation (rsync), and rclone is a Go project, versatile, popular, and appears to be well-maintained.

danielfdickinson:
Thank you. I understand the reticence although it helps to know about it (I know finding time to reply to PR can be difficult though, and not everyone takes 'no' well). The explanation was what I needed more than the merge, although of course I am happy about the merge. Thank you again.

danielfdickinson:
I wonder if treating this as another situation like #1573 would make sense, in terms of future handling?

Issue #1163, opened 2020-07-01
The list of available emojis is inaccurate 


jbleduigou:
I believe there are some inaccuracies in the list of emojis described in page content/en/functions/emojify.md

For instance ``` :bowtie: ``` does not convert properly.

My understanding is that the actual code map used is https://github.com/kyokomi/emoji/blob/master/emoji_codemap.go
Based from http://www.unicode.org/emoji/charts/emoji-list.html

jmooring:
Background information; not actionable.

The documentation points to <https://www.webpagefx.com/tools/emoji-cheat-sheet/> from these pages:

- <https://gohugo.io/functions/emojify/>
- <https://gohugo.io/getting-started/configuration/>

The referenced cheat sheet contains 877 emoji codes (e.g., `:foo:`), 17 of which are not recognized by Hugo; they are unique to GitHub. See <https://emojipedia.org/github/>.

Hugo's code map is an amalgamation of four sources, currently with 4670 codes representing 2995 unique emojis (1675 of the codes are aliases). Hugo does not render 55 of the 4670 codes due to the presence of one or more of these characters: `( ) ’ Å & ô ç é ã í . * # ! ñ`. 

While emojis are somewhat standardized through the Unicode Consortium (<https://home.unicode.org/emoji>), _the codes used to represent them are not_. That is why there are 1675 aliases in Hugo's implementation. For example, this emoji (:no_good:) can be written as any of the following in Hugo, though only the first three are recognized by GitHub:

```
:no_good: 
:ng_woman: 
:no_good_woman: 
:woman-gesturing-no: 
:woman_gesturing_NO: 
:woman_gesturing_no:
```

An ideal cheat sheet for Hugo users would:

- Have two versions
  - Curated (most commonly used, a few hundred at most)
  - Complete (all 2995 symbols)
- Organize emojis by category, and perhaps by sub-category on the complete list
- Provide one code for each symbol
  - Content authors want to know how to insert a symbol; they don't need to know about the aliases
  - Among several alternatives for the same symbol, choose most commonly used based on GitHub, Twitter, etc.
- Contain keywords to facilitate search (Ctrl+F in browser)
- Exist on a single page to facilitate search (Ctrl+F in browser)
- Load fast, probably using embedded PNGs

[...]

PR #1518, opened 2021-09-01
Update emojify.md

acvill:
The cheat sheet provided does not include a full list of the emojis available through emojify.  Need to link to go codemap until a full cheatsheet is available.


jmooring:
While technically accurate (mostly), I'm not sure that the information is useful. See related comments:
<https://github.com/gohugoio/hugoDocs/issues/1163#issuecomment-683751280>

PR #1520, opened 2021-09-08
Install on Windows: Correct + augment

hh-lohmann:
* PowerShell / "other CLIs" don't have "set"
* "setx" for global setting for current user (no admin rights required)


alfawal:
Thanks for letting your command delete all of my current user's environment variables just to add Hugo.

PR #1522, opened 2021-09-12
Fix mobile footer getting stacked below .tab-button elements

Chevindu:
Mobile footer is getting stacked below the tab buttons of language-specific implementation examples.

Screenshot 2021-09-12 at 09 00 41

This PR is an attempt to change the stacking index of the footer to overcome that.


jmooring:
This change need to happen in a different repository:
<https://github.com/gohugoio/gohugoioTheme/>

PR #1526, opened 2021-09-19
Sync docunented sitemap.xml with bundled template

jtatum:
The bundled template was updated in 4867cd1, but the docs weren't. This commit syncs the docs with the bundled template.


jmooring:
Superseded by #1699

PR #1528, opened 2021-10-02
Add Debricked showcase

sweoggy:
We recently moved our website to Hugo and have loved it every since.

Would love to get included in the showcases :)


jmooring:
Thank you for your submission, but in its current form this example does not rise to the level of a "showcase" item. It lacks the details required to make it interesting or notable. The format should be something like:

- Problem statement (where we were)
- Actions (what we did)
- Result (where we are now)

Where possible, describe how a specific Hugo feature or capability helped you to solve the problem:

- Feature (what it is)
- Function (what it does)
- Benefit (how it helped)

We will be happy to reconsider if you would care to resubmit. Thanks again.

PR #1529, opened 2021-10-04
Updates Hugo Pipes (JS building sourcemaps) documentation

bgdeutsch:
Based on [this commit](https://github.com/gohugoio/hugo/commit/2c8b5d9165011c4b24b494e661ae60dfc7bb7d1b) it looks like `external` is an option for JS building with Hugo Pipes.   Just adding this to the documentation.   Not sure if there are plans to add other options - if not i'm happy to remove the "(we will improve that)" part.   Thanks!


bep (approved):

coliff:
Realised this is already fixed now. https://gohugo.io/hugo-pipes/js/#options

Thanks though @bgdeutsch !

PR #1530, opened 2021-10-06
Added a brief "publishing source" section

Digital-Grinnell:
I found this document to be a little confusing because it doesn't mention (anywhere I could find) that the workflow creates a `gh-pages` branch in the repo, and that the `root` directory in this branch should be set as the GitHub "publishing source".  My wording isn't wonderful (I write a lot of code, not a lot of documentation), but I feel like the document could benefit from a little more explanation here.


jmooring:
Superseded by #1346

PR #1548, opened 2021-10-17
Clarify the distinction between resources and .Resources

PaulSD:
I've seen a number of references/comments that mix up resources.Get and .Resources.GetMatch (for example: https://discourse.gohugo.io/t/data-files-in-the-page-bundle/19284/2).  This leads to confusion (for example: https://discourse.gohugo.io/t/what-am-i-not-getting-about-resources-get/33517).

Ideally, `resources` would have a different name (like `assets`) to eliminate some of the ambiguity with `.Resources`.  However, since it is probably too late to change the name now, the next best thing would be to have the docs more explicitly distinguish between these.

This PR attempts to clarify the distinction between these in the documentation.

### DIFF
--- a/content/en/content-management/page-resources.md
+++ b/content/en/content-management/page-resources.md
@@ -13,10 +13,7 @@ menu:
     parent: "content-management"
     weight: 31
 ---
-Page resources are only accessible from [page bundles]({{< relref
-"/content-management/page-bundles" >}}), those directories with `index.md` or
-`_index.md` files at their root. Page resources are only available to the
-page with which they are bundled.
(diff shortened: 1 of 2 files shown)


PaulSD:
Ok.  I've updated my PR to use "global assets" instead of "asset resources".

Is there currently such a thing as a global resource or a page asset?  I had thought there were only page resources and global assets...

bep:
I agree about the global discussion, but I think an important distinction that I suspect is missing in all of this is that these two resource collections have very much different content.

* `site` and `.Site` is global vs. page, but they point to the same object
* `resources` and `.Resources` are two different collections.

PaulSD:
Yeah, that is actually most of what I changed in this PR - explicitly pointing out in the docs that `resources` and `.Resources` are different things. :)

lschierer:
I would love to see this merged in.  I'm currently struggling with exactly this distinction.

jmooring:
This will be addressed in https://github.com/gohugoio/hugoDocs/issues/2307

PR #1553, opened 2021-10-29
Display 'Configure Markup' in list of content

thewheat:
Related to #1473 

When on the [Configure Hugo](https://gohugo.io/getting-started/configuration/) page
- clicking the right arrow on the right sidebar 
- brings you to the [Configure Markdown](https://gohugo.io/getting-started/configuration-markup/) page
- This page doesn't show up in the list of content on the right and is a bit confusing as brought up in #1473 

image

* * * 

This PR makes the page visible in the list as shown below to make it less confusing

image


jmooring:
I've been thinking about this, and we're going to live with this minor annoyance for the moment. At some point I would be interested in consolidating the configuration documentation into one section per https://github.com/gohugoio/hugoDocs/issues/1888.

Issue #1112, opened 2020-05-18
Syntax for assigning to variables is hard to find


rgov:
The documentation has examples like

    {{ $foo := "bar" }}

However I'm not sure where this syntax is explained. One would expect to find them under "Variables" but those are more like structures that Hugo provides and their fields, rather than how variables like `$foo` are defined.

I searched for "assign" and did not see it in the results.

I think [this is what I want](https://gohugo.io/templates/introduction/#1-define-a-variable-independent-of-context) but it is deep into a very long document.

A weird thing is that the HTML sample is called "tags-range-with-page-variable.html" but if I look in the docs about Page Variables, it isn't described there. Instead it points me to .Scratch... so it is quite confusing.

I was hunting for this because you have to use the `:=` operator instead of `=`. Not sure how to input that into a search field so the navigability of the docs is important here.

rgov:
It seems simple enough to add a short blurb about assigning and using variables to the Variables page. This is not an esoteric part of Hugo's templating language.

PR #1554, opened 2021-10-29
Fixes: #1112 Add clarification of how to set and use variables

thewheat:
Added a section showing how to set a variable, access the value and a link to learn more 

**Preview**
image

### DIFF
--- a/content/en/variables/_index.md
+++ b/content/en/variables/_index.md
@@ -4,7 +4,7 @@ linktitle: Variables Overview
 description: Page-, file-, taxonomy-, and site-level variables and parameters available in templates.
 date: 2017-02-01
 publishdate: 2017-02-01
-lastmod: 2017-02-01
+lastmod: 2021-10-30
 categories: [variables and params]
 keywords: [variables,params,values,globals]
 draft: false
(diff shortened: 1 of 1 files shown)


rgov:
@jmooring I see that it is documented on that page but if you are trying to answer the question of "how do I use variables?" you would reasonably go to the part of the documentation labeled Variables > Variables Overview which seems more specific than Templates > Introduction which you might guess has all sorts of unrelated information about the templating system.

Even if you did go to the Introduction first, you might see the first section "Access a Predefined Variable" which doesn't describe how to assign to it.

In my opinion it doesn't hurt to repeat this, though I would maybe create a page called `Custom Variables` that describes it. And include the naming rules which I'm sure are more complex than just "must start with a `$`" (can you use hyphens, start with a number, etc?).

callaloo:
This is better than what is in the docs.

jmooring:
@coliff  I've already weighed in on this change. The "Variables" section of the documentation is about the immutable variables exposed to the templates. This PR is about initializing and assigning custom variables, and I think it belongs elsewhere. My two cents...

PR #1559, opened 2021-11-01
Update the timeout default

jmooring:
This was increased to 30s with:
https://github.com/gohugoio/hugo/pull/6535


jmooring:
Disregard. Handling this in https://github.com/gohugoio/hugo/pull/9114.

PR #1567, opened 2021-11-04
add project

sean-pwd:
add project to list (hugo-svg-font)


bep:
Thanks for this, I'm sure it's useful, but it does not reach the bar of "general useful" to be included in that section.

PR #1581, opened 2021-11-28
make the render heading example accessible

jmau111:
I've added HTML markup to make anchor links more accessible.


jmooring:
@jmau111 Thanks for this, but I think it would be better to publish this in the [forum](https://discourse.gohugo.io/) in the  "tips & tricks" category with an "accessibility" tag. The examples should demonstrate, simply and without opinion, how to access the context passed to each render hook. I am _not_ suggesting that accessibility is an opinionated objective, but I think accessible examples belong elsewhere.

jmau111:
@jmooring disappointing. Have a good saturday.

PR #1597, opened 2021-12-08
Add markdownlint configuration file

### LINKED ISSUE #1102: Issues with consistency and best practices across .md content files
I frequently use [markdownlint](https://github.com/markdownlint/markdownlint) to flag issues with my markdown files and find this tool very useful. Keeping the documentation files tidy and consistent has advantages in the long run.

When Markdown linter is used to detect issues with the files in `content` directory, it returns a lot of issues which would reduce the legibility of the md files and might cause problems in the future.

I don't necessarily agree with all the rules of this tool so I've created a style file to run the linter while ignoring some [rules](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md). It looks like this:

```ruby
all

# MD002 - First header should be a top level header
# Removing this rule as page titles are determined by the values in the
# frontmatter

exclude_rule 'MD002'

# MD013 - Line length
# This returns a lot of issues so decided to keep it out for the initial
# inspection. But this should be implemented as well with sole exception being
# the links and images with long URLs.

exclude_rule 'MD013'

# MD024 - Multiple headers with the same content
# Ignoring this as some documentation pages has multiple sub headers with the
# same content/parameter for different subjects

exclude_rule 'MD024'

# MD033 - Inline HTML
# Ignoring this to remove issues triggered by shortcodes.

exclude_rule 'MD033'

# MD041 - First line in file should be a top level header
# Not necessary

exclude_rule 'MD041'

# I also think an ordered list should be represented with an ordered list not
# just digit '1'

rule 'MD029', :style => 'ordered'
```

Adding this to a file named markdown_style.rb and running `mdl --style markdown_style.rb --ignore-front-matter .` returns a whopping 2375 issues for all the content files. The results are [in this paste](https://paste.tildeverse.org/?22fb197871641e43#7zLTcWZAFAaVArYuUzWvYox45hd5NfxvSet1XxU4HRyU).

Have you ever considered setting some rules around the formatting and consistency of the content files? I think everyone would benefit from it and it could futureproof the content files, which are right at the core of this documentation project. I am not sure how this would fit in the build process. But if you think this is a problem and decide to implement it (or at least fix the existing issues), I'd like to work on this.

jmooring:
Closes #1102

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,4 @@
+# https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
+
+MD013: false  # Line length
+MD033: false  # Inline HTML


coliff (approved):

coliff:
this is great! maybe we should use a Markdownlint GitHub Action running on each PR too?

jmooring:
> maybe we should use a Markdownlint GitHub Action running on each PR too?

Please create an issue for that. Needs some thought, discussion, testing.

PR #1612, opened 2021-12-25
docs: add Layer0 deployment guide

rishi-raj-jain:


coliff:
thanks but we are currently not adding any more "deployment options/guides" to the documentation site. We can't add guides for every hosting company. 

https://www.alexa.com/siteinfo/layer0.co

See reasons mentioned in #1429 for more related discussions on this topic.

rishi-raj-jain:
@coliff 

Appreciate the transparency.

Tho, with all due respect, I don't see how you've 21cloudbox which is ranked way lower than edg.io (the new layer0)
Screenshot 2023-02-14 at 11 53 58 AM

Screenshot 2023-02-14 at 11 54 00 AM

rishi-raj-jain:
We're not marketing ourselves here as per in the issue you referred, It can just be a link to https://docs.edg.io/guides/sites_frameworks/getting_started/hugo without anything else. More than happy with that.

I like how Nuxt does it, where providers maintain their own guide

PR #1615, opened 2021-12-26
docs: add a link to the mailmap documentation

davidsneighbour:
I find that Hugo should contain more links to relevant documentation of features in other tools/documentations. Please let me know if we don't want that and I'll stop. In this specific case I read about `.mailmap` in the Hugo docs and as a developer knew what was meant, but I think less dev-y users might not understand this. Instead of starting to add paragraphs about the format maybe it's ok to link to the features docs? As I said, let me know if that is overkill/bloat. It would make my life easier than seeing a word and googling it :)

(for instance, if we link to git in this case we don't need to add extra-documentation about what the mailmap file should look like and what should be in there)

### DIFF
--- a/content/en/variables/git.md
+++ b/content/en/variables/git.md
@@ -37,10 +37,10 @@ The `GitInfo` object contains the following fields:
 : the abbreviated commit hash (e.g., `866cbcc`)
 
 .AuthorName
-: the author's name, respecting `.mailmap`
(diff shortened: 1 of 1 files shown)


bep:
Informational links are great (but there is a thin line between these and "product linking", which I'm not a fan of. We do, however, have a little too much text information about non-Hugo-things, but that's a broader discussion,

PR #1625, opened 2022-01-12
Update image processing

jmooring:


jmooring:
This is a substantial rewrite based on my interactions with users on the forum. It provides some introductory information about resources and rendering that exists elsewhere in the documentation, but has apparently been difficult to find if you don't know what to search for.  

I've attempted to use active voice where possible, clarified some concepts, and provided additional examples (e.g., how do I convert an image without resizing it?).

coliff (approved):

PR #1626, opened 2022-01-15
Add FreeBSD install instructions

sergio-carlavilla:


sergio-carlavilla:
@bep can you commit this please?
It would be really great to add this to the Hugo Documentation

PR #1266, opened 2020-10-21
Hugo Modules plural typo

fgiuliani:

PR #1629, opened 2022-01-17
Update page content

sudomateo:
The Security Model page had some grammatical errors, incorrect spelling of words, and unclear context. These were all addressed in this commit.

### DIFF
--- a/content/en/about/security-model/index.md
+++ b/content/en/about/security-model/index.md
@@ -25,7 +25,7 @@ But when developing and building your site, the runtime is the `hugo` executable
 
 * Hugo has a virtual file system and only the main project (not third-party components) is allowed to mount directories or files outside the project root.
 * Only the main project can walk symbolic links.
-* User-defined components have only read-access to the filesystem.
+* User-defined components have read-only access to the filesystem.
 * We shell out to some external binaries to support [Asciidoctor](/content-management/formats/#list-of-content-formats) and similar, but those binaries and their flags are predefined and disabled by default (see [Security Policy](#security-policy)). General functions to run arbitrary external OS commands have been [discussed](https://github.com/gohugoio/hugo/issues/796), but not implemented because of security concerns.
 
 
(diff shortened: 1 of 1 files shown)


coliff (approved):

content/en/about/security-model/index.md
```diff
@@ -47,9 +47,9 @@ HUGO_SECURITY_HTTP_URLS=none hugo
...
+Hugo is built as a static binary using [Go Modules](https://github.com/golang/go/wiki/Modules) to manage its dependencies. Go Modules have several safeguards, one of them being the `go.sum` file. This is a database of the expected cryptographic checksums of all of your dependencies, including transitive dependencies.
 
-[Hugo Modules](/hugo-modules/) is built on top of Go Modules functionality, and a Hugo project using Hugo Modules will have a `go.sum` file. We recommend that you commit this file to your version control system. The Hugo build will fail if there is a checksum mismatch, which would be an indication of [dependency tampering](https://julienrenaux.fr/2019/12/20/github-actions-security-risk/).
+[Hugo Modules](/hugo-modules/) are built on top of the functionality of Go Modules, and a Hugo project using Hugo Modules will have a `go.sum` file. We recommend that you commit this file to your version control system. The Hugo build will fail if there is a checksum mismatch, which would be an indication of [dependency tampering](https://julienrenaux.fr/2019/12/20/github-actions-security-risk/).
```
jmooring, line 52:
```suggestion
[Hugo Modules](/hugo-modules/) is built on top of the functionality of Go Modules, and a Hugo project using Hugo Modules will have a `go.sum` file. We recommend that you commit this file to your version control system. The Hugo build will fail if there is a checksum mismatch, which would be an indication of [dependency tampering](https://julienrenaux.fr/2019/12/20/github-actions-security-risk/).
```

jmooring: "Hugo Modules" is a feature (one thing, singular).

jmooring (changes requested):

content/en/about/security-model/index.md
davidsneighbour: A feature can be consisting of multiple things. Maybe we should try to have documentation that is understood by not only English Majors but also normal people. Changing this specific wording from are to is will irritate readers. On the other side, if this is a forced concoction it should be phrased as "Hugo Modules is a feature that is ..." instead of "Hugo Modules is ..."

To compare: The "Go Modules" documentation has no issue to talk about a singular module: https://go.dev/blog/using-go-modules

sudomateo: The "is" read a bit weird to me at first. Then I thought the same thing that was mentioned above, that it was the name of a feature. However, the actual [Hugo Modules](https://gohugo.io/hugo-modules/) page it links to uses "Hugo Modules are" in multiple places. That prompted me to make the change.

jmooring: I thought we had consensus on this:
<https://github.com/gohugoio/hugoDocs/pull/1266#issuecomment-713394631>

sudomateo: In that case, I like what @davidsneighbour proposed with the "Hugo Modules is a feature that is ..." wording. Especially since in the context of this specific page this is the first time the Hugo Modules feature is being introduced to the reader.

jmooring: @sudomateo Works for me. We have a similar feature/name issue with Hugo Pipes...

sudomateo: Updated!

PR #1636, opened 2022-01-19
Add "Hosting&deployment" docs - deployment with Buddy

tomekpapiernik:
I've created a new doc for the "Hosting&deployment" section which walks the user through the process of setting up an automated deployment pipeline with Buddy. The tool's very simple to use and GUI-based and I think that many Hugo users can benefit from using it. 

I used the metadata from another doc from this section, I hope that's okay.


jmooring:
@tomekpapiernik Thanks, but it would be better to post this information elsewhere, perhaps on the buddy.works site, or as a tips and tricks posting on https://discourse.gohugo.io. We are gradually removing all but the most notable entries from the tools and deployment sections.

Issue #1622, opened 2022-01-04
transform.Unmarshal for XML docs needs clarification


faraixyz:
I'm excited for the XML support but I was wondering how things like attributes, tags with a namespace and multiple tags of the same type are handled as they aren't [covered in the docs](https://gohugo.io/functions/transform.unmarshal/#xml-data) . From the little work I've done, I've found that

* multiple tags of the same name return a map with each instance,
* attributes start with a `-`. Given that, you need to use the `index` function,
* it looks like the namespaces are stripped out.

jmooring:
Please provide a small example with attributes, tags with a namespace, and multiple tags of the same type.

faraixyz:
With an RSS XML feed like this:

```
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
    <channel>
        <title>RSS Feed</title>
        <link>/</link>
        <atom:id>tag:an.example,2020-01-10:/</atom:id>
        <atom:link href="/feed.rss.xml" rel="self" type="application/rss+xml" />
        <item>
            <title>A</title>
            <link>/a</link>
            <guid>/a</guid>
            <description>Some Stuff</description>
        </item>
        <item>
            <title>B</title>
            <link>/b</link>
            <guid>/b</guid>
            <description>Some More Stuff</description>
        </item>
        <item>
            <title>C</title>
            <link>/c</link>
            <guid>/c</guid>
            <description>Some Stuff Again</description>
        </item>
    </channel>
</rss>
```

This is dumped

```
map[-atom:http://www.w3.org/2005/Atom -version:2.0 channel:map[item:[map[description:Some Stuff guid:/a link:/a title:A] map[description:Some More Stuff guid:/b link:/b title:B] map[description:Some Stuff Again guid:/c link:/c title:C]] link:[/ map[-href:/feed.rss.xml -rel:self -type:application/rss+xml]] title:RSS Feed]] 
```

Accessing things will go like this

```
{{ $feed.-atom }} => error
{{ index $feed "-atom" }} => http://www.w3.org/2005/Atom
{{ $feed.channel.title }} => RSS Feed
{{ $feed.channel.item }} => [map[...] map[...] map[...]]
{{ $feed.channel.id }} => tag:an.example,2020-01-10:/
{{ index $feed.channel "atom:id" }} => 
[...]

PR #1637, opened 2022-01-19
Clarify transform.Unmarshal when used with XML

wgroeneveld:
See https://discourse.gohugo.io/t/parsing-xml-in-data-with-hugos-new-xml-support-attributes/36654/2 and https://github.com/gohugoio/hugoDocs/issues/1622
The new XML parsing abilities of Hugo are currently not well documented. 

I am not sure as to
1. How much I had to add, since everything is very much condensed;
2. If it deserves its own page, since it's a bit strange to put it all in transform.Unmarshall;

Since I don't frequently update the official docs, a second look is appreciated.

### DIFF
--- a/content/en/functions/transform.Unmarshal.md
+++ b/content/en/functions/transform.Unmarshal.md
@@ -6,7 +6,7 @@ categories: [functions]
 menu:
   docs:
     parent: "functions"
-keywords: []
+keywords: [xml]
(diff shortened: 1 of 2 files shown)


wgroeneveld:
This isn't moving forward, anything I can do to help?

wgroeneveld:
One year later, without any replies when I ask "anything I can do to help", this still isn't merged or closed?

PR #1658, opened 2022-02-17
Add explicit `contentDir` to module-config-mounts

chalin:
Context: https://discourse.gohugo.io/t/content-mount-problem-for-single-language-multilingual-site/37215

/cc @jmooring


jmooring:
@chalin Thanks. Will merge.

PR #1659, opened 2022-02-20
Add MPR installation instructions

hwittenborn:
The [MPR](https://mpr.makedeb.org/) is a project for [makedeb](https://makedeb.org/), which itself is an alternative packaging tool for Debian and Ubuntu based systems that still ultimately builds .deb packages.

I'm currently maintaining [Hugo on the MPR](https://mpr.makedeb.org/packages/hugo), as I needed some features myself that weren't present in Ubuntu 20.04.

The MPR is a platform is one I created (as well as makedeb itself), but I think the MPR package could provide a lot of benefits for users using Debian and Ubuntu based distros where they wouldn't otherwise be able to get the latest release of Hugo when sticking to just .deb packages.

### DIFF
--- a/content/en/getting-started/installing.md
+++ b/content/en/getting-started/installing.md
@@ -494,11 +494,9 @@ This option is not recommended because the Hugo in Linux package managers for De
 
 Hugo can be installed from the [makedeb Package Repository](https://mpr.makedeb.org/packages/hugo), which provides an up-to-date version of Hugo for users on Debian and Ubuntu based systems. To install it you need to first have [makedeb](https://makedeb.org/) installed, after which it can be installed via the `makedeb` command:
 
-```sh
-git clone 'https://mpr.makedeb.org/hugo.git'
-cd hugo/
-makedeb -si
-```
+    git clone 'https://mpr.makedeb.org/hugo.git'
+    cd hugo/
+    makedeb -si
 
 ### Arch Linux
 
--- a/content/en/getting-started/installing.md
+++ b/content/en/getting-started/installing.md
@@ -488,7 +488,17 @@ Hugo installed via Snap can write only inside the user’s `$HOME` directory---a
 
 What this installs depends on your Debian/Ubuntu version. On Ubuntu bionic (18.04), this installs the non-extended version without Sass/SCSS support. On Ubuntu disco (19.04), this installs the extended version with Sass/SCSS support.
 
-This option is not recommended because the Hugo in Linux package managers for Debian and Ubuntu is usually a few versions behind as described [here](https://github.com/gcushen/hugo-academic/issues/703)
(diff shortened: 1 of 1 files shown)


content/en/getting-started/installing.md
```diff
@@ -488,7 +488,15 @@ Hugo installed via Snap can write only inside the user’s `$HOME` directory---a
...
-This option is not recommended because the Hugo in Linux package managers for Debian and Ubuntu is usually a few versions behind as described [here](https://github.com/gcushen/hugo-academic/issues/703)
+This option is not recommended because the Hugo in Linux package managers for Debian and Ubuntu is usually a few versions behind as described [here](https://github.com/gcushen/hugo-academic/issues/703).
+
+### MPR Package
+
+Hugo can be installed from the [makedeb Package Repository](https://mpr.makedeb.org/packages/hugo), which provides an up-to-date version of Hugo for users on Debian and Ubuntu based systems. To install it you need to first have [makedeb](https://makedeb.org/) installed, after which it can be installed via the `makedeb` command:
+
+    git clone 'https://mpr.makedeb.org/hugo.git'
+    cd hugo/
+    makedeb -si
```
hwittenborn, line 499: If you guys decide that you want to merge this, I should also note that makedeb doesn't have update capabilities itself outside of running `git pull` for the MPR package and then rebuilding.

There are some [MPR helpers](https://docs.makedeb.org/mpr/list-of-mpr-helpers/) that do support auto-updating, though I wasn't sure if I should include that or not.

davidsneighbour:
> I needed some features myself that weren't present in Ubuntu 20.04.

What was missing? Just asking...

hwittenborn:
> What was missing?

I needed the [`heading` markdown hook](https://gohugo.io/getting-started/configuration-markup#markdown-render-hooks), which isn't supported in Ubuntu's version of Hugo which is only at 0.68.3.

It's a tiny feature I suppose, though I found myself wanting it a lot for header anchors for posts on the makedeb homepage, which is currently powered by Hugo too :)

davidsneighbour:
I am installing my Ubuntu version via the releases page. I guess that is the same long way :) I remember reading somewhere that it is complicated to get the current Hugo version into Ubuntu due to how this whole repo adding/updating works. I guess an "official" PPA might be the best way.

jmooring:
Thanks for this. 

I understand the limitation of installing Hugo with `apt`. The version in the distributions's package repository is, by design, not the latest. 

Ubuntu/Debian users already have several alternatives:
- binary (via gh cli)
- binary (manual download)
- brew
- git+go
- go
- snap (see [current limitations](https://github.com/gohugoio/hugo/issues/9078))

I am not convinced that adding another installation method to the documentation would be helpful. Additionally, without update capabilities, it seems like the user would be substituting one problem for another.

I am closing this, but I encourage you to post something to the [forum](https://discourse.gohugo.io/) to see if it gets any traction.

PR #1662, opened 2022-02-21
Update merge function examples

jmooring:
Supersedes #1641
Closes #1641

### DIFF
--- a/content/en/functions/merge.md
+++ b/content/en/functions/merge.md
@@ -1,6 +1,6 @@
 ---
 title: merge
-description: "`merge` deep merges two maps and returns the resulting map."
+description: "Returns the result of merging two or more maps."
 date: 2019-08-08
 categories: [functions]
 menu:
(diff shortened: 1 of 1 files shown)


davidsneighbour:
I still don't get this: "Merge creates a copy of the final `MAP` and merges any preceding `MAP` into it in reverse order." Doesn't that mean, `{{ $merged.x }}   --> foo` in regards to the first sample? I am not sure if I have a brain-freeze here... that language to me (non-native english speaker) means "merge a b c" is c, then merging b into it, then merging a into it. The "reverse order" seems to be the issue for me. Maybe it means the right thing, but the order of the words in that sentence make it hard to understand?

jmooring:
@davidsneighbour I haven't merged this yet, but note that I added the sentence, "Merging adds elements to the final map; it does not update existing elements." The examples are accurate. I have documented how it currently behaves, but I'm not thrilled about the description.

jmooring:
@davidsneighbour Also remember that these are equivalent:

```
{{ merge $m2 $m1 }}
{{ $m1 | merge $m2 }}
```

jmooring:
This describes how the result is obtained internally with `tpl/collections/merge.go`:

> Merge creates a copy of the final `MAP` and merges any preceding `MAP` into it in reverse order. Merging _adds_ elements to the final map; it does not _update_ existing elements.

Although this is technically correct, I think it makes more sense to describe what it does from the user's perspective:

> Returns the result of merging the given maps from left to right. If a key already exists, `merge` updates its value. If a key is absent, `merge` inserts the value under the new key.

PR #1677, opened 2022-03-08
Adds a Spellcheck GitHub Action and config

coliff:
Can run spellcheck via CLI with:

```
npx cspell \"content/**/*.md\" --no-progress
```

Note the `.cspell.json` config has added many words from this repo to the allowlist - but not all. Over time we can eventually add all false-positive words to this so that spellcheck test across the whole code base passes.


.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches-ignore:
+      - "dependabot/**"
+  pull_request:
+
+jobs:
+  spellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: streetsidesoftware/cspell-action@v1
```
bep, line 13: Can you add a specific commit (to replace `v1`) for the `streetsidesoftware` action, as is our (or mine) current policy for actions not living in the `actions` namespace. It's unfortunate and probably not enough as a security measure, but one could what would happen if some hacker org somehow managed to take over the `streetsidesoftware` repo.


Other than that this looks good.

davidsneighbour: > what would happen if some hacker org somehow managed to take over the streetsidesoftware repo.

Are you aware of permissions for actions? 

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

Longterm this could be useful to set permissions so that they can't wreak too much havoc.

coliff: I've specified the exact version now (v1.5.1).

bep: I should have been more specific, sorry, it needs to be a Git SHA ref -- tags can be overwritten.

@davidsneighbour whatever settings you apply, in my head it all starts with knowing what's running on "your computer".

davidsneighbour: On the releases page of the tag you want to use in the header area.

davidsneighbour: for this one: https://github.com/streetsidesoftware/cspell-action/releases/tag/v1.5.1
it's: 3bc52c3

coliff: failed. can't use shortened versions... need to use: `3bc52c39528d0214a947ff7f8b5c23ad3273435b`

bep: You see the commit hash on the tags: https://github.com/streetsidesoftware/cspell-action/releases/tag/v1.5.1

Also, @davidsneighbour I didn't know about the "permission thing", but my take in this still apply -- we need to use a GitHub token "somewhere", so we might as well be double careful.

PR #1691, opened 2022-03-19
Fix table syntax for strings.Count.md

kujenga:
Without these escape characters, the table is incorrectly rendered in a way that hides the results and splits the example text.


davidsneighbour:
#1686 ;)

coliff:
thanks @kujenga ! closing this in favour of #1686. thanks though!

coliff:
fixed now: https://gohugo.io/functions/strings.count/

PR #1692, opened 2022-03-22
New Header added.

sureshkumargondi:
Strict-Transport-Security,
Permissions-Policy

Validated from this tester. Gets more security grade by adding these two headers.
Validation tool - https://securityheaders.com


jmooring:
Thanks, but this configuration setting is only relevant when running `hugo server` (local development). The related documentation is merely an example of what you _can_ do, not what you _should_ do in any particular situation. 

Also, in the future, please build the documentation site locally _before_ submitting a PR to make sure your changes are free from errors.

PR #1698, opened 2022-03-26
add tags to "related content" configuration

christianspecht:
A new site will have the taxonomy `tags` enabled by default, but
the "related content" configuration will not work if `tags` is missing
there.

--> the example config should contain `tags` as well, so it works as-is for most sites.

### DIFF
--- a/content/en/content-management/related.md
+++ b/content/en/content-management/related.md
@@ -82,9 +82,12 @@ related:
     weight: 100
   - name: date
     weight: 10
+  - name: tags
+    weight: 80
 {{< /code-toggle >}}
 
-Note that if you have configured `tags` as a taxonomy, `tags` will also be added to the default configuration above with the weight of `80`.
+Note that `tags` will only be added to the default configuration if you have configured `tags` as a taxonomy.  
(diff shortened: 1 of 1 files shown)


jmooring:
The section you refer to shows the _default_ configuration.

screen capture

christianspecht:
@jmooring: Okay, maybe I didn't make myself clear enough.  

I *know* that this section shows the default configuration.  
But my experience as a Hugo user was this:

- When I needed a custom configuration for my site's related content, I found this section in the docs, and I copied & pasted this default configuration into my `config.toml`.
- This caused all related content in my site to disappear without any error message *(or I didn't know where to look for errors)*.
- After some hair-pulling, I found that my config must also contain the index for `tags` that I added in my pull request...because my site *had* configured `tags` as a taxonomy, because that's [set by default](https://gohugo.io/content-management/taxonomies/#default-taxonomies) (which I didn't know) and I didn't explicitly disable it.

**→ I suppose everyone who wants a custom configuration for their related content will first copy & paste the default configuration from the documentation, so IMO they should know that the configuration shown there right now is probably not enough, and that they need an index for `tags` as well.**

Maybe I should rephrase my changes? English is not my native language.

Issue #1685, opened 2022-03-16
Add information on deploying to Cloudflare Pages


arif254:
Reopening #1616. I recently discovered Cloudflare Pages after using Hugo based on the Netlify guide. So, I concur with the user of the closed issue that at least mention in the page someone can host on Cloudflare Pages and link to the documentation.

carerragt:
This needs to be closed. right?

Issue #1616, opened 2021-12-27
Add information on deploying to Cloudflare Pages


afrophi:
The [Hosting and Deployment page](https://gohugo.io/hosting-and-deployment/) could use a guide on how to deploy to Cloudflare Pages.

jmooring:
With limited exceptions, deployment documentation should be provided by the service provider. Cloudflare has already done this:

https://developers.cloudflare.com/pages/framework-guides/deploy-a-hugo-site

PR #1709, opened 2022-04-07
Hosting on Cloudflare Pages

callaloo:
Solves #1616 & #1685


coliff:
I agree with @jmooring that we can't/shouldn't add hosting guides for every provider but Cloudflare is one of the biggest players and I don't think it's a problem just having a short page with a link to their docs.

PR #1714, opened 2022-04-10
Update comments.md

callaloo:


bep:
Thanks, but we're currently not adding any more services without a very convincing argument, which I don't see here.

callaloo:
> but we're currently not adding any more services without a very convincing argument
then make that clear in your docs.

PR #1717, opened 2022-04-11
Fix broken anchor links

deining:
This PR fixes more than 20 broken anchor links in the hugo docs. These broken links where detected using the great [hyperlink](https://github.com/untitaker/hyperlink) tool. With current HEAD of repo, this tool reports:

```
$ hyperlink.exe public/ --check-anchors --sources content/en/
Reading files
Checking 100383 links from 1308 files (540 documents)
Found some broken links, reading source files
...
Found 233 bad links
Found 25 bad anchors
```
With this PR applied, this comes down to:

```
Found 228 bad links
Found 0 bad anchors
```

Please note that not all of these 228 bad links are actually bad links, a lot of them are covered by aliases, `hyperlink` produces false alerts for this kind of links.

**Idea:** What about integrating link checking as a part of the build process when committing of receiving PR (e.g. via GitHub actions)?

### DIFF
--- a/content/en/about/features.md
+++ b/content/en/about/features.md
@@ -60,7 +60,7 @@ toc: true
 [content summaries]: /content-management/summaries/
 [content types]: /content-management/types/
 [Disqus]: https://disqus.com/
-[Dynamic menu]: /templates/menus/
+[Dynamic menu]: /templates/menu-templates/
 [Extremely fast]: https://github.com/bep/hugo-benchmark
 [front matter]: /content-management/front-matter/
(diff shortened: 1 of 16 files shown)


danielfdickinson:
Interesting. The name of the program got me confused at first, since I use https://github.com/Munter/hyperlink in https://github.com/danielfdickinson/link-check-action-hugo-dfd which I use for the very purpose you mention. I also have scheduled tasks where I enable external link checking (I don't block push based on external link checks as that would be pretty unpredictable).

PR #1723, opened 2022-04-18
Delete hosting-on-bitbucket.md

coliff:
This article depends on https://www.aerobatic.com/ which is no longer online. Let's delete this page.


jmooring:
I agree this should be removed because, in its current form, it is useless.

However, since Bitbucket is a "notable" Git provider along with GitHub and GitLab, I think we would accept a PR that documents using Bitbucket Cloud with Bitbucket Pipelines.

PR #1731, opened 2022-04-24
Update name of command that enables printing i18n warnings

santiagoroman:
This was changed in commit 837fdfdf45014e3d5ef3b00b01548b68a4489c5f of
the hugo repository.

Note: The same command that was renamed is referenced in an example of the `hugo help` command to verify that hugo was installed correctly, but I saw that the output is not only outdated for this command but there are new ones that also do not appear there so I left it the way it is.


santiagoroman:
Closing as this was fixed on a separate PR, #1743.

PR #1732, opened 2022-04-24
Update quick-start.md

sifigi4335:
Add an option to create a blank theme from scratch

### DIFF
--- a/content/en/getting-started/quick-start.md
+++ b/content/en/getting-started/quick-start.md
@@ -58,7 +58,9 @@ The above will create a new Hugo site in a folder named `quickstart`.
 
 ## Step 3: Add a Theme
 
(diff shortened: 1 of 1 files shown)


jmooring:
Given what `hugo new theme` creates today, this should _not_ be part of the quick start guide. The quick start guide should allow a new user to produce an attractive, functional site with minimal effort. The quick start guide needs to just... work.

The `hugo new theme` command currently creates empty single.html and list.html files, so when you run `hugo server` and visit 
`http://localhost:1313/` all you see is "Page Not Found." That's not a good experience for any user, let alone a new one.

kiongomutwe:
As someone who knows some HTML/CSS but is unfamiliar with Go, I wanted to create my own theme. I wasted more than 6 hours on the quick start guide trying to import themes, while all I wanted was a blank theme to start building from. I suggest this be relooked upon.

jmooring:
Comments on closed issues and PRs are generally ignored. If you feel strongly about it, please create a new issue. Thanks.

PR #1733, opened 2022-04-28
Update hosting-on-github.md

tarzanwill:
After changing my baseURL to `/`, my website works fine finally.


bep:
Thanks, but this is not corret.

PR #1742, opened 2022-05-22
Update hugo.md

carerragt:


coliff:
I think we should replace the version number with this `{{ hugo.Version }}">` so it’s automatically updated.

coliff (approved):

PR #1746, opened 2022-05-30
Add link to security policy from getenv.md

fekete-robert:


content/en/functions/getenv.md
```diff
@@ -41,3 +41,6 @@ And then retrieve the values within a template:
 {{ os.Getenv "MY_VAR1" }} --> foo
 {{ os.Getenv "MY_VAR2" }} --> bar
 ```
+
+Note that starting with Hugo version 0.91.0, you must explicitly enable access to environmental variables. For details, see the getenv example in [Security Policy](/about/security-model/#security-policy).
```
jmooring, line 45:
```suggestion
With Hugo v0.91.0 and later, you must explicitly allow access to environment variables. For details, review [Hugo's Security Policy](/about/security-model/#security-policy). By default, environment variables beginning with `HUGO_` are allowed when using the `os.Getenv` function.
```

jmooring (changes requested):

PR #1749, opened 2022-06-01
Create hugo-serve.md

rbn30117:
document `hugo serve` command.

### DIFF
--- a/content/en/commands/hugo-serve.md
+++ b/content/en/commands/hugo-serve.md
@@ -1,0 +1,12 @@
+---
+title: "hugo serve"
+slug: hugo_serve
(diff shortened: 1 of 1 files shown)


jmooring:
Files in content/en/commands are generated from source via `hugo gen doc`. While `hugo serve` is an alias for `hugo server`, it is not referenced in:
- The documentation
- The CLI auto-completion options
- The CLI help output
- The CLI man pages

So there's no need to document its existence---think of it as a hidden (or even deprecated) command.

kiongomutwe:
I use this command a lot! I came across it first. `hugo server` sounds like a mouthful.

PR #1750, opened 2022-06-01
Update introduction.md

rbn30117:
using `resources.Copy` with `resources.GetRemote`


jmooring:
Thanks for this, but an another example is not required. The documentation already states:

> `resources.Copy` allows you to copy almost any Hugo `Resource` (the one exception is the `Page`)

If we were to include another example of GetRemote (including Copy), it would be elsewhere in the documentation, and more robust (defensive coding, error checking).

PR #1764, opened 2022-06-23
remove erroneous foo

coliff:
not sure why/how that got added....


bep:
>not sure why/how that got added....

I suspect it's me, by accident; I sometimes test the Hugo watcher with "echo foo > somefile.md" :-)

PR #1776, opened 2022-07-13
Update internal.md

tuthugo:
Fixes [this issue](https://discourse.gohugo.io/t/config-is-services-googleanalytics-id-an-alias-for-googleanalytics/39469/5)


tuthugo (commented):
Add a warning not to use both `googleAnalytics` and `services.googleAnalytics.id`.

PR #1778, opened 2022-07-13
Update configuration.md

tuthugo:


chalin (changes requested):
Replace the external links you've added (https://gohugo.io) by local paths.  Also please give this PR a more meaningful title.

Technically, those config params are not alias. I'll have to think more about how to describe the relationship.

PR #1797, opened 2022-08-17
Notify user about variable that escape some symbols

anakinsleftleg:


jmooring:
This PR contains some inaccuracies. There's a lot going on here including the effects of Goldmark's typographer extension and Go's [html/template](https://pkg.go.dev/html/template) package. Using `.Plain` and `.PlainWords` with a text output format (example: JSON) produces different results, as expected.

Let's take this discussion to the forum (https://discourse.gohugo.io/). Either create a new topic or reference an existing one. Thanks.

PR #1805, opened 2022-09-04
Better explanation of Kind, Session, Type, Layout on final layout path

dgasparri:
Previous version was not clear about what the various parameters would affect the final layout path.

Extracted info from output/layout.go and tried to clarify it, creating also a table of folders and files searched by Hugo in order of priority.

### DIFF
--- a/content/en/templates/lookup-order.md
+++ b/content/en/templates/lookup-order.md
@@ -21,7 +21,7 @@ toc: true
 
 ## Hugo Layouts Lookup Rules
 
-A typical folder/file structure for a site will look like the following: 
+A typical layout folder/file structure for a site will look like the following: 
 
 
 ```bash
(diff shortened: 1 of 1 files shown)


jmooring:
You obviously put a lot of work into this, but it is way too long and difficult to read. I tried to read it several times using the Netlfiy preview site, but my eyes just glazed over. In addition to being too long and dense, this syntax makes it  really difficult to read:

- `<LAYOUT>` (11 instances)
- `<ROOT FOLDER>` (10 instances)
- `<THEME>` (6 instances)
- etc.

This page needs to have a _brief_ introduction, lookup tables, and perhaps some simple examples at the bottom of the page. Users commonly want to know answers to these simple questions:
- Where do I put a template to override the list view for my posts?
- Where do I put a template to override the page view for my posts?
- Where do I put a template to override my home page?
- Where do I put a template to override my taxonomy pages?
- Where do I put a template to override my term pages?

I think your detailed explanation is better suited to a blog post or a "tips and tricks" article somewhere.

PR #1819, opened 2022-09-16
Update purgecss with same website guide example

MarlonLuan:
Update purgecss with same website guide example: https://purgecss.com/guides/hugo.html#postcss-config-file

### DIFF
--- a/content/en/hugo-pipes/postprocess.md
+++ b/content/en/hugo-pipes/postprocess.md
@@ -44,18 +44,23 @@ The below configuration will write a `hugo_stats.json` file to the project root 
 
 ```js
 const purgecss = require('@fullhuman/postcss-purgecss')({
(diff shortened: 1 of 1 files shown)


content/en/hugo-pipes/postprocess.md
```diff
@@ -44,27 +44,37 @@ The below configuration will write a `hugo_stats.json` file to the project root
...
+  ]
+};
 ```
 
 Note that in the example above, the "CSS purge step" will only be applied to the production build. This means that you need to do something like this in your head template to build and include your CSS:
 
 ```go-html-template
-{{ $css := resources.Get "css/main.css" }}
-{{ $css = $css | resources.PostCSS }}
+{{ $css := resources.Get "css/style.css" | resources.PostCSS }}
```
bep, line 69: These changes does not make much sense.

PR #1821, opened 2022-09-22
Remove outdated IE conditional comments example

coliff:
This only applied to IE9 and lower which hasn't been supported for a long time and has 0.0% browser marketshare.


bep:
Good catch.

PR #1825, opened 2022-10-01
docs: :memo: Add Hebrew README

YonLiud:
## Add Hebrew README file, part of translating the Hugo Docs project to Hebrew

As a daily user of Hugo and trying to "convert" my friends into using Hugo I found the need to have Hugo docs in Hebrew as well.

I will attempt to translate as much as possible as soon as possible!


jmooring:
Thanks for this, but at present we do not have trusted resources to review or maintain translations of the documentation. Again, thank you for your contribution.

PR #1826, opened 2022-10-01
Update search.md to include Pagefind

brycewray:


coliff:
Good suggestion - pagefind is great 👍

PR #1838, opened 2022-10-14
Update gomodules-info.html

RohitYandigeri:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=dhFePCa)


coliff:
Thanks for the PR but I think this is fine as an `<aside>` - what's the advantage in changing it to a div which has no semantic meaning?

PR #1843, opened 2022-10-18
docs: Link data templates to index function

CollierCZ:
Someone was having difficulty accessing their data in their front matter. They were looking at the [data templates page](https://gohugo.io/templates/data-templates) and not seeing any relevant help. When I linked to the [index function](https://gohugo.io/functions/index-function/) it was clear for them.

This PR is to try to help anyone else who might have the same issue.

### DIFF
--- a/content/en/templates/data-templates.md
+++ b/content/en/templates/data-templates.md
@@ -114,12 +114,21 @@ You can use the following code to render the `Short Description` in your layout:
 
 Note the use of the [`markdownify` template function][markdownify]. This will send the description through the Markdown rendering engine.
 
-## Use data in front matter
+## Keys as variables
 
-You might want to access different properties of your data in front matter.
-For example, you want to specify a `location` for a page with specific information about that location.
+You might want to access a key name as a variable.
+To do so, use the [`index` function][index] with the variable as the second parameter.
+```plaintext
+{{ $keyName := "Name" }}
+{{ index .Site.Data.User0123 $keyName }}
+```
 
-To do so, use the [`index` function][index].
+If the `Name` key is an object, you can also access its properties:
+
+```plaintext
+{{ $keyName := "Name" }}
+{{ (index .Site.Data.User0123 $keyName).propertyName }}
+```
 
 ## Get Remote Data
 
--- a/content/en/templates/data-templates.md
+++ b/content/en/templates/data-templates.md
@@ -114,6 +114,12 @@ You can use the following code to render the `Short Description` in your layout:
 
 Note the use of the [`markdownify` template function][markdownify]. This will send the description through the Markdown rendering engine.
 
+## Use data in front matter
+
+You might want to access different properties of your data in front matter.
+For example, you want to specify a `location` for a page with specific information about that location.
+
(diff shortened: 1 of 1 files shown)


content/en/templates/data-templates.md
```diff
@@ -114,6 +114,12 @@ You can use the following code to render the `Short Description` in your layout:
 
 Note the use of the [`markdownify` template function][markdownify]. This will send the description through the Markdown rendering engine.
 
+## Use data in front matter
+
+You might want to access different properties of your data in front matter.
+For example, you want to specify a `location` for a page with specific information about that location.
+
+To do so, use the [`index` function][index].
```
ZQ-PSK, line 122: That someone was me, thank you for providing the solution, @CollierCZ !
I will take the liberty of suggesting a change more specific to what the root cause of my problem was:
```suggestion
### Named values in variables

You might want to access the key name as a variable. To do so, use the [`index` function][index], with the variable as the second parameter.

{{< highlight plaintext >}}
{{ $keyName := "Name" }}
{{ index .Site.Data.User0123 $keyName }}
{{< /highlight >}}

If the `Name` key was an object, you could also access its properties:

{{< highlight plaintext >}}
{{ $keyName := "Name" }}
{{ (index .Site.Data.User0123 $keyName).propertyName }}
{{< /highlight >}}
```

I had to suggest the highlight shortcode here, using three backticks seems to collide with Github's suggestions.

CollierCZ: You can get around that on GitHub by using four backticks for the suggestion itself

CollierCZ:
````suggestion
## Named values in variables

You might want to access a key name as a variable.
To do so, use the [`index` function][index] with the variable as the second parameter.
```plaintext
{{ $keyName := "Name" }}
{{ index .Site.Data.User0123 $keyName }}
```

If the `Name` key is an object, you can also access its properties:

```plaintext
{{ $keyName := "Name" }}
{{ (index .Site.Data.User0123 $keyName).propertyName }}
```
````

File: content/en/templates/data-templates.md [hunk 2]
```diff
@@ -114,6 +114,21 @@ You can use the following code to render the `Short Description` in your layout:
 
 Note the use of the [`markdownify` template function][markdownify]. This will send the description through the Markdown rendering engine.
 
+## Keys as variables
+
+You might want to access a key name as a variable.
+To do so, use the [`index` function][index] with the variable as the second parameter.
+```plaintext
```
polarweasel, line 121: Markdown style suggestion to match the rest of the file... (Seems like they're not using language names on their fenced code blocks, either, which seems odd. I like that you provided one, although I bet calling these two `go` would get nice highlighting. :) )

````suggestion
You might want to access a key name as a variable. To do so, use the [`index` function][index] with the variable as the second parameter.

```plaintext
````

polarweasel: Also: need to define the link target for `[index]` at the bottom of the file with the others.

jmooring:
Superseded by https://github.com/gohugoio/hugoDocs/pull/1855

Issue #1857, opened 2022-10-29
Markdown formatting: Fenced code blocks should have a language specified


coliff:
In the Markdown linter we should set [MD040](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md040---fenced-code-blocks-should-have-a-language-specified)to true and specify languages for all the code snippets. This will give them nice syntax highlighted colors in the markdown and rendered HTML on the site.

Most of the code blocks with unspecified languages are either yml, html, or go-html-template.

PR #1861, opened 2022-10-31
Fenced code blocks should have a language specified

coliff:
Fixes: #1857


coliff:
this is a nice improvement aross the docs!
Before:
image
After:
image

PR #1869, opened 2022-11-07
Remove `#+KEY: VALUE` to avoid deprecated warning

tianheg:
If you use sth. like this in your org posts file:

```org
#+TITLE: ...
#+DATE: ...
#+TAGS: CSS
```

WARNING you may get:

```shell
Please use '#+tags[]:' notation, automatic conversion is deprecated.
```

### DIFF
--- a/content/en/content-management/front-matter.md
+++ b/content/en/content-management/front-matter.md
@@ -36,7 +36,7 @@ JSON
 
 ORG
 : a group of Org mode keywords in the format '`#+KEY: VALUE`'. Any line that does not start with `#+` ends the front matter section.
-  Keyword values can be either strings (`#+KEY: VALUE`) or a whitespace separated list of strings (`#+KEY[]: VALUE_1 VALUE_2`).
+  Keyword values can be a whitespace separated list of strings (`#+KEY[]: VALUE_1 VALUE_2`).
 
 ### Example
 


jmooring:
The existing documentation is correct. Org mode properties are scalar strings, assigned with this notation:

```
#+Foo: wibble wobble
```

```text
{{ .Params.foo | printf "%#v" }} --> "wibble wobble"
```

Hugo users often need to assign multiple values to a property, so [we added syntax](https://github.com/gohugoio/hugo/commit/fad183c4ae55069be9246e64ab1c8b2f43d08d06) that is unique to Hugo:

```text
#+Foo[]: wibble wobble
```

```text
{{ .Params.foo | printf "%#v" }} --> []string{"wibble", "wobble"}
```

However, for backwards compatibility, we allow you to specify `aliases`, `categories`, and `tags` using the scalar notation:

```text
#+Aliases: /posts/post-1/ /posts/post-2/
#+Categories: huey dewey
#+Tags: wibble wobble
```

```text
{{ .Params.aliases | printf "%#v" }}     --> []string{"/posts/post-1/", "/posts/post-2/"}
{{ .Params.categories | printf "%#v" }}  --> []string{"huey", "dewey"}
{{ .Params.tags | printf "%#v" }}        --> []string{"wibble", "wobble"}
```

When you use the scalar notation for `aliases`, `categories`, or `tags` we warn you that you should use the array notation instead. 

https://github.com/gohugoio/hugo/blame/master/parser/metadecoders/decoder.go#L244-L245

PR #1890, opened 2022-11-13
update Static Web Apps docs

coliff:
Azure Static Web Apps is a specific product so it's important to use that for the title rather than just 'Azure'.


jmooring:
@coliff I am in the middle of multiple edits in the hosting/deployment section. I have merged this, but please hold off on any additional edits in the sections. Thanks!

PR #1891, opened 2022-11-13
Replaced sudo in OpenBSD with doas

9glenda:
OpenBSD doesn't ship with sudo. It ships with the sudo alternative doas.


jmooring:
True, but with a virgin OpenBSD install:

```bash
doas pkg_add hugo
```

> doas: doas is not enabled. /etc/doas.conf: no such file or directory

So I installed sudo because I didn't want to learn yet another configuration system.

PR #1914, opened 2022-11-23
Improving function signatures (#98)

deining:
This PR improves the signature of various functions.

Closes #98.

PR #1916, opened 2022-11-23
Follow-up: improve function signatures (#1914)

deining:
This PR fixes two issues raised in the review of #1914.


jmooring:
I merged the other too soon. Sorry about that.

PR #1920, opened 2022-11-24
prevent Hugo Showcase template from showing on site

coliff:
Marking this as a `draft:` prevents it from displaying on the sidebar in the https://gohugo.io/showcase/ section


bep:
The only reason it's there is so people can see it.

PR #1925, opened 2022-11-29
enableSourceMap requires true value in dict

bradryanbice:
Doesn't it?


jmooring:
This `(not hugo.IsProduction)` will return `true` or `false`, so the existing example is correct.

PR #1928, opened 2022-12-06
Update quick-start.md

tyytytytyiigo:
minor addition for clarification

### DIFF
--- a/content/en/getting-started/quick-start.md
+++ b/content/en/getting-started/quick-start.md
@@ -35,7 +35,7 @@ You must also be comfortable working from the command line.
 
 Run these commands[^1] to create a Hugo site with the [Ananke] theme. The next section provides an explanation of each command.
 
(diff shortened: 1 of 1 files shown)


jmooring:
I understand your objective, but there's probably a better way of saying it.

> and prefer not to use Git

This is a bit confusing, as we have already told the user that they _must_ install Git.

Existing:

> If you are a Windows user, you must run these commands with [PowerShell](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-windows). You cannot use Windows Powershell, which is a different application, or the Command Prompt.

Suggestion:

> If you are a Windows user, you must run these commands with [PowerShell](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-windows). You cannot use Windows Powershell, which is a different application, or the Command Prompt. You may also use a Linux shell if available.

jafaican:
If you mention Git, then it confuses some users why you mention Powershell. At least, someone I sent the link did. I had to clarify to them to use Git by default. But that was an outlier.

PR #1933, opened 2022-12-18
docs: fix typo in taxonomy variables

VaibhavSys:
The actual variable seems to be `.Data.Term` instead of `.Data.Terms`


jmooring:
Superseded by #1935.

PR #1946, opened 2023-01-01
Added missing `/` in URL generator

arun-mani-j:
In the example to render tags and topics, we use:
```html
<a href="{{ "tags" | absURL }}{{ . | urlize }}">
```
But this will produce links like `www.example.com/tagsTerm`  where `Term` is a tag. So added the missing `/`.

Hope I didn't mistake :sweat_smile:

### DIFF
--- a/content/en/templates/single-page-templates.md
+++ b/content/en/templates/single-page-templates.md
@@ -49,14 +49,14 @@ This single page template makes use of Hugo [base templates], the [`.Format` fun
     {{ with .Params.topics }}
     <ul id="topics">
       {{ range . }}
-        <li><a href="{{ "topics" | absURL}}{{ . | urlize }}">{{ . }}</a> </li>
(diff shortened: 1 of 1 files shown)


jmooring:
That example code is, among other things, old. Please use this construct instead:

```text
{{ with .GetTerms "tags" }}
  <ul>
    {{ range . }}
      <li><a href="{{ .RelPermalink }}">{{ .LinkTitle }}</a></li>
    {{ end }}
  </ul>
{{ end }}
```

arun-mani-j:
I found that example in Taxonomy but if I search for `.GetTerms`, Algolia gives me only https://gohugo.io/templates/taxonomy-templates/#example-list-tags-in-a-single-page-template as result. Is there any other reference to .GetTerms function (like as a method of `.Page`)?

PR #1952, opened 2023-01-11
Fix example to convert image with same dimensions

azeemba:
The example shows how to convert image format without changing the image dimensions. This is similar to the example above about rotating an image without resizing.

However, the height/width were swapped. Height should come first (as it does in the rotation example)


jmooring:
The height and width are in the correct sequence. You only need to reverse them when rotating.

PR #1957, opened 2023-01-24
remove `version` from SVG example

coliff:
version is deprecated and is not needed

REF: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/version


jmooring (commented):
<https://www.w3.org/TR/SVG11/struct.html#SVGElementVersionAttribute>

No indication of deprecation. I have no idea if Mozilla is correct.

coliff:
It's safe to remove and will keep the example code shorter. Maybe it's not deprecated though, can't find a source for that - so up to you if you want to merge or close. No big deal, I'm fine with either.

PR #1970, opened 2023-02-14
Simplify ordinal abbreviation example...

kcastellino:
...by using the 'humanize' function to cut down template code

### DIFF
--- a/content/en/functions/format.md
+++ b/content/en/functions/format.md
@@ -95,7 +95,7 @@ More examples can be found in Go's [documentation for the time package][timecons
 
 Spelled-out cardinal numbers (e.g. "one", "two", and "three") are not currently supported.
 
-Ordinal abbreviations (i.e., with shorted suffixes like "1st", "2nd", and "3rd") are not currently directly supported. By using `{{.Date.Format "Jan 2nd 2006"}}`, Hugo assumes you want to append `nd` as a string to the day of the month. However, you can use the `humanize` function to create something like this:
+Use the [`humanize`](/functions/humanize) function to render the day of the month as an ordinal number:
 
 ```
 {{ humanize .Date.Day }} of {{ .Date.Format "January 2006" }}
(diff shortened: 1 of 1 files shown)


content/en/functions/format.md
```diff
@@ -95,10 +95,10 @@ More examples can be found in Go's [documentation for the time package][timecons
 
 Spelled-out cardinal numbers (e.g. "one", "two", and "three") are not currently supported.
 
-Ordinal abbreviations (i.e., with shorted suffixes like "1st", "2nd", and "3rd") are not currently directly supported. By using `{{.Date.Format "Jan 2nd 2006"}}`, Hugo assumes you want to append `nd` as a string to the day of the month. However, you can chain functions together to create something like this:
+Ordinal abbreviations (i.e., with shorted suffixes like "1st", "2nd", and "3rd") are not currently directly supported. By using `{{.Date.Format "Jan 2nd 2006"}}`, Hugo assumes you want to append `nd` as a string to the day of the month. However, you can use the `humanize` function to create something like this:
```
jmooring, line 98:
```suggestion
Use the [`humanize`](/functions/humanize) function to render the day of the month as an ordinal number:
```

jmooring (changes requested):

content/en/functions/format.md
kcastellino: Yeah, I thought later about adding a link. Do you think we need a sentence stating you can't use ordinals in layout strings, or is it unnecessary?

```suggestion
Ordinal abbreviations cannot be used directly; using `{{.Date.Format "Jan 2nd 2006"}}`, will append `nd` as a string to the day of the month. However, you can use the [`humanize`](/functions/humanize) function to render the day of the month as an ordinal number:
```

jmooring: > or is it unnecessary?

Unless users (plural) repeatedly make the same mistake, the documentation should describe what you must/should/can do. Because the list of what you must/should/can _not_ do is long.

kcastellino: True, I'll add your change

jmooring:
Thank you.

PR #1979, opened 2023-02-21
Add Related fragments config

bep:


jmooring:
This is covered in https://gohugo.io/content-management/related/.

PR #1981, opened 2023-02-22
Explain includeFiles with duplicate source, target

willfaught:


willfaught:
@gohugoio Ping

jmooring:
> If multiple mounts have the same source and target

What's the use case? This doesn't make any sense to me. Or are you describing something else?

```
[[module.mounts]]
source = 'static'
target = 'static'

[[module.mounts]]
source = 'a'
target = 'static'
includeFiles = 'a1.txt'

[[module.mounts]]
source = 'a'
target = 'static'
includeFiles = 'a2.txt'
```

willfaught:
Right, if the includeFiles are different. The workaround is to combine both includeFiles into one.

jmooring:
> The workaround is to combine both includeFiles into one.

That is not a workaround. 

image

willfaught:
I don't follow what you mean. I meant that

```toml
[[module.mounts]]
source = 'a'
target = 'static'
includeFiles = 'a1.txt'

[[module.mounts]]
source = 'a'
target = 'static'
includeFiles = 'a2.txt'
```

only mounts a1.txt with no error, but the workaround is

```toml
[[module.mounts]]
source = 'a'
target = 'static'
includeFiles = ['a1.txt', 'a2.txt']
```

The point is that it's documented that a2.txt will not be mounted in the first case.

jmooring:
This configuration is nonsensical:

```text
[[module.mounts]]
source = 'a'
target = 'static'
includeFiles = 'a1.txt'

[[module.mounts]]
source = 'a'
target = 'static'
includeFiles = 'a2.txt'
```

And this PR is closed.

willfaught:
I seem to have caused offense, for which I'm sorry. I was trying to make sure we were on the same page about what the reported issue was about.

PR #1985, opened 2023-03-01
Modify inner variable shortcode-template explanation

ginaesps:
Added "without the closing TAG by using the self-closing syntax" so that it is more comprehensible. Based on videos and blogs regarding the topic to make sure it makes sense.


ginaesps:
@bep Would really appreciate your feedback on this

jmooring:
@ginaesps Thank you for this. We will review it when we have time to review it. Pinging us for a review will not accelerate the process.

PR #1987, opened 2023-03-03
add headings to distinguish render hook context params

AlanBreck:
Looking at the docs for [render hooks](https://gohugo.io/templates/render-hooks/#render-hooks-for-headings-links-and-images), it's not immediately apparent which params apply to which hooks. At first glance, it appears as though there are duplicates, but that's only because the descriptions visually blend into the descriptions for each param.

I'm proposing headings to improve the distinction between the two sets of params.

### DIFF
--- a/content/en/templates/render-hooks.md
+++ b/content/en/templates/render-hooks.md
@@ -49,6 +49,8 @@ Some use cases for the above:
 
 ## Render Hooks for Headings, Links and Images
 
+### Context passed to `render-link` and `render-image`
+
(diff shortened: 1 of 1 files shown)


jmooring:
Thanks for this. It's a good idea, but there should be a separate subsection for each of the types:

- Code block render hook
  - Context
  - Methods
- Heading render hook (no methods)
- Image render hook
  - Context
  - Methods
- Link render hook (no methods)

Each one receives a slightly different context, and some have methods.

PR #1988, opened 2023-03-04
Add Winget installation method

sitiom:
Resolves https://github.com/gohugoio/hugo/issues/9201

### DIFF
--- a/content/en/installation/windows.md
+++ b/content/en/installation/windows.md
@@ -38,6 +38,16 @@ scoop install hugo-extended
 
 [Scoop]: https://scoop.sh/
 
+### Winget
+
+[Winget] is Microsoft's official free and open source package manager for Windows. This will install the extended edition of Hugo:
+
+```sh
+winget install Hugo.Hugo.Extended
+```
+
+[Winget]: https://learn.microsoft.com/en-us/windows/package-manager/
+
 {{% readfile file="/installation/common/04-docker.md" %}}
 
 {{% readfile file="/installation/common/05-build-from-source.md" %}}


jmooring:
See <https://github.com/gohugoio/hugo/issues/9201#issuecomment-1454789183>

sitiom:
Works for me. I don't know about you, but assuming you have an updated Windows 11 system, Winget should work out of the box. It is the *official* package manager by Microsoft, so this use case should be supported.

jmooring:
@sitiom Winget did _not_ work out-of-the-box on a fully updated Win 11 system. I finally got it working, by doing BOTH of the following:
- Fix corrupted PATH [^1]
- Reset app with elevated Power Shell

I figured this out because many others have had the same or similar problem.

Although this experience and time wasted leave a bad taste in my mouth, I am going to merge this. The package acceptance process for [github.com/microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs/) is similar to that of Scoop and Chocolatey.

[^1]: This alone was not enough. Updating PATH, then opening new shell: FAIL. Updating PATH, then rebooting: FAIL. I am guessing that an earlier version, or perhaps another app, clobbered the PATH, which is unexpected because this is _almost_ a virgin install.

sitiom:
> I figured this out because many others have had the same or similar problem.

Can you point me to the specific thread for this issue? I would like to reproduce this, although it worked out of the box for me.

jmooring:
<https://www.google.com/search?q=winget+not+recognized>

This issue closed. Further discussion should be taken to <https://discourse.gohugo.io/>

sitiom:
Your problem seems to be https://github.com/microsoft/winget-cli/issues/210, but that seems to be a problem with "App execution aliases" in general and not specifically Winget.

PR #1991, opened 2023-03-06
Update GitHub hosting instructions

### LINKED ISSUE #1990: Update GitHub hosting instructions
<https://gohugo.io/hosting-and-deployment/hosting-on-github/>

A confirmed problem was identfied by @hienvl125 in https://github.com/gohugoio/hugo/pull/10796.

We (I) need to do more testing before we make any changes. 

The Hugo "[starter workflow]" provided by GitHub runs without errors using these permissions:

[starter workflow]: https://github.com/actions/starter-workflows/blob/main/pages/hugo.yml
```
permissions:
  contents: read
  pages: write
  id-token: write
```

But that does not work with the `peaceiris/actions-hugo@v2` workflow in our documentation.

I am inclined to replace our example workflow file with the GitHub starter workflow:
- It addresses the permissions problem
- It includes installation of the Dart Sass transpiler
- It installs Node.js dependencies if `package-lock.json` or `npm-shrinkwrap.json` are present.

I think the clone is shallow (need to confirm), so we should add `depth 0` to the workflow.

Again, we (I) need to do more testing before we make any changes. 

---

 Related [GitHub documentation](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow).

jmooring:
Closes #1990


coliff:
:+1 Looks good at a glance, wonder if it's worth removing this frontmatter though:
```yml
date: 2014-03-21
publishdate: 2014-03-21
```

PR #1993, opened 2023-03-06
Updating Syntax of Links

AdamFermier:
Syntax of the links took me 15 min to experiment and get working - so trying to add some more here to help future people


jmooring:
Thanks for this, but markdown tutorials and explanations of markdown syntax are out of scope. A list of external markdown resources is provided here:
<https://gohugo.io/content-management/formats/>

PR #2006, opened 2023-03-17
Information concerning `sass` npm package

brycewray:
This mentions an alternative method of using Dart Sass with Hugo Pipes. *Technically*, it goes *around* Hugo Pipes — but the proposed edit corrects the statement that using the Embedded Dart Sass binary in one’s `PATH` is the only way to use Dart Sass. (It’s the *best* way, to be sure, but not the only way.)

### DIFF
--- a/content/en/hugo-pipes/scss-sass.md
+++ b/content/en/hugo-pipes/scss-sass.md
@@ -24,7 +24,11 @@ Any Sass or SCSS file can be transformed into a CSS file using `resources.ToCSS`
 
 transpiler [string]
 
-: The `transpiler` to use, valid values are `libsass` (default) and `dartsass`. If you want to use Hugo with Dart Sass you need to download a release binary from [Embedded Dart Sass](https://github.com/sass/dart-sass-embedded/releases) and make sure it's in your PC's `$PATH` (or `%PATH%` on Windows).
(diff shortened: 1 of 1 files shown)


jmooring:
So you would run this post-build?

brycewray:
@jmooring Actually, I run `sass` twice: (1.) before Hugo starts, to generate CSS files in `assets/css` (from `assets/scss`) so Hugo won’t error out when it runs; and (2.) again for Sass processing in general. See my admittedly convoluted scripting at:
https://github.com/brycewray/hugo-site/blob/main/package.json

. . . and I certainly welcome any correction about what I’m doing there, of course.

jmooring:
The transpiler has no knowledge of or access to Hugo's virtual file system (mounts), and cannot access the huge:vars module. Correct?

brycewray:
@jmooring TBH, I had to reference your [explanation](https://discourse.gohugo.io/t/initialize-sass-variables-from-hugo-templates/42053) to understand the question. Looks like the answer is: “Yes, that’s correct.” So, if that eliminates the underpinnings of my proposed edit, that’s completely understandable.

jmooring:
Given the limitations, and the questions they will raise, I'm going to close this.

PR #2009, opened 2023-03-19
Instruct the user to copy the example content

boosh:
Tell users how to see the theme's sample content so they don't only see the header and footer.


jmooring:
Thanks for this, but copying example content is out-of-scope.

image

boosh:
OK well having spent the afternoon trying to find a non-trivial theme to use as a base for my own, a note to this effect would have been helpful in the quickstart.

PR #2027, opened 2023-03-25
Specify target path caching for resources.ExecuteAsTemplate

willfaught:


content/en/hugo-pipes/resource-from-template.md
```diff
@@ -19,7 +19,7 @@ signature: ["resources.ExecuteAsTemplate TARGET_PATH CONTEXT RESOURCE"]
 
 In order to use Hugo Pipes function on an asset file containing Go Template magic the function `resources.ExecuteAsTemplate` must be used.
 
-The function takes three arguments: the target path for the created resource, the template context, and the resource object.
+The function takes three arguments: the target path for the created resource, the template context, and the resource object. The target path is used to cache the result.
```
jmooring, line 22:
```suggestion
The result is cached using the target path as the cache key.
```

jmooring (changes requested):

content/en/hugo-pipes/resource-from-template.md
willfaught: I found it useful for it to explain what the first param is. It would be too bad to lose that.

jmooring: Sorry, I meant to replace just the last sentence.

willfaught, line 22:
```suggestion
The function takes three arguments: the target path for the created resource, the template context, and the resource object. The result is cached using the target path as the cache key.
```

willfaught: @jmooring How's that?

jmooring: Thank you.

jmooring:
@willfaught If you have a moment, resources.FromString is also cached, using the target path as the cache key. Make similar clarification there?

willfaught:
Sure thing, will do it today or tomorrow.

PR #2048, opened 2023-04-08
Remove misplaced note in Hugo functions docs

rhcarvalho:
Remove misplaced note about `hugo.Generator`.

### DIFF
--- a/content/en/functions/hugo.md
+++ b/content/en/functions/hugo.md
@@ -16,6 +16,10 @@ relatedfuncs: []
 hugo.Generator
 : `<meta>` tag for the version of Hugo that generated the site. `hugo.Generator` outputs a *complete* HTML tag; e.g. `<meta name="generator" content="Hugo 0.63.2">`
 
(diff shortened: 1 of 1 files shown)


jmooring:
I think you should just remove the note.

By default, Hugo's injects something like this into the home page:
<meta name="generator" content="Hugo 0.111.3">

rhcarvalho:
Is the part that says "The generator tag allows the Hugo team to track the usage and popularity of Hugo" accurate?

If you feel like the note is not providing any value I'm okay with removing it altogether.

jmooring:
> Is the part that says "The generator tag allows the Hugo team to track the usage and popularity of Hugo" accurate?

Indirectly via https://builtwith.com/ and perhaps others, but we insert the generator tag by default, so...

> I'm okay with removing it altogether.

Yeah, that one.

rhcarvalho:
Done

jmooring:
Thank you.

PR #2052, opened 2023-04-11
Clarify module.replacements wording

willfaught:


willfaught:
Why is the syntax `->` instead of `=>` like in go.mod?

PR #2064, opened 2023-04-19
Clarify findRE, replaceRE, and findRESubmatch

### LINKED ISSUE #2063: Improve documentation of findRE and findRESubmatch
**findRE**: `The syntax of the regular expression is the same general syntax used by Perl, Python, and other languages.` This is at best misleading, at worst plain wrong. (Hu)Go's RE language is missing a lot of features other languages offer. At the very least, the sentence should be amended by `, but not all of the features from those languages are implemented`

**findRESubmatch**: The example code is probably borken – what are `§` signs supposed to stand for? If they mean anything, _what_? And what does this sentence

> In Hugo 0.110.0 we added a variant of findRE that returns a slice of strings holding the text of the leftmost match of the regular expression in s and the matches, if any, of its subexpressions.

mean? Where is `s`? What is `s`? What is the "leftmost match of the regular expression" – is that the _first_ match? Is it the first capturing group? What are "subexpressions" – capturing groups? In my opinion, introducing new terminology like "subexpression" does not help to clarify things. Simple question:

If I have the string `"ab"`, I use `` `a+(.*)` `` to match everything but the leading a – how can I access the content of the capturing group? What seems to work is something like this
``(index (index (findRESubmatch `a+(.*)` "ab" 0) 1)``
which seems to indicate that the return value of `findRESubmatch` is _not_ a "slice of strings" but a "slice containing a single element which is a slice of strings", something like [["ab" "b"]] in the example case. The first element of the inner slice would be the complete match, followed by the contents of the matched capturing groups. 

Perhaps it's technically necessary to return a slice of slices, but it certainly looks a bit awkward. In any case, the text should be amended to clarify how this return value is structured and how users can access the capturing group matches. Preferably with a _simple_ example that doesn't require deeper knowledge of regular expressions.

jmooring:
Closes #2063

### DIFF
--- a/content/en/functions/findRe.md
+++ b/content/en/functions/findRe.md
@@ -9,13 +9,25 @@ keywords: [regex]
 signature:
   - "findRE PATTERN INPUT [LIMIT]"
   - "strings.FindRE PATTERN INPUT [LIMIT]"
-relatedfuncs: [replaceRE]
+relatedfuncs: [findRESubmatch, replaceRE]
 ---
-By default, the `findRE` function finds all matches. You can limit the number of matches with an optional LIMIT parameter.
+By default, `findRE` finds all matches. You can limit the number of matches with an optional LIMIT parameter.
 
 When specifying the regular expression, use a raw [string literal] (backticks) instead of an interpreted string literal (double quotes) to simplify the syntax. With an interpreted string literal you must escape backslashes.
 
-The syntax of the regular expression is the same general syntax used by Perl, Python, and other languages. More precisely, it is the syntax accepted by [RE2] except for `\C`.
+[string literal]: https://go.dev/ref/spec#String_literals
+
+This function uses the [RE2] regular expression library. See the [RE2 syntax documentation] for details. Note that the RE2 `\C` escape sequence is not supported.
+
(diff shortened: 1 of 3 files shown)


chrillek:
Thanks for all the good work. I have two minor issues, though:
- The usage of "left-most" and "left to right" is misleading, since it should be "right-most" and "right to left" in some writing systems, notably Arabic and Hebrew. "left-most" could be replaced by "first" and "left to right" by "in the direction of writing", for example
- The statement that `findRESubmatch` returns a "slice of strings". In your last example, you show a JSON structure for the return value that looks roughly like this:
```json
[ 
["complete match", "capturing group"],
["complete match", "capturing group"]
]
```
And you're retrieving the strings in a `range` like so:
```
{ range $matches }}
  {{ index . 1 }}
{{ end }}
```
The `range` obviously works with a slice, and the inner `index` expects a slice or a dict as well. Which, at least to me, means that `$matches` is a _slice of slices of string_. As the JSON object is an `Array` of `Array`s of `String`s. Calling the return value of `findRESubmatch` a "slice of string" is, I think, not correct.

PR #2069, opened 2023-05-01
Update note for Windows users with detail about PowerShell versions

bradykelly:
The older text "PowerShell and Windows PowerShell are different applications" could be quite confusing for many Windows users. I Added info to make the difference between older 'Windows PowerShell' and the newer, required 'PowerShell' versions


jmooring:
Closing per comments on https://github.com/gohugoio/hugoDocs/issues/2070.

PR #2072, opened 2023-05-03
Update quick-start.md with more detail about PowerShell

### LINKED ISSUE #2070: The Quick Start section is very vague about the difference between PowerShell and Windows PowerShell.
The 'Create a site' section for Windows users says to use `PowerShell` and not `Windows PowerShell`. It goes on to be greatly helpful explaining the difference between these versions with the terse sentence, "PowerShell and Windows PowerShell are different applications."

For the average Windows user just wanting to build a simple website, that could be tremendously confusing. I think you should state that one must use the cross-platform version of PowerShell and supply a link to instructions for installing the correct PowerShell version.

bradykelly:
Added detail to the distinction between Windows PowerShell and the newer, cross-platform PowerShell. Closes #2070

### DIFF
--- a/content/en/getting-started/quick-start.md
+++ b/content/en/getting-started/quick-start.md
@@ -36,10 +36,10 @@ You must also be comfortable working from the command line.
 **If you are a Windows user:**
 
 - Do not use the Command Prompt
-- Do not use Windows PowerShell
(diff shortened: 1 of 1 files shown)


jmooring:
Duplicate of #2069

PR #2087, opened 2023-05-22
Add Hugo Module examples and remove 'brand new' text

dufcrule:
The latest version of Hugo is 0.111.3, so it feels that Hugo Modules are no longer "all very much brand new" since they were introduced in 0.56.0.

As well as removing that text, this PR also adds 2 more example projects which users who are new to Hugo Modules might find useful.


coliff (approved):

jmooring:
This will be addressed in https://github.com/gohugoio/hugoDocs/issues/2308

PR #2090, opened 2023-05-24
Change config.xxx to hugo.xxx throughout the documentation

jmooring:


DanielRuf:
Which hugo version supports this instead of `config.toml`?

Because with `sudo apt install hugo` I get a version that generates `config.toml`, not `hugo.toml`.

PR #2092, opened 2023-05-24
Update CLI documentation

jmooring:


bep:
We do have a command to keep that updated, but of course I stupidly added the `hugo config --format` thing after I updated the docs.

PR #2093, opened 2023-05-26
Update quick-start.md

StefanHolmes:
I followed the instructions and hit an error because the configuration was trying to load from config.toml and not hugo.toml


jmooring:
The name of the config file create by Hugo **v0.112.0 and later** is `hugo.toml`, though Hugo looks for config.toml as well (backwards compatible).

See the "prerequisites" bit?

image

PR #2114, opened 2023-06-23
Fireship.io uses Svelte instead of Hugo

vedantmgoyal2009:


jmooring:
image

vedantmgoyal2009:
I can't see Hugo in the "Languages" section on the source code repo (https://github.com/fireship-io/fireship.io), and the commit was authored a long time ago, so I thought that maybe the maintainer shifted to Svelte.

PR #2164, opened 2023-08-01
security: Allow to run mod commands on Windows

deining:
As discussed in this [thread](https://discourse.gohugo.io/t/hugo-mod-vendor-not-working-on-windows-v0-116-0/45543) in hugo discourse.


config/_default/security.toml
```diff
@@ -3,7 +3,7 @@
 
   [exec]
     allow = ['^go$']
-    osEnv = ['^PATH$']
+    osEnv = ['^PATH$|TMP']
```
bep, line 7: I suggest that we remove `allow` and `osEnv` config and just rely on the default settings for those, which has been improved a lot since the above was added.

bep (changes requested):

config/_default/security.toml
deining: Reverted to the default settings, as requested.

PR #2181, opened 2023-08-07
Replace docs.json with docs.yaml

bep:
See https://github.com/gohugoio/hugo/issues/11345


jmooring (approved):
Thank you. Looks great.

Issue #1888, opened 2022-11-13
Centralize and elevate configuration documentation 


jmooring:
This would be a major change, so we should not act on this without bep's approval.

Currently the configuration documentation is spread around throughout the site. I propose creating a section for configuration, with a page for each of the configuration tables: root, markup, taxonomies, menus, params, etc.

This section would be located near the bottom, as a reference. It would be nice if docs.json, generated from source, contained the defaults for all of the keys.

Also incorporate intent from:

- [x] <https://github.com/gohugoio/hugoDocs/issues/427>
- [x] <https://github.com/gohugoio/hugoDocs/pull/1781>
- [x] <https://github.com/gohugoio/hugoDocs/issues/1779>
- [x] <https://github.com/gohugoio/hugoDocs/issues/2154>
- [x] <https://github.com/gohugoio/hugoDocs/pull/2186>
- [x] https://github.com/gohugoio/hugoDocs/issues/1628

High level keys:

- [x] intro
- [x] all
- [x] build
- [x] caches
- [x] cascade
- [x] contentTypes ([#2922](https://github.com/gohugoio/hugoDocs/issues/2922))
- [x] deployment
- [x] frontmatter
- [x] httpcache
- [x] imaging
- [x] languages
- [x] markup
- [x] mediatypes
- [x] menus
- [x] minify
- [x] module
- [x] outputformats
- [x] outputs
- [x] page
- [x] pagination
- [x] params
- [x] permalinks
- [x] privacy
- [x] related
- [x] security
- [x] segments
- [x] server
- [x] services
- [x] sitemap
- [x] taxonomies
- [x] uglyurls

Other:

- [x] Add note about per-language configuration

bep:
>This section would be located near the bottom, as a reference. It would be nice if docs.json, generated from source, contained the defaults for all of the keys.

I think the above is true now.

As to this issue, I somehow didn't notice it before now, but I certainly agree that creating a section for `configuration` would make perfect sense.

PR #2186, opened 2023-08-09
Improve config merge section

salim-b:
Since https://github.com/gohugoio/hugo/issues/10620 was resolved, the config merging also applies to *language-specific* config.


jmooring:
Related to https://github.com/gohugoio/hugo/issues/11355

jmooring:
This will be addressed with https://github.com/gohugoio/hugoDocs/issues/1888

PR #2189, opened 2023-08-12
Create glossary of terms

### LINKED ISSUE #2184: Create a glossary of terms
Examples:
- slice
- array
- dictionary
- map
- section

Format depends on https://github.com/gohugoio/gohugoioTheme/pull/267.

jmooring:
Closes #2184

### DIFF
--- a/content/en/getting-started/external-learning-resources/index.md
+++ b/content/en/getting-started/external-learning-resources/index.md
@@ -5,8 +5,8 @@ keywords: [books, tutorials, learning, usage]
 menu:
   docs:
     parent: getting-started
-    weight: 60
-weight: 60
+    weight: 70
+weight: 70
 ---
 
 ## Books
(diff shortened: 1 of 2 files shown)


jmooring:
Appearance will improve when https://github.com/gohugoio/gohugoioTheme/pull/267 is merged.

This is _mostly_ accurate, though I did omit some details that site and theme authors really don't need to know (e.g., that a slice is backed by an array).

After going through this exercise, I believe we should stop referring to "site variables", "page variables", etc. Instead, we should call these "site methods" and "page methods". 

1. They are not variables. The values are immutable within template actions.
2. Why is `.Page.Title` a variable but `.Page.Pagination` a method? It's inconsistent.

I think we should reserve the word "variable" for user-defined variables within template actions... because they are variable.

I suspect the word choices we made 10 years ago were largely influenced by Jekyll.

bep:
I have merged/pushed the theme update, could you rebase this branch?

jmooring:
> I have merged/pushed the theme update, could you rebase this branch?

Thanks for theme merge. This is rebased, and I think I'm done futzing with it.

PR #2207, opened 2023-08-21
Don't hammer the snap store if not needed

cameronkerrnz:
My build failed today because the Snap Store was timing out today. In the workflow provided I saw it was installing dart-sass using snap, and also installing NodeJS.

I didn't think I needed either for my Hugo site, so I commented them out; the build worked nicely and was much faster too.

### DIFF
--- a/content/en/hosting-and-deployment/hosting-on-github/index.md
+++ b/content/en/hosting-and-deployment/hosting-on-github/index.md
@@ -171,6 +171,9 @@ Under the deploy step, you will see a link to your live site.
 
 In the future, whenever you push a change from your local repository, GitHub will rebuild your site and deploy the changes.
 
+Step 11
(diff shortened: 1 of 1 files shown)


jmooring:
> If you aren't using your own SCSS, you won't need dart-sass or NodeJS installed.

While I understand the motivation behind this PR, the statement above is incorrect on several levels. 

> your own SCSS

If a theme transpiles Sass to CSS via Dart Sass, and the resulting CSS is not in the theme's resources directory (or if the site author has set `useResourceCacheWhen` to `never`), the build will fail if Dart Sass is not installed.

> or NodeJS installed

1. The GitHub Action does not install Node.js; that is already included in the container. The GitHub Action simply looks for a package file, then installs (via `npm ci`) the required packages, if any. The step is skipped if there isn't a package file.
1. If a theme autoprefixes (for example) CSS using PostCSS, and the resulting CSS is not in the theme's resources directory (or if the site author has set `useResourceCacheWhen` to `never`), the build will fail if the Node.js packages are not installed.
1. NodeJS may be required to include third-party CSS, Sass, and/or JavaScript.

> These steps add a significant amount of the time to the build process

No, they don't. 

image

In a separate PR I will add a note about optionally commenting out the Dart Sass installation step.

PR #2230, opened 2023-09-08
remove duplicate go-html-template in chroma lexer

yoserizalfirdaus:
this PR to remove duplicate go-html-template in chroma lexer


jmooring:
The data file is generated from code. This is an upstream issue. See <https://github.com/alecthomas/chroma/issues/843>.

PR #2231, opened 2023-09-09
Fix small typo

LITUATUI:
ES instead of DE.


jmooring:
Thanks for the catch, but it needs to match the directory tree above. Change one or the other; I have no preference.

PR #2233, opened 2023-09-18
Fix minor nitpicks

salim-b:
BTW, the new Glossary is an awesome effort! ❤

### DIFF
--- a/content/en/getting-started/glossary.md
+++ b/content/en/getting-started/glossary.md
@@ -192,9 +192,9 @@ A [template](#template) that overrides standard markdown rendering. See [details
 
 ### resource
 
-Any file consumed by the build process to augment or generate content, structure, behavior, or presentation. For example: images, videos, content snippets, CSS, Sass, Javascript, and data.
+Any file consumed by the build process to augment or generate content, structure, behavior, or presentation. For example: images, videos, content snippets, CSS, Sass/SCSS, JavaScript, and data.
 
-Hugo supports three types of resources: page resources (located in a [page bundle](/getting-started/glossary/#page-bundle)), global resources (located in the assets directory), and remote resources (typically accessed via https).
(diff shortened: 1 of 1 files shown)


content/en/getting-started/glossary.md
```diff
@@ -192,9 +192,9 @@ A [template](#template) that overrides standard markdown rendering. See [details
 
 ### resource
 
-Any file consumed by the build process to augment or generate content, structure, behavior, or presentation. For example: images, videos, content snippets, CSS, Sass, Javascript, and data.
+Any file consumed by the build process to augment or generate content, structure, behavior, or presentation. For example: images, videos, content snippets, CSS, Sass/SCSS, JavaScript, and data.
```
jmooring, line 195:
```suggestion
Any file consumed by the build process to augment or generate content, structure, behavior, or presentation. For example: images, videos, content snippets, CSS, Sass, JavaScript, and data.
```

jmooring (changes requested):
Sass has tho syntaxes: indented and SCSS.

content/en/getting-started/glossary.md
salim-b: I know.

But nowadays, SCSS (also called "Sassy CSS") is mostly used, while the indented syntax is rarely recommended anymore. It's a common source of confusion that the Sass authors just kept using the name "Sass" for both formats.

Since both "Sass" and "SCSS" are also used as the respective file extensions, whenever in doubt, "Sass" refers to the indented syntax, while "SCSS" refers to the newer CSS-backwards-compatible syntax.

Thus I think it makes sense to use both terms. But feel free to revert my change if you disagree.

jmooring: When the Sass team changes their terminology, we will change ours. Please make the change I requested.

PR #2239, opened 2023-09-25
Explain that markdownify results are safe

willfaught:


jmooring:
See my comments on your other PR:
<https://github.com/gohugoio/hugoDocs/pull/2240#issuecomment-1739623630>

jmooring:
This is about specifying return types for each function, which is in the works.

PR #2240, opened 2023-09-25
Explain that replaceRE results are unsafe

willfaught:


jmooring:
I'm not sure what "unsafe" means, because at face value that is not true.

It seems like this PR might be about specifying that this string function returns a string, not `template.HTML`. Is that correct?

willfaught:
"Unsafe" as in not the result of safeHTML, as in HTML content is escaped. In other words, `someHTMLString | safeHTML | replaceRE <x> <y>` needs to pipe through another safeHTML to be unescaped when printed.

willfaught:
replaceRE "strips" the safety marked by safeHTML that is on its input.

jmooring:
A function that returns a string returns a string. I think there are a couple of actionable items:

1. Specify return types for each function, which is in the works.
2. Provide a better description of what Go's `html/template` package does, and how that relates to the `safe.Whatever` functions.

PR #2241, opened 2023-09-25
Update quick-start.md, changing 'hugo.toml' to 'config.toml'

legallyahc:
Was encountering an error following the directions to start a Hugo site; the instructions added the theme information to 'hugo.toml' instead of the necessary 'config.toml'. Feel free to double check this, I'm quite new to the project.


jmooring:
From the quick start guide...

image

See the version requirement. Although config.toml works with v0.112.0 and later, the `hugo new site` and `hugo new theme` command generate a hugo.toml file with v0.112.0 and later.

Issue #2000, opened 2023-03-16
relLangURL doc doesn't specify the return type


willfaught:
It's implied it's a string, or at worst a url.URL:

>Returns a relative URL with a language prefix, if any.

However, it appears to be template.HTML. Is that correct?

bep:
>However, it appears to be template.HTML. Is that correct?

That is correct (and `template.HTML` is a `string`)

I'm (in this moment) working on a new and complete (auto generated) reference section for the Hugo Docs where all functions, methods and types should be documented.

willfaught:
It's surprising that a "URL" is represented by something called template.HTML, even if it's just a string under the hood.

This breaks calling .GetPage with the output of relLangURL. Perhaps relLangURL could be made to convert template.HTML to string, rather than we having to call string explicitly?

jmooring:
@willfaught 

> Perhaps relLangURL could be made to convert template.HTML to string, rather than we having to call string explicitly?

`ref`, `relref`, `relURL`, `relLangURL`, `absURL`, and `absLangURL` return `template.HTML`. The original implementations returned a string. It in unclear when/why the change was made, but it may have happened when the template functions were namespaced.

I've been thinking about this, and can't come up with a good reason for the current data type. We added the `urls.JoinPath` a few months ago, which returns a string. 

Although I've never had a problem with the current data type, create an issue if you feel strongly about it.

willfaught:
I'll create an issue. In the meantime, this issue can be for specifying the existing return type of relLangURL.

PR #2249, opened 2023-09-29
Specify return type of relLangURL

willfaught:
Fixes https://github.com/gohugoio/hugoDocs/issues/2000


jmooring:
This is about specifying return types for each function, which is in the works.

And the return type for the `ref`, `relref`, `relURL`, `relLangURL`, `absURL`, and `absLangURL` functions may change pending acceptance of https://github.com/gohugoio/hugo/issues/11511.

PR #2259, opened 2023-10-09
Update 04-docker.md

floryn90:
Added new repo that provides updated Docker images of latest Hugo versions


jmooring:
Thanks for this, but we're going to decline the addition. There are many Docker images for Hugo, by many authors. We're not going to list all of them, and there are trust issues to consider as well. I think any update to the documentation depends on the resolution of <https://github.com/gohugoio/hugo/issues/10760>.

Images authored and maintained by the Hugo project would be ideal, but we have limited resources and, at the moment, higher priority issues.

Thanks again.

PR #2271, opened 2023-10-16
Fix ESBuild sourcemap option

robsonsobral:
The [ESBuild docs says `sourcemap`](https://esbuild.github.io/api/#sourcemap), but Hugo Docs says `sourceMap`.


jmooring:
Either one works. Does this really matter?

PR #2273, opened 2023-10-17
Replace getCSV with getJSON in GetJSON.md

samwilson:
Fix what appears to be a copy-paste typo with getCSV being used in the examples rather than getJSON.


samwilson:
Oh oops, I should've checked first! This has already been proposed by @pawelgrzybek in #2256.

PR #2280, opened 2023-10-25
Mention Go variable naming rules

salim-b:
https://www.w3schools.com/go/go_variable_naming_rules.php is not official Go doc, but the best resource a quick web search revealed.


jmooring:
Please point to this (just added):
https://gohugo.io/getting-started/glossary/#identifier

If you follow the link, you'll see the keywords just below, stating...

> The following keywords are reserved and may not be used as identifiers.

Issue #1573, opened 2021-11-15
Truncate the /tools section


bep:
I had a look at the tools section

https://gohugo.io/tools/

And that needs some gardening. Looking through it, 1) I have never heard of 90% of the items 2) I suspect many of them aren't "notable enough".

I understand that people want a link to their project they have hard on, but we cannot be a link collection.

I suggest that we start a little fresh -- make "Tools" a one-pager and remove everything but the obvious (not sure what that should be, but I can certainly try if we agree on this).

/cc @jmooring

coliff:
I agree! I think instead it could be useful to include a link to something like:
https://github.com/topics/hugo

That way users can see all projects tagged with Hugo and sort by Best match, most stars, most recently updated etc.

danielfdickinson:
I agree the Tools page is better kept short. As I commented to @jmooring in [#1512#issuecomment-1079740244](https://github.com/gohugoio/hugoDocs/pull/1512#issuecomment-1079740244) I think that there should be a pointer to a 'wiki' post or a discussion on the [Hugo forum](https://discourse.gohugo.io/).

I agree with him that this is better done as full HOWTOs elsewhere and the forum seems the best place to me.

jmooring:
Also move https://gohugo.io/content-management/comments/ to the tools section, or whatever it will called once trancated and regorganized. The Disqus setup is referenced elsewhere in the documentation.

PR #2286, opened 2023-10-28
Add Glotta to Other Tools

simeononsecurity:
Added a tool @1nf053c and I developed to create translated index bundles for hugo sites.

### DIFF
--- a/content/en/tools/other.md
+++ b/content/en/tools/other.md
@@ -13,6 +13,7 @@ weight: 60
 
 And for all the other small things around Hugo:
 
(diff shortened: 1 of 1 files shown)


jmooring:
Per https://github.com/gohugoio/hugoDocs/issues/1573 we will either prune or remove the tools section.  You are welcome to create a "Tips and Tricks" posting on the [forum](https://discourse.gohugo.io/), and I suggest you tag your repository with the word "hugo" if you haven't already done so.

PR #2051, opened 2023-04-09
Add params variable in shortcodes

pratikmota:
Closes #2018

Adding details screenshots:
image

Issue #2196, opened 2023-08-13
Use "method" instead of "variable" for methods 


jmooring:
Recently, I spent a significant amount of time attempting to define the "language of Hugo" while creating a [glossary of terms](https://gohugo.io/getting-started/glossary/). This was a challenging exercise, as I had to reconcile current usage with terms that are specific and (mostly) narrowly defined. This is comparable to, but much less time-consuming than, writing the Go language specification ten years after the language was first released.

Although our language choices have, for the most part, been logical and consistent, I think it's time to use the term "method" instead of "variable", with the exception of variables initialized within template actions (e.g., `{{ $myVar := 42 }}`).

Justification:

1. Things like "page variables" are not variable. They are immutable within a template.
2. Our current usage is confusing. We call `.Page.Title` and `.Page.IsSection` variables, but we call `.Page.RenderString` a method. 
3. This would cleanup the "Functions" section. Although we call `.Page.RenderString` a method, we document it under the "Functions" section.

We've already done this for [`.Pages`](https://gohugo.io/variables/pages/). Although under the "Variables" section, the title is "Pages methods".

#### Tasks

Completed on or before 4 Nov 2023:

- [x] Namespace all functions (effort: medium)
- [x] Create Methods section (effort: low)
- [x] Move methods (anything that begins with a dot) from Functions to Methods (effort: low)
- [x] Add and namespace methods on Page, Site, etc. (effort: high)
- [x] Create quick reference guides for functions and methods (effort: medium)
- [x] Add notice to all Variables pages, pointing readers to the Methods quick reference guide and the Methods section

Completed on 19 Feb 2024:

- [x] Remove Variables section after a few months

Completed on 8 Mar 2024:

- [x] Rework the [introduction to templating](https://gohugo.io/templates/introduction/)

Completed on 15 Jun 2024:

- [x] Ensure consistent use of "function", "method", and "variable" throughout the documentation (effort: high)
- [x] Use "argument" instead of "parameter" when referring to function, method, and shortcode arguments (effort: medium)

I've been thinking about doing this for a while: <https://github.com/gohugoio/hugoDocs/issues/1639>

bep:
So, it's still very much on my TODO list to get an auto generated reference section happening in Hugo. The recent config consolidation was a important part of this, and then I may have gotten distracted.

With that setup you would get this "Reference":

* Function namespaces
* Objects (Page, Site etc.)
* Configuration
* Go Types

[...]

PR #2296, opened 2023-11-02
Namespace functions and methods

### LINKED ISSUE #53: Functions should live in namespaces (sub-sections)

### LINKED ISSUE #447: Document the .Sub, .After and other calculation functions that can be used for time
- https://godoc.org/time#Time
- https://discourse.gohugo.io/t/convert-a-string-to-a-time-datetime/11515/3
- https://discourse.gohugo.io/t/parsing-dates-in-templates/603/15
- https://gohugohq.com/howto/compare-date-strings-in-hugo/#calculations

### LINKED ISSUE #546: Clarify the difference in behavior of Site vs Page GetPage
For a path like `/foo/bar/abc.md`, `.Site.GetPage "abc"` will work (assuming that the "abc" name is unique in the whole site). 

But given that the /foo page context is stored in `$foo`, `$foo.GetPage "abc"` is **not** expected to always? work. 

The Page version of GetPage expects either an absolute path (starting with /) or a valid relative path **from the current context**. It is not designed to do a "lookup" anywhere in the given context. 

This distinction needs to be made explicit in the docs. 

- [Ref][1]
- [PR for the Page version GetPage][2]

[1]: https://discourse.gohugo.io/t/ambiguous-reference-in-new-getpage/13161/30?u=kaushalmodi
[2]: https://github.com/gohugoio/hugo/pull/4796

### LINKED ISSUE #1163: The list of available emojis is inaccurate

### LINKED ISSUE #1400: resources.PostProcess has no alias but documentation disagrees
https://gohugo.io/hugo-pipes/introduction/#method-aliases explains that every `resources` transformation has an alias except for those explicitly listed, e.g. `resources.FromString`. `PostProcess` isn't listed but doesn't have an alias. AFAIK *tpl/resources/init.go* defines the aliases as `fingerprint`, `minify`, `toCSS`, `postCSS`, and `babel`.

Either an alias should be added or the list of exceptions updated. (I think it's wrong to have an exception list in the documentation. Instead, it could say aliases exist for some of them and they're each documented on the individual page for the transform. Otherwise, when reading about PostCSS, I don't know there's an alias.)

### LINKED ISSUE #1406: Rework the PostProcess page
<https://gohugo.io/hugo-pipes/postprocess/>

### LINKED ISSUE #1547: Functions Quick Reference is truncated
https://gohugo.io/functions/ contains an incomplete list of functions.  It stops at `safeCSS`, and is missing `safeHTML` and everything after it.

On a large display, clicking "Functions" on the left expands a complete list of functions.  However, on a small display (eg. mobile), clicking "Functions" takes you to the Quick Reference, and it appears to be impossible to see the complete list of functions.

### LINKED ISSUE #1622: transform.Unmarshal for XML docs needs clarification

### LINKED ISSUE #1639: Please document .GetTerms
`.GetTerms` is not listed in the page variables, only in an example snippet on the Taxonomy Templates doc page.

Can you please document it?

### LINKED ISSUE #1667: Document .Page.Codeowners
https://github.com/gohugoio/hugo/pull/9529

### LINKED ISSUE #1745: Document boolean operators in their own page
Version 0.92.0 broke several "ne" function occurrences in my theme. "not" works, but it is undocumented. Perhaps document the boolean operators in their own page like [here](https://hugoif.github.io/hugo-by-example/operators/boolean/).

### LINKED ISSUE #2001: .Fragments need clearer explanation
> The .Fragments method returns a list of fragments for the current page.

is too short, self-referring and redundant to help much. Readers would like to know what "fragments of a page" even _are_. Something like

> The .Fragments method returns HTML elements or attributes from the rendered content of the current page. Currently, it can deliver all headings (`.Fragments.Headings`) and ID attributes (`Fragment.Identifiers`). 

Then, under the heading `.Headings`, the sentences

> A recursive list of headings for the current page. Can be used to generate a table of contents.

might be rewritten as 

> .Fragments.Headings returns a list of all headings for the current page in order of appearance. This list can be used to generate a table of contents. 

The term "recursive" is at best misleading (I do _not_ see any recursion in a simple list of all `h*` elements on a page), at worst irritating.

### LINKED ISSUE #2018: variables/shortcodes.md: .Params variable is missing

### LINKED ISSUE #2264: Explicitly state sort order within groups when grouped by a date
Applicable to:

```text
GroupByDate
GroupByExpiryDate
GroupByLastmod
GroupByParamDate
GroupByPublishDate
```

For example, if you group by expiration date, the pages within each group are sorted by expiration date.

### LINKED ISSUE #2275: `range` page: example for `dict` missing
The `range` page has an example using `range` with slice, followed by the headline "Maps" (https://gohugo.io/functions/go-template/range/#maps). Which again ranges over a `slice` (albeit one consisting of `dict`s). 

I'd expect something like 
```
{{ range $key, $val := (dict "key1" "value1" "key2" "value2" …}}
```
under the heading "Maps".

### LINKED ISSUE #2276: Term `collection` used differently for `range` and `append`
`range` is specified as
`range COLLECTION` with the explanation "Iterates over slices, maps, and page collections" Which must be interpreted as "`COLLECTION` is a synonym for `map`, `slice` and `page collection`" (which seems a bit cyclic, as it does not answer what the "collection" in "page collection" is...).
OTOH, `collections.append` is specfied as
`COLLECTION | collections.Append COLLECTION ⟼ any` with the explanation "Appends one or more elements to a slice and returns the resulting slice". So, in _this_ context, `COLLECTION` means `slice`, and only slice.

In the glossary, the term "collection" is explained as

>Typically, a collection of pages, but may also refer to an [array](https://gohugo.io/getting-started/glossary/#array), [slice](https://gohugo.io/getting-started/glossary/#slice), or [map](https://gohugo.io/getting-started/glossary/#map). For example, the pages within a site’s “articles” section are a page collection.

As before, defining a "collecting" by a "page collection" is cyclic. I'd suppose that a "page collection" is an "array of page objects". 

In any case, the term "collection" (or whatever the super-term for `map`, `slice` and `array` is called) should always refer to the same things. Having different meanings as in `range` and `append` is confusing.

jmooring:
Closes #53
Closes #447
Closes #546
Closes #1163
Closes #1400
Closes #1406
Closes #1547
Closes #1622
Closes #1637
Closes #1639
Closes #1667
Closes #1745
Closes #2001
Closes #2018
Closes #2051
Closes #2256
Closes #2264
Closes #2275
Closes #2276
Closes #2297


jmooring:
This is imperfect, but I think more complete than what we have today. This addresses, but does not resolve, https://github.com/gohugoio/hugoDocs/issues/2196. 

The UI changes are minimal, though I had to remove the stickiness from the right sidebar due to the number of elements in some of the section menus. For example, look at the right sidebar on the [`GetTerms`](https://deploy-preview-2296--gohugoio.netlify.app/methods/page/getterms/) page. 

The site seems to view well on mobile devices.

There's some duplication between the "Methods" pages and some of the "Content Management" pages, but we can clean that up over time.

Ultimately I'd like to remove the "Variables" section, but for now the important ones simply list the relevant methods using a shortcode.

PR #2322, opened 2023-11-11
templates/internal.md: use `ID` key name for `googleAnalytics`

chalin:
The proper field according to the docs is [`.Site.Config.Services.GoogleAnalytics.ID`](https://gohugo.io/templates/internal/#use-the-google-analytics-template), not <code>.Site.Config.Services.<strong>GoogleAnalytics.GoogleAnalytics</strong></code>. This fixes what is probably a search/replace error introduced via #2296?


jmooring:
Thank you.

Issue #2252, opened 2023-09-29
Provide better overview of what Go's html/template package does


jmooring:
...and how that relates to the `safe.Whatever` functions and the resulting data types.

PR #2329, opened 2023-11-14
improve safe function document

ngdangtu-vn:
This commit is the result of my mistake. I would like to note it down so other people wouldn't step on it like I did. I also attempt to define what safe means in `safe/_index.md`. I would like to have proofreadings a few time of my updates.

### DIFF
--- a/content/en/functions/safe/HTMLAttr.md
+++ b/content/en/functions/safe/HTMLAttr.md
@@ -52,4 +52,35 @@ To indicate that the HTML attribute is safe:
 As demonstrated above, you must pass the HTML attribute name _and_ value through the function. Applying `safeHTMLAttr` to the attribute value has no effect.
 {{% /note %}}
 
+## Inside partial template
+
+Even if you mark the attribute is safe with `safeHTMLAttr` inside a partial template, you still have to do it again when calling the template.
+
+```go-html-template
+<!-- partials/a-href.html -->
+{{ printf "href=%q" . | safeHTMLAttr }}
+```
+
+```go-html-template
+<!-- index.html -->
+{{ range site.Menus.main }}
+  <a href="{{ partial "a-href.html" .URL | safeHTMLAttr }}">{{ .Name }}</a>
+{{ end }}
+```
+
+So a good practice is never make a partial template for an attribute, instead, go for at least 1 element.
+
+```go-html-template
+<!-- partials/li.html -->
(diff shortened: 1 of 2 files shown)


jmooring:
Thanks for sharing your experience; it pointed out two deficiencies in the documentation that we just addressed with https://github.com/gohugoio/hugoDocs/pull/2330:

1. Documentation for the [`partial`] and [`partialCached`] functions did not explain the returned data type. Although "any" is accurate, it lacked important details.
2. The [`return`] statement was not documented in the functions/go-templates section.

To address your experience...

First, for the example you provided, you can use either [`safeURL`] or [`safeHTMLAttr`]. For example:

```text
<!-- layouts/_default/baseof.html -->
{{ range site.Menus.main }}
  {{ partial "li.html" . }}
{{ end }}
```
```text
<!-- layouts/partials/li.html -->
<li>
  <a href="{{ .URL | safeURL }}">{{ .Name }}</a>
</li>
```

Second, the reason your first attempt failed is that, without a `return` statement, the `partial` function returns data type `template.HTML`, a string declared as safe for Go's [html/template] package. The revised documentation makes this clear. Regardless how you cast data within a partial template, without a `return` statement it returns `template.HTML`.

Third, your first attempt would have worked if you had _returned_ a value from the partial instead of _rendering_ the partial. The return statement is now documented [here][`return`], and we have provided more examples on the [`partial`] and [`partialCached`] pages. 

This would have worked for you:

```text
<!-- layouts/_default/baseof.html -->
{{ range site.Menus.main }}
  <li>
    <a {{ partial "href.html" .URL }}>{{ .Name }}</a>
  </li>
{{ end }}
</ul>
```
```text
<!-- layouts/partials/href.html -->
{{ return (printf "href=%q" . | safeHTMLAttr) }}
```

Fourth, we will improve the documentation regarding Go's [html/template] package in https://github.com/gohugoio/hugoDocs/issues/2252.

Finally, when documenting a function or method, we document the one thing it can do, not all the things it cannot do. An exception to that is surprising behavior, which was addressed in https://github.com/gohugoio/hugoDocs/pull/2330.

Again, thanks for bringing this to our attention.

[`partial`]: https://gohugo.io/functions/partials/include/
[`partialCached`]: https://gohugo.io/functions/partials/includecached/
[`return`]: https://gohugo.io/functions/go-template/return/
[`safeHTMLAttr`]: https://gohugo.io/functions/safe/htmlattr/
[`safeURL`]: https://gohugo.io/functions/safe/url/
[html/template]: https://pkg.go.dev/html/template

PR #2332, opened 2023-11-16
Fix wrong command

rmoff:
```
hugo new content posts/my-first-post.md
Error: failed to resolve "content" to a archetype template
```


jmooring:
You are using an older version of Hugo. The documentation is correct. If you need assistance with upgrading, please create a topic in the forum (https://discourse.gohugo.io/). Thanks.

rmoff:
oops. my bad, sorry for the PR noise.

PR #2351, opened 2023-12-04
Add Gentoo installation instructions

rahilarious:


jmooring (changes requested):
Required changes:

1. The distributions are, obviously, listed alphabetically. Please place this is the correct position on the page.
2. You cannot include markdown reference links in a fenced code block. Please preview your changes locally before submitting pull requests.
3. According to the Gentoo documentation, you cannot install the extended edition of Hugo. You can install the standard edition plus Sass support via LibSass, but you cannot include WebP support. Please update the description.

rahilarious:
> Required changes:
> 
>     1. The distributions are, obviously, listed alphabetically. Please place this is the correct position on the page.

Okay done. Obvious things are not obvious to me
>     2. You cannot include markdown reference links in a fenced code block. Please preview your changes locally before submitting pull requests.
> 
I did try to build the whole site locally using official instructions of `hugo server` on README.md but it was throwing some `<{{ img` error. I'm just starting out with hugo. Removed that link itself. Didn't make much sense anyway.
>     3. According to the Gentoo documentation, you cannot install the extended edition of Hugo. You can install the standard edition plus Sass support via LibSass, but you cannot include WebP support. Please update the description.

Yes you can install extended edition of Hugo by enabling `sass` USE flag and about the webp support is there too (unconditionally). [src](https://github.com/gentoo/gentoo/blob/06001d963251a6b3fb59d2a17ad7a695789e70f0/www-apps/hugo/hugo-0.119.0.ebuild#L32) 

I've requested some changes from package maintainer of Gentoo. https://bugs.gentoo.org/919134

PR #2357, opened 2023-12-07
Update Gentoo installation instructions

rahilarious:
Replaced `sass` USE flag with `extended` in
https://github.com/gentoo/gentoo/pull/34149
which is merged now.


jmooring:
Thank you.

PR #2366, opened 2023-12-18
Fix default value of `pluralizeListTitles`

royarg02:
The value is incorrectly listed as `false` in the docs.


jmooring:
Thank you.

PR #2373, opened 2023-12-23
fix: hugo new content/posts/my-first-post.md instaed of content posts…

abdelmounim-dev:
…/...

fixed a typo at line 106


jmooring:
The "new content" sub-command was added in v0.112.0, so the existing documentation is correct. The quick start guide notes that v0.112.0 is a prerequisite.

PR #2377, opened 2023-12-28
Templates Lookup Order: Add missing parenthesis

arun-mani-j:
The paragraph under Output Format misses parenthesis around an example.

> We prefer matches with both (e.g. index.amp.html, but look for less specific templates.

This change adds the missing parenthesis.


jmooring:
Thank you.

PR #2392, opened 2024-01-09
Fix two links, one to bonsai.io and one to digital.gov

interrobot:
Fixed a couple links in the markdown content.


jmooring:
Thank you.

PR #2395, opened 2024-01-09
Fix inverted examples in Permalink and RelPermalink

huc-vu:
Stumbled upon this small typo in the documentation


jmooring:
Thank you.

PR #2397, opened 2024-01-11
Fix closing tag in notes docs

froboy:
On https://gohugo.io/contribute/documentation/#note the closing tag does not match the opening tag:

```
{{% note %}}
Use the [`math.Mod`] function to control...

[`math.Mod`]: /functions/math/mod/
{{% /code %}}
```

It looks like a copy/paste error that somehow also didn't break things. This resolves the issue.


jmooring:
Thank you.

PR #2399, opened 2024-01-14
Use TOML instead of YAML as YAML is not default anymore

jensilo:
As the title states, the getting started uses YAML for the `my-first-post.md` metadata. As TOML is default by now this should be changed. 

Might affect other parts of the documentation but here it's pretty obvious as the user creates the file and sees that his content is different from what the docs say his content would look like.


jmooring:
Thank you.

PR #2409, opened 2024-01-25
Create page describing how to include LaTeX/TeX in markdown

jmooring:


jmooring:
@j2kun

Please review this draft when you have a moment. I'm open to suggestions.
<https://deploy-preview-2409--gohugoio.netlify.app/content-management/mathematics/>

content/en/content-management/mathematics.md
```diff
@@ -0,0 +1,208 @@
...
+  docs:
+    parent: content-management
+    weight: 250
+weight: 250
+toc: true
+math: true
+---
+
+$$
+KL(\hat{y} || y) = \sum_{c=1}^{M}\hat{y}_c \log{\frac{\hat{y}_c}{y_c}}
```
j2kun, line 17:
```suggestion
KL(\hat{y} || y) = \sum_{c=1}^{M}\hat{y}_c \log{\frac{\hat{y}_c}{y_c}} \\
```

Adding `\\` to the end will render as a line break, which is I think what you want. Even better would be to use the following for a more precise alignment of the equals signs:

```
$$
\begin{aligned}
KL(\hat{y} || y) &= \sum_{c=1}^{M}\hat{y}_c \log{\frac{\hat{y}_c}{y_c}} \\ 
JS(\hat{y} || y) &= \frac{1}{2}(KL(y||\frac{y+\hat{y}}{2}) + KL(\hat{y}||\frac{y+\hat{y}}{2}))
\end{aligned}
$$
```

$$
\begin{aligned}
KL(\hat{y} || y) &= \sum_{c=1}^{M}\hat{y}_c \log{\frac{\hat{y}_c}{y_c}} \\ 
JS(\hat{y} || y) &= \frac{1}{2}(KL(y||\frac{y+\hat{y}}{2}) + KL(\hat{y}||\frac{y+\hat{y}}{2}))
\end{aligned}
$$

Ditto for the preformatted equation in lines 28-31

content/en/content-management/mathematics.md
```diff
@@ -0,0 +1,208 @@
...
+
+The configuration above allows you to delimit inline equations and expressions using a single `$` symbol. If you need to include a single `$` to represent currency anywhere on the page, you must double-escape the `$` symbol in your markdown. For example:
+
+```text
+The price is \\$67.42, but members get a 25% discount.
+```
+
+Note that the configuration above enables mathematical rendering on every page unless you set the "math" parameter to `false` in front matter. To enable mathematical rendering as needed, set the "math" parameter to `false` in your site configuration, and set the "math" parameter to `true` in front matter.
+
+To disable passthrough of inline snippets, omit the `inlne` key from the configuration:
```
j2kun, line 68:
```suggestion
To disable passthrough of inline snippets, omit the `inline` key from the configuration:
```

content/en/content-management/mathematics.md
```diff
@@ -0,0 +1,208 @@
...
+math = true
+{{< /code-toggle >}}
+
+The configuration above allows you to delimit inline equations and expressions using a single `$` symbol. If you need to include a single `$` to represent currency anywhere on the page, you must double-escape the `$` symbol in your markdown. For example:
+
+```text
+The price is \\$67.42, but members get a 25% discount.
+```
+
+Note that the configuration above enables mathematical rendering on every page unless you set the "math" parameter to `false` in front matter. To enable mathematical rendering as needed, set the "math" parameter to `false` in your site configuration, and set the "math" parameter to `true` in front matter.
```
j2kun, line 66: A forward reference to Step 3 might help clarify why there is a `math` parameter introduced here.

j2kun: It may also be worth noting that, even if `math` is set to false, you still should escape dollar signs if you configure them as delimiters, because otherwise the passthrough extension could prevent the rendering of other markdown syntax, as in the following example.

```
A $5 bill _saved_ is a $5 bill _earned_
```

j2kun (commented):
Looks great! Just a few small comments.

content/en/content-management/mathematics.md
jmooring: > otherwise the passthrough extension could prevent the rendering of other markdown syntax, as in the following example

Yeah, this is a common enough "gotcha" that the MathJax docs explain why $ isn't a default inline delimiter:
<https://docs.mathjax.org/en/latest/input/tex/delimiters.html#tex-and-latex-math-delimiters>

I think I'll do the same, removing the single $ delim from the examples, and create another section at the bottom of page describing how to enable, what to watch out for, etc.

jmooring:
Note that this page (/content-management/mathematics/) will not render correctly until Netlify builds with v0.122.0 or later.

PR #2413, opened 2024-01-27
do not merge: Just a test to avoid flickering in math

bep:
Just a test PR, close at will.

The flickering on page load is a little annoying, and I think conceptually this is the only (simple) way to fix this:

https://deploy-preview-2413--gohugoio.netlify.app/content-management/mathematics/

/cc @jmooring

### DIFF
--- a/_vendor/github.com/gohugoio/gohugoioTheme/layouts/_default/baseof.html
+++ b/_vendor/github.com/gohugoio/gohugoioTheme/layouts/_default/baseof.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html
   class="no-js"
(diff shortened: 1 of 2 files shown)


jmooring:
Thanks. I'll take care of this later today.

The FOUC with MathJax < v3 was pretty bad. With v3 it's on par with KaTeX, but neither are great. At some point, if/when there's enough demand, render hooks for inline and block would make stuff like this a bit cleaner (style the div/span instead of the body, set page Store instead of using page param, etc.).

bep:
Yea, also tempting to add a some options to allow some block/inline wrappers (e.g. `<div class="math"> etc.).

jmooring:
Thanks, handled in https://github.com/gohugoio/hugoDocs/pull/2415

PR #2433, opened 2024-02-09
Fix typo data-templates.md

savtrip:
The file name was called "user0123" in the example but the sample code accidentally utilised a capital "User0123" for the file name.


jmooring:
Thank you.

PR #2440, opened 2024-02-19
Add note on menus

nikolausschueler:
Make extra clear that the sectionPagesMenu setting only works for sections.

### DIFF
--- a/content/en/content-management/menus.md
+++ b/content/en/content-management/menus.md
@@ -34,7 +34,7 @@ Although you can use these methods in combination when defining a menu, the menu
 
 ## Define automatically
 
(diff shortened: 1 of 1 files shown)


jmooring:
I agree that a little clarification is needed here, but would prefer to simply change this:

```
To automatically define menu entries for each top-level section
```

To this:

```
To automatically define a menu entry for each top-level section of your site, 
```

If you feel that the word "section" should be linked to something, link to its glossary entry.

Please preview your changes locally before submitting a PR. For example, the markdown link in your prior submission is broken.

Finally, for future reference, we try to document what a thing does (something), not what it doesn't do (everything else).

Thanks.

jmooring:
Thank you.

PR #2446, opened 2024-02-21
command 'hugo server': minor improvements

deining:
This PR improves documentation of the 'hugo server' command: the two dashes in front of the `renderToMemory` flag showed up as one single `&ndash;` sign in the rendered page HTML code. This PR corrects this.


jmooring:
The CLI documentation is generated from code with the `hugo gen doc` command, so you'd have to submit a change to the main repository. However, since the text is displayed in the console as well, HTML entities are not desirable. So if you submit a change, find another way.

PR #2451, opened 2024-02-21
command 'hugo server': preserve two hypens for option 'renderToMemory'

deining:
This PR is a follow-up of rejected PR #2446. I found another way to preserve the two dashes and hope this approach can be accepted.


jmooring:
Sorry, no.

jmooring:
The right way to handle this is in the hugo source, using backticks.

```text
The ` + "`" + `hugo server` + "`" + ` command will by default write and serve files from disk, but
you can render to memory by using the ` + "`" + `--renderToMemory` + "`" + ` flag. This can be
faster in some cases, but it will consume more memory.
```

image

image

PR #2458, opened 2024-02-28
Update linux.md

uxai:
Updated to let users know that apt is not always up to date and causes issues in the getting started guide. It is advisable to use snap for installation instead.


jmooring:
Thanks, but this is sufficiently clear in the existing documentation:

image

And using the Snap package may not be the best choice:

image

PR #2468, opened 2024-03-07
Add the Hugo dedicated Quiqr CMS

mipmip:


jmooring:
Please place this entry in the list alphabetically.
Please hyphenate "open-source" and "cross-platform".
Please remove the word "hackable"... this is not a good thing (implies that it is insecure).

> so it can serve any requirement website version

I don't know what this means.

PR #2486, opened 2024-03-14
add pubcrank.com

pizzapanther:
I've built an editor for static site generators like Hugo and added it in the docs.

https://www.pubcrank.com/

### DIFF
--- a/content/en/tools/front-ends.md
+++ b/content/en/tools/front-ends.md
@@ -22,7 +22,7 @@ aliases: [/tools/frontends/]
 : DatoCMS is a fully customizable administrative area for your static websites. Use your favorite website generator, let your clients publish new content independently, and the host the site anywhere you like.
 
 [PubCrank](https://www.pubcrank.com/)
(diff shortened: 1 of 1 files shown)


content/en/tools/front-ends.md
```diff
@@ -21,6 +21,9 @@ aliases: [/tools/frontends/]
 [DatoCMS](https://www.datocms.com)
 : DatoCMS is a fully customizable administrative area for your static websites. Use your favorite website generator, let your clients publish new content independently, and the host the site anywhere you like.
 
+[PubCrank](https://www.pubcrank.com/)
+: PubCrank is a static site editor which lets you define templates for different front matter layouts in your site. This gives writers an easy to use visual interface to create and edit content while also maintaining the guardrails that the developer has setup. PubCrank is free for local editing.
```
jmooring, line 25:
```suggestion
: PubCrank is a static site editor which lets you define templates for different front matter layouts in your site. This gives writers an easy-to-use visual interface to create and edit content while maintaining the guardrails that the developer has created. PubCrank is free for local editing.
```

jmooring (changes requested):

PR #2488, opened 2024-03-16
Update versions in GitHub hosting example

etzngr:
This avoids the following warning when using the action workflow from the GitHub hosting example: `Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/upload-artifact@v3.`


jmooring:
Good call. Thank you.

PR #2502, opened 2024-03-25
Add `wp2hugo` migration tool

ashishb:


jmooring:
I'll merge this, but it may be removed when https://github.com/gohugoio/hugoDocs/issues/1573 is addressed. I suggest you add the "hugo-utility" topic to your repository, because at some point that's all we're going to point to (https://github.com/topics/hugo-utility).

ashishb:
Thanks. I added `hugo-utility` topic.

PR #2516, opened 2024-04-06
Fix a typo in templates docs

majaha:


jmooring:
You might as well fix the other one too.

majaha:
Good point, I missed that.

I also changed an ordering issue a few lines down, correct me if I'm wrong on that. Like it says in the docs, the piped value becomes the last argument.

PR #2518, opened 2024-04-07
add missing links to warnidf function

irkode:
add warnidf function to data/page_filters.yaml section functions_fmt_logging
add hint to use warnidf function for supression to warnidf.md


jmooring:
Thank you.

PR #2527, opened 2024-04-14
Update site.md: fix conflicting text

dogweather:
The current page gives conflicting info about whether to prefer `site` vs. `.Site`. This edit removes the ambiguity.

### DIFF
--- a/content/en/functions/global/site.md
+++ b/content/en/functions/global/site.md
@@ -19,7 +19,8 @@ At the top level of a template that receives the `Site` object in context, these
 {{ site.Params.foo }}
 ```
 
-When the `Site` object is not in context, use the global `site` function:
+When the `Site` object is not in context, only the global `site` function is guaranteed to return
+the current site:
 
 ```go-html-template
 {{ site.Params.foo }}


jmooring:
Thanks, but I think what we have is a bit clearer, because this isn't true:

> When the Site object is not in context, only the global site function is guaranteed to return the current site.

Using `$.Site` gets you the same thing as along as the Site object is available in the template context.

dogweather:
> as long as the Site object is available in the template context.

That's the problem, it's context dependent. [The current page](https://gohugo.io/functions/global/site/) gets it right at the bottom:

Screenshot 2024-04-14 at 00 39 37

But above that, this contradicts that advice:

Screenshot 2024-04-14 at 00 40 27

I.e., for the beginner and 99% of sites, and so the target audience of these docs, the key advice is: `site` is reliable. 

I.e., I have a massively multilingual site, but have no need for `$.Site`.

irkode:
I'll go with @dogweather for having the more general safer variant first

what about

image

<pre>
Use the `site` function to return the `Site` object regardless of current context.

```go-html-template
{{ site.Params.foo }}
```

When the `Site` object is in context you can use the `Site` property:

```go-html-template
<!-- current context -->
{{ .Site.Params.foo }}
<!-- template context -->
{{ $.Site.Params.foo }}
```

{{% note %}}
To simplify your templates, use the global `site` function regardless of whether the `Site` object is in context.
{{% /note %}}
</pre>

jmooring:
@irkode If you submit a PR I'll merge it.

irkode:
done - https://github.com/gohugoio/hugoDocs/pull/2528

PR #2528, opened 2024-04-14
fix conflicting text for global site function

irkode:
updated page for clarifying the use of the global site function as suggested by @dogweather. in https://github.com/gohugoio/hugoDocs/pull/2527


jmooring:
Thanks.

PR #2538, opened 2024-04-20
Fix netlify.toml examples in hosting-on-netlify

friek:
The build.environment line was missing in both configuration examples.
See also [Hugo on Netlify](https://docs.netlify.com/frameworks/hugo/)


jmooring:
Thank you.

PR #2550, opened 2024-05-01
Add Discourse to the comment system list

xuhdev:
Discourse has builtin support for being used as a comment system


jmooring:
I'll merge this, but it's a bit wonky:
>One important thing to note with this setup is that users have to navigate to your forum to post replies.

PR #2560, opened 2024-05-06
Document content adapters

jmooring:
TODO

- [x] Add a "Multilingual" section to the bottom of the page showing the content structure: one example when using one content directory, another example when using multiple content directories
- [x] Revisit "content.type" (now content.mediaType)
- [x] Revisit "markup" front matter field (identifiers)
- [x] Revisit content format classification table (https://gohugo.io/content-management/formats/#classification)
- [x] Describe difference between resource (use existing resource) and string (create new resource) in map passed to AddResource


content/en/content-management/content-adapters.md
```diff
@@ -0,0 +1,215 @@
...
+---
+
+{{< new-in 0.126.0 >}}
+
+## Overview
+
+A content adapter is a template that dynamically creates pages when building a site. For example, use a content adapter to create pages from a remote data source such as JSON, TOML, YAML, or XML.
+
+Unlike templates that reside in the layouts directory, content adapters reside in the content directory, no more than one per section. When a content adapter creates a page, the page's [logical path] will be relative to the content adapter.
+
```
bep, line 21: Re "one per section":

* I guess the correct would be "one per directory" -- there's nothing on the technical side of it that prevents `books/fiction/_content.gotmpl` and `books/documentary/_content.gotmpl`
* To complicate it further, you could also have one per language.

But I suspect it's best to keep the documentation a little simple in the first iteration.

jmooring: Changed to "one per directory". I'll add a "Multilingual" section to the bottom of this page, at some point, showing the content structure (one example when using one content directory, another example when using multiple content directories).

PR #2562, opened 2024-05-09
use-modules: clarify the example of initializing a new module

tanyabouman:
Every time I read this documentation to using Hugo Modules, I get confused about what initializing a new module is for, and run `hugo mod init <name of dependency>`, rather than `hugo mod init <name of my project>`.
I think this clarifies that a bit.


jmooring:
This is definitely an improvement. Thank you.

PR #2564, opened 2024-05-10
Clarify `languageCode` default in multilingual page

chalin:
According to my tests, `languageCode` defaults to the active language's language key.

### DIFF
--- a/content/en/content-management/multilingual.md
+++ b/content/en/content-management/multilingual.md
@@ -79,7 +79,7 @@ disabled
 : (`bool`) If `true`, Hugo will not render content for this language. Default is `false`.
 
 languageCode
-: (`string`) The language tag as described in [RFC 5646]. This value does not affect localization or URLs. Hugo uses this value to populate the `language` element in the [built-in RSS template], and the `lang` attribute of the `html` element in the [built-in alias template]. Examples:
+: (`string`) The language tag as described in [RFC 5646]. This value does not affect localization or URLs. Hugo uses this value to populate the `language` element in the [built-in RSS template], and the `lang` attribute of the `html` element in the [built-in alias template]. Defaults to the language key when unspecified. Examples:
 
 - `en`
 - `en-GB`


jmooring:
> According to my tests, languageCode defaults to the active language's language key.

I'm not sure what you're referring to, but the [opengraph template](https://github.com/gohugoio/hugo/blob/b1bf0bff2c8ce1e2cace5010a81911a3dbf8e674/tpl/tplimpl/embedded/templates/opengraph.html#L15) does this:

```text
{{- with or .Params.locale site.Language.LanguageCode site.Language.Lang }}
```

So the template defaults to the language key, but .Language.LanguageCode has a default of "">

chalin:
I have a setup where `languageCode` isn't explicitly set but `.Site.Language.LanguageCode` yields the (active) language key value. I can put together a simple repro if you'd like.

jmooring:
You're right. It looks like we changed the behavior in v0.112.0 and didn't update the docs.

chalin:
We can also see this at work in https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/alias.html, which uses `site.Language.LanguageCode` and (at least on my sites), that has the same value as the active language key rather than being `""` when not explicitly set.

jmooring:
This is surprising...

```
languageCode = 'foo'

[languages.en]
# languageCode = 'en-US'

[languages.de]
languageCode = 'de-DE'
```

The en site returns "en" instead of "foo". Something's not right here.

jmooring:
@chalin Technically the `languageCode` config does not have a default value. Instead, the `PAGE.Language.LanguageCode` and `SITE.Language.LanguageCode` methods fall back to `PAGE.Language.Lang` and `SITE.Language.Lang` respectively.

So I think it would be better to update those method descriptions:

- <https://gohugo.io/methods/page/language/#methods>
- <https://gohugo.io/methods/site/language/#methods>

Also, if you wouldn't mind, please update the description of `languageCode` here:
<https://gohugo.io/getting-started/configuration/#languagecode>

We should find a way to indicate that this value is ignore if a language key is present (or something to that effect).

And they we should go back through the embedded templates and remove the fallback logic since Hugo handles that with v0.112.0 and later. But that's for another day in the main repo.

IUSR:
> This is surprising (sort of)...
> 
> ```
> languageCode = 'foo'
> 
> [languages.en]
> # languageCode = 'en-US'
> 
> [languages.de]
> languageCode = 'de-DE'
> ```
> 
> The en site returns "en" instead of "foo". Something doesn't seem right here.
> 
> The above is true for monolingual sites as well. As soon as you define a language, the `languageCode` in the root config is ignored.

Sorry to hijack the discussion but I just want to confirm the TOML table name for individual languages: is it `languages.en` or `Languages.en` with a capital "L"? I spent 30 minutes trying to find the right language code for the non-English language I want to support, as Hugo kept on complaining `config value "zh" for defaultContentLanguage does not match any language definition`; then I saw some tutorial for earlier version of Hugo using the capitalized "languages" name and gave it a try, and it finally worked. I'm using Hugo version 0.125.7 extended under Arch Linux.

```
❯ hugo version
hugo v0.125.7+extended linux/amd64 BuildDate=unknown
```

jmooring:
@IUSR Keys are converted to lowercase internally, but you should get in the habit of using the case as shown in the [documentation](https://gohugo.io/content-management/multilingual/#configure-languages). 

Please use the forum (https://discourse.gohugo.io/) for questions and troubleshooting.

PR #2569, opened 2024-05-10
Fix parameter list for collections.Dictionary

irkode:
must always be a key-value pair, so the current text is missleading.

moved the `...` behind the square brackets, so it stands for repeating the pair

### DIFF
--- a/content/en/functions/collections/Dictionary.md
+++ b/content/en/functions/collections/Dictionary.md
@@ -8,7 +8,7 @@ action:
   related:
     - functions/collections/Slice
   returnType: mapany
-  signatures: ['collections.Dictionary KEY VALUE [VALUE...]']
+  signatures: ['collections.Dictionary KEY VALUE [VALUE KEY]...']
 aliases: [/functions/dict]
 ---
 


content/en/functions/collections/Dictionary.md
```diff
@@ -8,7 +8,7 @@ action:
   related:
     - functions/collections/Slice
   returnType: mapany
-  signatures: ['collections.Dictionary KEY VALUE [VALUE...]']
+  signatures: ['collections.Dictionary KEY VALUE [VALUE KEY]...']
```
jmooring, line 11: That doesn't look right to me either.

How about `collections.Dictionary KEY VALUE [...]`

irkode: Guess both variants will be understood by Readers. And both solve the issue with repeating just the value.

Found no defined rules in contributing doc, so _right_ then depend on (Hugo Docs) convention. Reading the signatures I would say "Written, so it can be understood by anyone (also non tekkis)". 

your choice I believe.

Feel free to update the PR - or gimme your decision and I will do.

### TL;TR;

When I look at the collections Page the three dots always stand for "repeat the element before any time you want". So  for me the explicit grouping is more precise. What's the meaning here? `function COLLECTION KEY VALUE [...]`. Is it a `[...]` _repeat operator_ or just "checkout there's more. OK that's tekki.

Also seen there are some variants in the docs --- guess used to be more expressive --- which seem not to follow the conventions. Don't get we wrong here - I'm fine with that.

- `collections.KeyVals KEY VALUES... ⟼ types.KeyValues`
  same as `collections.KeyVals KEY VALUE...`
  -   _VALUES_ some special thing like _PAGES_ 
  - if not I would use VALUE... here. - plural already implies repeat

- `collections.Append ELEMENT [ELEMENT...] COLLECTION ⟼ any`
  same as `ELEMENT... COLLECTION`.
  - my tekki ego: Uh that's inconistent.
  - my human ego perfectly understands that (and would not even have recognized that)

There's always a balance ... and from my experience will never be perfect

jmooring:
The changes in https://github.com/gohugoio/hugoDocs/commit/fb76580f81615def6e979f629b5bee648e5fbf6c make the `slice`, `querify`, and `dict` signatures symmetrical.

PR #2574, opened 2024-05-13
Describe .Language.LanguageCode fallback value

jmooring:
Closes #2564


chalin:
Great, thanks @jmooring - you got to this more quickly than I could. 🙌🏻 ✨

PR #2580, opened 2024-05-15
Clarify that a numeric timezone must be "-07:00"

xuhdev:


xuhdev:
I'm guessing this is true based on https://github.com/gohugoio/hugo/issues/12495#issuecomment-2111509436?

I'm also thinking that removing "America/Los_Angeles" can reduce confusion...

jmooring:
This is incorrect. Those are examples of parsable values, not layout strings, and the offset can be anything. Again, I think there's a conceptual disconnect here. The layout string is described here:

<https://gohugo.io/functions/time/format/#layout-string>

xuhdev:
Just wanna make sure I understand it right: If the layout string contains a timezone offset, it must be "-7:00". Is this correct?

jmooring:
As described here...

<https://gohugo.io/methods/time/format/#layout-string>

...it can be any of these:

image

PR #2584, opened 2024-05-16
Emphasize impact when changing virus scanner settings

irkode:
referencing https://discourse.gohugo.io/t/content-adapter-examples-and-performance/49830/20 ff I'm pleased to provide a PR

So here's my suggestion that will 

* create a prominent statement that one has to think about that
* makes it clearer that it's an option but has impact

Especially with a varying audience it's important to set a point.  The unexperienced users just see the performance  gain and they'll just do what suggested. And keep in mind: you are the _persons of trust_ for your users.

I would skip the link to Microsoft. It's better to have an emphasized statement about impacts here and not in a link to the documentation (the normal user will not read because it's already described ;-)

* the text states clearly it has impacts (emphasized by two notes at applicable places)
* the link hast to be maintained
* other scanners have such documentation, too

### DIFF
--- a/content/en/troubleshooting/performance.md
+++ b/content/en/troubleshooting/performance.md
@@ -14,17 +14,31 @@ aliases: [/troubleshooting/build-performance/]
 
 ## Virus scanning
 
-Virus scanners are an essential component of system protection, but the performance impact can be severe for applications like Hugo that frequently read and write to disk. For example, with Microsoft Defender Antivirus, build times for some sites may increase by 400% or more.
+{{% note %}}
+**Be vigliant**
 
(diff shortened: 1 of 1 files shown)


jmooring:
Thank you for your suggestion.

Perhaps there's a language barrier here, but when I read this it's scary and implies that it's dangerous. I'm not going to merge this as-is, and will review and make suggestions when I have time.

In the future, when a maintainer suggests an alternate approach, please give it ample consider before doing what you want to do.

irkode:
I'm fine with your decision. To not merge and rethink / rewrite.

I promise I did an ample consideration (had to lookup that in a dictionary).  And explained my different opinion in the PR. 

Didn't want to be scaring. Maybe something in between 'safe' and 'dangerous'.

PR #2586, opened 2024-05-17
content adapters: Use --printPathWarnings to detect page collisions

jmooring:
Merge upon release of v0.127.0.

Related to https://github.com/gohugoio/hugo/pull/12512

### DIFF
--- a/content/en/content-management/content-adapters.md
+++ b/content/en/content-management/content-adapters.md
@@ -342,7 +342,9 @@ content/
     └── the-hunchback-of-notre-dame.md
 ```
 
-If the content adapter also creates books/the-hunchback-of-notre-dame, the content of the published page is indeterminate. You can not define the rendering order. Avoid this content structure.
+If the content adapter also creates books/the-hunchback-of-notre-dame, the content of the published page is indeterminate. You can not define the processing order.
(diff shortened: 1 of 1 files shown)


content/en/content-management/content-adapters.md
```diff
@@ -342,7 +342,9 @@ content/
     └── the-hunchback-of-notre-dame.md
 ```
 
-If the content adapter also creates books/the-hunchback-of-notre-dame, the content of the published page is indeterminate. You can not define the rendering order. Avoid this content structure.
+If the content adapter also creates books/the-hunchback-of-notre-dame, the content of the published page is indeterminate. You can not define the rendering order.
```
bep, line 345: I think it's better to use _processing order_ or something about this.

In general, Hugo's build is:

1. Process: Read files,  insert pages and resources into a set of prefix trees. We do this in parallel (last item wins), so the order is not entirely deterministic.
2. Assemble: Apply cascades, build taxonomies, calculate dates etc.
3. Render: We render each site and output format in serialized order, but the pages within e.g. a site is rendered in parallel. 

So, in some sense, the user can "define rendering order", but not the "processing order".

PR #2607, opened 2024-06-01
Clarify cardinalityThreshold

xuhdev:
- Clarify that `0` has a special meaning.
- Use 60% instead of 50% in the example, because 50% is symmetric: 50% is above the threshold and 50% is below the threshold. Using a different percentage to remove ambiguity from the example.
- Text polishing.


bep:
Looks good to me.

PR #2637, opened 2024-06-23
Document changes to template functions in v0.128.0

jmooring:
- New: css.TailwindCSS
- Deprecate/rename: resources.Babel => js.Babel
- Deprecate/rename: resources.PostCSS => css.PostCSS
- Deprecate/rename: resources.ToCSS => css.Sass

### DIFF
--- a/content/en/functions/css/TailwindCSS.md
+++ b/content/en/functions/css/TailwindCSS.md
@@ -1,0 +1,81 @@
+---
+title: css.TailwindCSS
+description: Processes the given resource with the Tailwind CSS CLI.
+categories: []
+keywords: []
+action:
+  aliases: []
+  related:
+    - functions/resources/Fingerprint
+    - functions/resources/Minify
+  returnType: resource.Resource
+  signatures: ['css.TailwindCSS [OPTIONS] RESOURCE']
+toc: true
+---
+
+{{< new-in 0.128.0 >}}
+
+<!-- TODO remove this admonition when feature is stable. -->
+
+{{% note %}}
+This is an experimental feature pending the release of TailwindCSS v4.0.
+
+The functionality, configuration requirements, and documentation are subject to change at any time and may be not compatible with prior releases.
+{{% /note %}}
+
+## Prerequisites
+
+To use this function you must install the Tailwind CSS CLI v4.0 or later. You may install the CLI as an npm package or as a standalone executable. See the [Tailwind CSS documentation] for details.
+
+[Tailwind CSS documentation]: https://tailwindcss.com/docs/installation
+
+{{% note %}}
+Use npm to install the CLI prior to the v4.0 release of Tailwind CSS.
+
+`npm install --save-dev tailwindcss@next @tailwindcss/cli@next`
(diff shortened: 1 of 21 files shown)


jmooring:
This is obviously just a stub at this point:
<https://deploy-preview-2637--gohugoio.netlify.app/functions/css/tailwindcss/>

The deprecation/renaming stuff is complete.

bep:
>Deprecate/rename: resources.ToCSS => css.SASS

Hmm... Maybe `Sass` would be more correct?

jmooring:
> Maybe Sass would be more correct?

Yes, it would be.

bep:
OK, I have 

1. Renamed SASS => Sass
2. Moved the aliases.

jmooring:
I'm assuming the options map will look something like this:
<https://deploy-preview-2637--gohugoio.netlify.app/functions/css/tailwindcss/#options>

bep:
Tailwind options:

```go
type TailwindCSSOptions struct {
	Minify        bool // Optimize and minify the output
	Optimize      bool //  Optimize the output without minifying
	

// Service `mapstructure:",squash"`
	// Enable inlining of @import statements.
	// Does so recursively, but currently once only per file;
	// that is, it's not possible to import the same file in
	// different scopes (root, media query...)
	// Note that this import routine does not care about the CSS spec,
	// so you can have @import anywhere in the file.
	InlineImports bool

	// When InlineImports is enabled, we fail the build if an import cannot be resolved.
	// You can enable this to allow the build to continue and leave the import statement in place.
	// Note that the inline importer does not process url location or imports with media queries,
	// so those will be left as-is even without enabling this option.
	SkipInlineImportsNotFound bool
}
```

content/en/functions/css/TailwindCSS.md
```diff
@@ -0,0 +1,62 @@
...
+targetPath
+: (`string`) If not set, the transformed resource's target path will be the original path of the asset file.
+
+## Example
+
+```go-html-template
+{{ with resources.Get "css/main.css" }}
+  {{ $opts := dict "targetPath" "css/style.css" "minify" true }}
+  {{ with . | css.TailwindCSS $opts }}
+    {{ if eq hugo.Environment "development" }}
```
bep, line 53: ... hugo.IsDevelopment ...

content/en/functions/css/TailwindCSS.md
```diff
@@ -0,0 +1,81 @@
...
+target = 'css'
+{{< /code-toggle >}}
+
+Process the resource:
+
+```go-html-template
+{{ with resources.Get "css/main.css" }}
+  {{ $opts := dict "minify" true }}
+  {{ with . | css.TailwindCSS $opts }}
+    {{ if eq hugo.IsDevelopment }}
```
bep, line 70: remove eq

jmooring: Thanks for the catch.

bep:
Merged in 012162e0d2ede071ad2b9637db5fb1dbd5995d52

PR #2655, opened 2024-07-22
chore: add more information on Summary method

hugo-sid:


jmooring:
The summary is wrapped within a `p` element if (a) the summary was defined manually, and (b) the summary is a single line.

See <https://gohugo.io/content-management/summaries/#comparison>.

PR #2659, opened 2024-07-28
Removed fireship.io as they moved to svelte

CommanderStorm:
Video: https://www.youtube.com/watch?v=rFP7rUYtOOg

Previous PR https://github.com/gohugoio/hugoDocs/pull/2114 is not valid anymore:
image


bep:
Handled by 392afc8f93a7135d3604ef9e88a4932db1ceabd1

CommanderStorm:
I think you forgot about the link in the navbar 😉

Also for the future people who will need to look at the showcase it might be better to accept this PR. This way, they won't have to waste time by looking into fireship.io..

PR #2661, opened 2024-07-31
Add output format information to lookup order section

mrled:
Hugo supports different shortcodes per output format. This PR updates the docs to describe this.

I think this has been the case ever since https://github.com/gohugoio/hugo/issues/3220 was fixed in https://github.com/dmerejkowsky/hugo/commit/af72db806f2c1c0bf1dfe5832275c41eeba89906

### DIFF
--- a/content/en/templates/shortcode.md
+++ b/content/en/templates/shortcode.md
@@ -46,7 +46,7 @@ Shortcode templates have a simple [lookup order], which first looks for a shortc
 3. `/layouts/shortcodes/<SHORTCODE>.html`
 4. `/themes/<THEME>/layouts/shortcodes/<SHORTCODE>.html`
 
-For example, the built-in RSS output format uses a baseName of `index` and suffix of `xml` by default. When calling a shortcut via `{{< example >}}`, the RSS template will first look for a file `example.index.xml` and if one doesn't exist will fall back to `example.html`.
+For example, the built-in RSS output format uses a baseName of `index` and suffix of `xml` by default. When calling a shortcut via `{{</* example */>}}`, the RSS template will first look for a file `example.index.xml` and if one doesn't exist will fall back to `example.html`.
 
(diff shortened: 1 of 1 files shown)


content/en/templates/shortcode.md
```diff
@@ -39,10 +39,14 @@ Note the forward slash.
...
+Shortcode templates have a simple [lookup order], which first looks for a shortcode unique to the [output format] and then falls back to a generic HTML version.
 
-1. `/layouts/shortcodes/<SHORTCODE>.html`
-2. `/themes/<THEME>/layouts/shortcodes/<SHORTCODE>.html`
+1. `/layouts/shortcodes/<SHORTCODE>.<OUTPUT-BASENAME>.<OUTPUT-SUFFIX>`
+2. `/themes/<THEME>/layouts/shortcodes/<OUTPUT-BASENAME>.<OUTPUT-SUFFIX>`
+3. `/layouts/shortcodes/<SHORTCODE>.html`
+4. `/themes/<THEME>/layouts/shortcodes/<SHORTCODE>.html`
+
+For example, the built-in RSS output format uses a baseName of `index` and suffix of `xml` by default. When calling a shortcut via `{{</* example */>}}`, the RSS template will first look for a file `example.index.xml` and if one doesn't exist will fall back to `example.html`.
```
bep, line 49: I'm pretty sure the above isn't correct: We use the output `name`,  eg. `rss`, so `foo.rss.xml` should work.

mrled: Ahhh, good catch. I tested this, but failed to remember that my config is custom in this area, specifically I change the RSS output basename to "rss", which confused me. Fixed.

jmooring:
@mrled Thank you for bringing this to our attention. The entire section (lookup order) needed a rewrite, so this PR is superseded by #2678.

PR #2662, opened 2024-08-05
Improve embedded.md: fix filename patterns for OG and Twitter images.

mazgi:
I've read the following implementations.
`opengraph.html` and `twitter_cards.html` are both using `get-page-images.html`. So, the descriptions for the two embedded templates should be the same.

https://github.com/gohugoio/hugo/blob/2fcc53780f765c05dd6b7ec41c6bb76308362e5e/tpl/tplimpl/embedded/templates/opengraph.html#L38-L42

image

https://github.com/gohugoio/hugo/blob/2fcc53780f765c05dd6b7ec41c6bb76308362e5e/tpl/tplimpl/embedded/templates/twitter_cards.html#L1-L5

image

https://github.com/gohugoio/hugo/blob/2fcc53780f765c05dd6b7ec41c6bb76308362e5e/tpl/tplimpl/embedded/templates/partials/_funcs/get-page-images.html#L5-L9

image

### DIFF
--- a/content/en/templates/embedded.md
+++ b/content/en/templates/embedded.md
@@ -203,7 +203,7 @@ description = "Text about this post"
 images = ["post-cover.png"]
 {{</ code-toggle >}}
 
-If `images` aren't specified in the page front-matter, then hugo searches for [image page resources](/content-management/image-processing/) with `*feature*`, `*cover*`, or `*thumbnail*` in their name.
+If [page bundles](/content-management/page-bundles/) are used and the `images` array is empty or undefined, images with file names matching `*feature*`, `*cover*`, or `*thumbnail*` are used for image metadata.
 If no image resources with those names are found, the images defined in the [site config](/getting-started/configuration/) are used instead.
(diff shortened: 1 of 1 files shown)


content/en/templates/embedded.md
```diff
@@ -203,7 +203,7 @@ description = "Text about this post"
 images = ["post-cover.png"]
 {{</ code-toggle >}}
 
-If `images` aren't specified in the page front-matter, then hugo searches for [image page resources](/content-management/image-processing/) with `feature`, `cover`, or `thumbnail` in their name.
+If `images` aren't specified in the page front-matter, then hugo searches for [image page resources](/content-management/image-processing/) with `*feature*`, `*cover*`, or `*thumbnail*` in their name.
```
jmooring, line 206:
```suggestion
If [page bundles](/content-management/page-bundles/) are used and the `images` array is empty or undefined, images with file names matching `*feature*`, `*cover*`, or `*thumbnail*` are used for image metadata.
```

mazgi:
@jmooring thanks for your suggestion. i've applied it!

jmooring:
Thank you.

PR #2669, opened 2024-08-13
Document Page.Markup

jmooring:
See:
- <https://github.com/gohugoio/hugo/pull/12759>
- <https://github.com/gohugoio/hugo/issues/12761>

### DIFF
--- a/content/en/methods/page/Markup.md
+++ b/content/en/methods/page/Markup.md
@@ -1,0 +1,72 @@
+---
+title: Markup
+description: # TODO
+categories: []
+keywords: []
+params:
+  functions_and_methods:
+    returnType: # TODO
+    signatures: [PAGE.Markup]
+---
+
+<!-- TODO -->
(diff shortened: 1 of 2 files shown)


content/en/methods/page/Markup.md
```diff
@@ -0,0 +1,71 @@
...
+keywords: []
+action:
+  related: []
+  returnType: # TODO
+  signatures: [PAGE.Markup]
+---
+
+<!-- TODO -->
+
+<!-- JMM It is unclear to me if things like .Page.WordCount will be deprecated in favor of .Page.Markup.CountWords. Three levels (Page, Markup, Countwords) is a bit of a pain to document with the current docs theme. Several of the Markup methods required their own page.
```
bep, line 14: I will probably make the method names the same (e.g. WordCount and not CountWords) to make it easier for people to migrate (that change was a little bit of a vanity change), but I don't think we want to have the method live in both places forever.

PR #2672, opened 2024-08-15
testing: New Markup/Content section

bep:


bep:
/cc @jmooring so, I notice that our current theme leaves a lot to be desired in the nested sections area (and is long overdue for an overhaul), but I think my quick change above should be relatively simple/safe.

I think a new `Markup` (or similar object)  needs its own section:

https://deploy-preview-2672--gohugoio.netlify.app/methods/page/markup/

PR #2673, opened 2024-08-16
Chomp whitespace in example templates for `transform.ToMath`

ldeso:
Chomp whitespace in the example templates for `transform.ToMath` to avoid generating code blocks and/or unwanted newlines (see https://github.com/gohugoio/hugo/issues/12736#issuecomment-2293340971).

### DIFF
--- a/content/en/functions/transform/ToMath.md
+++ b/content/en/functions/transform/ToMath.md
@@ -101,7 +101,7 @@ There are 3 ways to handle errors from KaTeX:
   {{ with .Err }}
     {{ errorf "Failed to render KaTeX: %q. See %s" . $.Position }}
   {{ else }}
-    {{ . }}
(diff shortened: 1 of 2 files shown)


jmooring:
The problem you reported [here](https://github.com/gohugoio/hugo/issues/12736#issuecomment-2293340971) is related to the code formatting in your shortcode, and is triggered when the shortcode is called using the `{{% %}}` notation, which requires the following in your site configuration in order for it to work:

```text
[markup.goldmark.renderer]
unsafe = true
```

The code formatting of your shortcode is irrelevant when the shortcode is called using the `{{< >}}` notation.

Both examples in the documentation are correct; neither one is a shortcode example.

ldeso:
Thanks a lot for the detailed explanation.

PR #2674, opened 2024-08-19
Update quick-start.md

seb49:
syntax error cause command hugo server fail


jmooring:
The existing documentation is correct. If you are having problems please open a new topic on our forum:
<https://discourse.gohugo.io/>

Thanks.

PR #2684, opened 2024-08-25
docs: Update migrations.md

ashishb:
Move actively maintained `wp2hugo` above the other unmaintained migration tools


jmooring:
Sorting these by last maintenance date (descending) isn't something we're going to do. These should be sorted alphabetically; feel free to submit a PR.

As you and I have [discussed before](https://github.com/gohugoio/hugoDocs/pull/2502#issuecomment-2017945727), the entire "Tools" section is likely to be removed at some point.

PR #2694, opened 2024-09-05
fix name of config file in quickstart guide

matthias567:
In the quickstart guide, the name of the theme is written to hugo.toml, which does not work. Instead, it should be written into config.toml


jmooring:
The existing documentation  is correct, including the minimum version requirement as shown:


image

PR #2749, opened 2024-10-29
Move the [build] documentation to its own page

bep:
Mostly to make it easier to find (via search etc.).


jmooring:
Related to <https://github.com/gohugoio/hugoDocs/issues/1888>, which is a lot of work.

PR #2769, opened 2024-11-18
docs: Make section def more useful

salim-b:
Note that this section definition clearly overlaps with the glossary entry for ["section"](https://gohugo.io/getting-started/glossary/#section). If we would convert the glossary terms to structured data, we could actually re-use them at various places. What do you think @jmooring? I can open a separate issue if you agree.

### DIFF
--- a/content/en/methods/page/_common/definition-of-section.md
+++ b/content/en/methods/page/_common/definition-of-section.md
@@ -2,4 +2,4 @@
 # Do not remove front matter.
 ---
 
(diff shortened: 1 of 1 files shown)


salim-b:
> Let's leave it as-is and copy/paste when necessary, which isn't very often.

In that case I'd suggest to remove this "definition-of-section" snippet from most places (e.g. like [here](https://gohugo.io/methods/page/isdescendant/)) and instead link to the glossary term where applicable. It's confusing.

PR #2771, opened 2024-11-19
Add hugo.Store, site.Store and Shortcode.Store

bep:


jmooring:
I've got a branch that's almost finished that covers this and some other updates. I should have created a draft PR so that you would have not duplicated effort. Sorry about that.

PR #2775, opened 2024-11-20
docs(transform.Unmarshal): match lang attribute to title language in examples

kwentine:
Thanks for great documentation and software.

Unless I misunderstood the examples, I reckoned  `fr` and `en` attribute values should be swapped to match the language of the book title.


jmooring:
Thank you.

PR #2781, opened 2024-11-29
Update hosting-on-aws-amplify.md

dallmo:
in the section `Hosting`, added instructions on how to modify `amplify.yml` - because i just tried myself and Amplify does failed to detect, and i have to manually adjust the build settings like these proposed content changes.


jmooring:
> add a file named `amplify.yml` to override the build settings in Amplify ;

Below this sentence you should show the content of a working file, not the default. Users want to cut/paste, and the default file is a dead-end.

dallmo:
sorry i missed that "/public" in the default file image attachment, which is missing in my AWS Amplify config so i thought my addition in line 47 can make a difference.

if you consider this proposed change is redundant, i will close this request to avoid confusions.

PR #2801, opened 2024-12-27
Provide an example of how to update the base URL for GitLab

alecthegeek:
Add an example CI/CD script fragment to configure the GitLab pages URL

### DIFF
--- a/content/en/hosting-and-deployment/hosting-on-gitlab.md
+++ b/content/en/hosting-and-deployment/hosting-on-gitlab.md
@@ -21,13 +21,11 @@ aliases: [/tutorials/hosting-on-gitlab/]
 
 The `baseURL` in your [site configuration](/getting-started/configuration/) must reflect the full URL of your GitLab pages repository if you are using the default GitLab Pages URL (e.g., `https://<YourUsername>.gitlab.io/<your-hugo-site>/`) and not a custom domain.
 
-A consistent way to acheive this is by executing the following script fragment in your pages deployment job (assumes you are using a TOML based config file)
+A consistent way to acheive this is by using the --baseURL option to the Hugo generate command.
 
 ```sh
-script:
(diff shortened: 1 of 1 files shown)


jmooring:
Why can't we just pass `CI_PAGES_URL` to the `--baseURL` command line flag?

alecthegeek:
> Why can't we just pass `CI_PAGES_URL` to the `--baseURL` command line flag?

And this is the difference between the merely competent amateur and the very export professional. 😄 

Change done. Now I need to go update my projects...

content/en/hosting-and-deployment/hosting-on-gitlab.md
```diff
@@ -21,13 +21,11 @@ aliases: [/tutorials/hosting-on-gitlab/]
 
 The `baseURL` in your [site configuration](/getting-started/configuration/) must reflect the full URL of your GitLab pages repository if you are using the default GitLab Pages URL (e.g., `https://<YourUsername>.gitlab.io/<your-hugo-site>/`) and not a custom domain.
 
-A consistent way to acheive this is by executing the following script fragment in your pages deployment job (assumes you are using a TOML based config file)
+A consistent way to acheive this is by using the --baseURL option to the Hugo generate command.
```
deining, line 24: Typo: `acheive` -> `achieve`

deining (changes requested):

jmooring:
Yeah, let's hold off on this. Some more testing is required, but we probably just want to change the build command in the example. No need to add a paragraph describing this.

jmooring:
@alecthegeek 

Thanks for letting us know about the GitLab environment variable. I've update the example workflow file to pass this variable to the build command; there is no need to document its use (i.e., it's self-documenting). I did not remove the following:

image

Why? Because it's important to locally test (i.e., `hugo server`) how your site is served when server from a subdirectory.

PR #2873, opened 2025-02-07
Fix: Typo

njchristoffersen:
Change "English" to "German" when referring to the German version of the site.


jmooring:
Thank you.

PR #2896, opened 2025-02-13
Add Prettier and config

bep:
One thing I like about Go that it's no discussing about formatting.

But when multiple people contribute to the same Go templates, there will eventually be conflicts coming from different formatting.

`prettier` has its issues, but I don't think there are any better option for Go/Hugo templates, but I open this PR so we can discuss and come to an agreement.


jmooring:
I'm fine with this.

PR #2909, opened 2025-02-14
Add warning for AVIF image format

djibe:
Nowhere in the doc the absence of AVIF processing is stated.


jmooring:
In general, documentation should state what something does, not what it doesn't do.

Add a bulleted list of raster image formats supported by Hugo's imaging pipeline. Place it beneath this:

image

djibe:
As AVIF is now a common web file format, I thought it was supported.

I lost 40 minutes struggling and Hugo's search engine did not help. I think stating this case is important.

I read in <https://gohugo.io/functions/images/config/> that "Supported image formats include GIF, JPEG, PNG, TIFF, and WebP."
Should I add there a message to add an AVIF keyword ?

PR #2914, opened 2025-02-16
theme: Add anchor links via JS

bep:
This needs Hugo v0.144.0 to be released before merged.


jmooring:
I added a `dl` to `content/en/about/features.md`, then clicked the link beneath it, then went back to original page, etc. The number of link icons increases by one each time.

[Screencast from 2025-02-16 18-23-48.webm](https://github.com/user-attachments/assets/b29b9b6f-f4b1-4e64-b678-e2ebced026db)

bep:
>The number of link icons increases by one each time.

The more the merrier. Thanks, this is the Turbo cache that plays me a trick. I'll have a look at it.

bep:
I have pushed a fix for the "double anchors".

PR #2925, opened 2025-02-17
fix: add conflict warning on relative url

yunhao-jiang:
Change:
- Added a warning message that `relativeURLs` setting will conflict with `baseURL`

Reason:
- When `relativeURLs` and `baseURL` is set at the same time, `relativeURLs` will not go in effect and `baseURL` will still be used.
- Discovered this issue when trying to deploy hugo in Cloudflare Pages where relative URL not showing when using custom domain (i.e. Even with `relativeURLs` set to `true`, if there is a `baseURL` config, when I open my custom domain (CNAME to the Cloudflare Pages), the links are still showing the Cloudflare pages one.)

### DIFF
--- a/content/en/content-management/urls.md
+++ b/content/en/content-management/urls.md
@@ -397,7 +397,8 @@ canonifyURLs = true
 #### Relative URLs
 
 {{< note >}}
-Do not enable this option unless you are creating a serverless site, navigable via the file system.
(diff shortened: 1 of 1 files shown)


jmooring:
> When relativeURLs and baseURL is set at the same time, relativeURLs will not go in effect

That isn't true.

image

Please create a new topic on the [forum](https://discourse.gohugo.io/) if you need assistance.

PR #2927, opened 2025-02-18
Rename .prettierrc to .prettierrc.json

coliff:
This is recommended by Prettier itself and it adds syntax highlighting and linting to Prettier by adding the `.json` file extension.


jmooring:
> This is recommended by Prettier itself

Can you provide a reference for this? All I can find is that `.prettierrc` has a higher precedence than `.prettierrc.whatever`.

<https://prettier.io/docs/configuration>

coliff:
ah, sorry then! I thought it recommended it there in the docs. Perhaps they changed things around. Still, I think it's worthwhile for the syntax highlighting and linting but I'm fine if you want to close it for any reason...

jmooring:
> syntax highlighting and linting

This is what I'm seeing, lexed and linted.

image

coliff:
I can't quite understand what's going on there... When I format that file with Prettier it makes no changes. There shouldn't be a comma after `"singleQuote": true`...

jmooring:
> shouldn't be a comma after

Yeah, I put them there to show linting in effect. I think we can close this.

coliff:
no problem :-)

PR #2931, opened 2025-02-18
theme: Autoformat layouts and assets

bep:
- **theme: Format layouts**
- **theme: Format assets**


bep:
I have tried to be careful and manually inspected the diff, I think this should be safe/good. It would be good going forward if we could not have to worry too much about formatting.

bep:
>Instead of this?

I tried that, but that doesn't work with Prettier (it ERRORs).

PR #2942, opened 2025-02-21
Mention `_build` alias

chalin:
- Context: https://discourse.gohugo.io/t/are-the-build-and-build-front-matter-keys-aliases-or-distinct-in-their-functionality/53627
- Adds a footnote mentioning that `_build` is a recognized alias for `build`. Putting it in a footnote avoids calling too much attention to the alias, but at least documents its existence.

### DIFF
--- a/content/en/content-management/build-options.md
+++ b/content/en/content-management/build-options.md
@@ -12,7 +12,7 @@ toc: true
 aliases: [/content/build-options/]
 ---
 
-Build options are stored in a reserved front matter object named `build` with these defaults:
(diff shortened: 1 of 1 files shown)


jmooring:
@bep Can we deprecate the setting instead?

chalin:
+1 to deprecating, but it might be worth keeping the note since `_build` is out in the wild. WDYT?

jmooring:
@chalin https://github.com/gohugoio/hugo/pull/13441

Issue #2936, opened 2025-02-19
theme: Swap side bar order


jmooring:
In the right side bar we display a list of pages in the current section ("In this section") and the table of contents ("On this page"). I'm thinking it would be better if the TOC were first. My eyes naturally go the top right to look for the TOC (perhaps that's learned behavior from the previous theme). I'm not sure what other well-known, high-traffic sites do.

bep:
The ToC is sticky, so that wouldn't work, me think.

bep:
We could possibly make the entire right side sticky, though ...

jmooring:
I'm wavering on this. Sometimes it's more useful to have "in this section" op top. Closing for now.

irkode:
just to mention for consideration sometime later

> Sometimes

imho In general reading an article you want to be guided within that. references  to additional resources come twice.

maybe that sometimes cases are due to overall site organization ...

jmooring:
Re-opening. I had it right the first time.

PR #2945, opened 2025-02-26
theme: Swap sidebar order

bep:
A quick take on this from me.

So, I agree in the situations where the toc is relatively short, but I don't think it's a great idea when the toc covers most of the screen.

See #2936

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -60,14 +60,14 @@
 {{ end }}
 
 {{ define "rightsidebar_content" }}
-  {{/* in-this-section.html depends on these being reneredc first. */}}
-  {{ $related := partial "layouts/related.html" . }}
-  {{ $toc := partial "layouts/toc.html" . }}
-  {{ if not .Params.hide_in_this_section }}
-    {{ partial "layouts/in-this-section.html" . }}
-  {{ end }}
-  {{ $related }}
-  {{ if $.Store.Get "hasToc" }}
-    {{ $toc }}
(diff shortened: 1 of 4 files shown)


jmooring:
Thanks. I waffled on this one for the reason stated above, and my opinion may change again when we have a section menu. For now I vote to merge this.

jmooring:
@bep I'd like to revisit this. I asked Gemini and it had this to say:

> It's generally recommended that "On This Page" comes before "In This Section" in a webpage's sidebar.
> 
> Here's why:
> 
> - **Immediate Relevance**: "On This Page" provides a table of contents for the current page the user is viewing. This information is immediately relevant and helps the user quickly navigate within the content they are actively consuming.
> - **Hierarchical Understanding**: "In This Section" typically refers to other pages or sub-sections within a larger, broader category of content. While important for broader navigation, it's a step removed from the immediate content. Users usually want to understand the current page's structure before exploring related pages.
> - **User Flow**: Users often land on a specific page with a particular goal in mind. "On This Page" helps them achieve that goal more efficiently by pinpointing relevant sections. Once they've explored the current page, they might then look to "In This Section" for further related information.
> 
> Think of it like reading a book: you'd look at the table of contents for the current chapter (On This Page) before looking at the overall table of contents for the entire book (In This Section).

The above aligns with my thoughts and opinions.

And when I say "I'd like to revisit this" what I actually mean is, "Would you please have another look at this PR and make the necessary changes?"

PR #2954, opened 2025-03-01
content: Add hosting instructions for Codeberg Pages

cujohub:


jmooring:
Is the last planned addition to the hosting and deployment section, or do you have others in mind? I think it's good that we're covering the major players, but I don't want to document the minor players.

cujohub:
This would be the last addition, yes. Both SourceHut and Codeberg (especially the latter) have a growing userbase, hence why I wanted to document them.

PR #2962, opened 2025-03-04
Make code header text selectable

willfaught:
It's annoying that I can't copy the file path `layouts/_default/_markup/render-heading.html` from the code header in https://gohugo.io/render-hooks/headings/.

### DIFF
--- a/layouts/shortcodes/code.html
+++ b/layouts/shortcodes/code.html
@@ -15,8 +15,7 @@
       {{- $codeLang = cond (eq $ext "html") "go-html-template" $ext }}
     {{- end }}
     <div
-      class="san-serif text-sm inline-block leading-none pl-2 py-3 bg-gray-300 dark:bg-slate-700 dark: w-full select-none
-">
+      class="san-serif text-sm inline-block leading-none pl-2 py-3 bg-gray-300 dark:bg-slate-700 dark: w-full">
       {{ . }}
     </div>
   {{- end -}}
--- a/layouts/shortcodes/code.html
+++ b/layouts/shortcodes/code.html
@@ -21,7 +21,7 @@
   {{- end -}}
 
 
-  <div class="" x-ref="code">
+  <div x-ref="code">
     {{ $inner := trim .Inner "\n" | safeHTML }}
     {{ if .Get "nocode" }}
       {{ $inner }}


jmooring:
Making the file name unselectable was intentional, implemented here: <https://github.com/gohugoio/hugoDocs/commit/b03b96082d05cdadd5259d5f91c4515dd47f3ef7>

Why? Because it was too easy to inadvertently select the code header while you're selecting the code, and you rarely need to copy the code header anyway. This UI choice is not unique to Hugo. GitHub does the same thing with header information when viewing files.

Yes, you say, but GitHub provides a copy button everywhere, so you don't have to manually select text.

And we've been down that path before too. Our user base has a broad range of expertise. To avoid foot guns and visual distraction, we only want to display a copy button for useful, non-trivial, context-independent examples that we _want_ users to copy/paste.

My views on this may evolve as we receive more feedback in the coming months, but I'm closing this for now.

willfaught:
>we only want to display a copy button for useful, non-trivial, context-independent examples that we want users to copy/paste

I wanted to copy the file name to create a file with that name, as the example was showing. If the doc is saying to create layouts/_default/_markup/render-heading.html with content foo, I should be able to copy layouts/_default/_markup/render-heading.html, open a terminal at my Hugo module, type `vim<space>`, paste layouts/_default/_markup/render-heading.html, press enter, then copy foo and paste it in the terminal. With the way it is now, I have to either cd into _markup manually across multiple commands and then type the file name manually (which is long), or write the entire path for the vim command manually (which is very long) (tab completion can help with either approach, but each tab is still a separate step).

>GitHub provides a copy button everywhere, so you don't have to manually select text.

There isn't a copy button for the file path.

>To avoid foot guns and visual distraction

I don't see how it's a footgun, since the header spans the entire code block. How can you accidentally select both rows instead of the much larger second row? Even if you somehow do, you just have to delete the first line of the pasted text.

>GitHub does the same thing with header information when viewing files.

I don't know what you mean. On https://github.com/willfaught/paige/blob/master/hugo.toml, all text is selectable except for button text like "Raw".

>And we've been down that path before too. Our user base has a broad range of expertise.

I do sympathize if there's a history of this causing a problem for people, somehow. In my humble opinion, the line should be drawn at expecting people to be able to copy text correctly, but then again, I'm usually not the one to help those people, so I defer to your judgment here.

Perhaps a middle ground is to ensure that code block header text is also in an adjacent paragraph so it can be copied if needed:

```diff
-To create a render hook that does the same thing:
+To create a render hook that does the same thing, create layouts/_default/_markup/render-heading.html with this content:
```

willfaught:
Sounds good.

PR #2976, opened 2025-03-07
feat: use cache when deploy on github pages

tiennm99:
I think use GitHub Action Cache to cache `resources` directory is a good practice when deploy on GitHub Pages. It helps build faster, especially when we have many resource like images or scss that need to be processed.


jmooring:
This is a great idea, but I'd to change the implementation. With what you propose, will we not cache remote resource calls or modules. This is the default configuration for our file cache:

<https://gohugo.io/configuration/caches/>

We should instruct users to override the image cache location by placing this in their site configuration:

```toml
  [caches.images]
    dir = ':cacheDir/_gen'
    maxAge = -1
```

Then we need to pass the `--cacheDir something` to the build command. Please test the setup before resubmitting the PR. Thanks!

jmooring:
We should have done this a long time ago.

tiennm99:
Thank you for your comment. I will improve it.

PR #2977, opened 2025-03-08
Misc edits 2025 03 06

jmooring:


jmooring:
One of the nice things about using a blockquote render hook instead of the "note" shortcode: you can use the "Source action > Organize link definitions" command without fear of breaking anything. With the "note" shortcode, link references must be placed between the opening and closing tags. If you ran the "Organize link definitions" it would pull the link references to the bottom of the document, silently leaving you with stuff like "a [short] link" in the content.

PR #2983, opened 2025-03-09
theme: Hide anchor until mouse-over

### LINKED ISSUE #2982: Hide heading and dt anchors until related element is moused-over
The fact that we can easily get a link+fragments to headings and `dt` elements is great:

Image

But as seen above, there's a lot going on visually.

Would it be possible to hide them until the user mouses over them? Here's an example:
<https://pkg.go.dev/text/template>

bep:
Fixes #2982


jmooring:
Bingo. That's perfect.

Issue #3004, opened 2025-03-17
debug.Timer.STop method is not documented


irkode:
No search result for that "Stop"

For a full reference all methods should be documented.

Issue #2996, opened 2025-03-13
content: Restructure using the Diátaxis approach


jmooring:
Top-level sections per the [Diátaxis](https://diataxis.fr) approach:

- Tutorials
- Concepts
- Guides
- Reference

Additional top-level sections that do not belong in the above:

- Contribute
- Glossary
- Installation
- Troubleshooting

Other issues to be addressed while the body's open:

- [ ] #389
- [x] #1289
- [x] #1573
- [ ] #2218
- [ ] #2307
- [x] #2308
- [x] #2355
- [x] #2448
- [x] #2461
- [ ] #2465
- [ ] #2494
- [ ] #2703
- [ ] #2747
- [x] #2850
- [ ] #3129
- [ ] #3139
- [ ] #3036
- [x] #3172
- [ ] #3200
- [x] Add Cloudflare hosting guide
- [x] Add Render hosting guide
- [x] Add Vercel hosting guide

See also:

- <https://discourse.gohugo.io/t/thoughts-about-hugos-documentation/55837>

<details>
<summary>Site structure and status</summary>

- [x] about/
  - [x] introduction
  - [x] features
  - [x] security
  - [x] license
- [ ] docs/
  - [x] installation/
      - [x] bsd
      - [x] linux
      - [x] macos
      - [x] windows
  - [ ] tutorials/
    - [ ] quick-start
    - [ ] tbd-1
    - [ ] tbd-2
    - [ ] tbd-3
    - [x] external-learning-resources
  - [ ] concepts/
    - [x] archetypes
    - [x] asset-pipelines
    - [ ] content-adapters
    - [x] content-formats
    - [x] content-summaries
    - [x] data-sources
    - [ ] localization
    - [x] markdown-attributes
    - [x] markdown-render-hooks
    - [ ] multidimensional content model
    - [ ] menus
    - [x] modules
    - [x] pagination
    - [x] partial decorators
    - [ ] project-structure
    - [x] shortcodes
    - [x] syntax-highlighting
    - [ ] taxonomies
    - [x] template-types
    - [x] templating
    - [x] themes
    - [x] unified-file-system
    - [x] url-management
  - [ ] guides/
    - [ ] asset-pipelines/
      - [ ] css
      - [ ] imaging
      - [ ] javascript
      - [ ] sass
      - [ ] tailwind
    - [ ] authors
    - [ ] build-options
    - [ ] comments
    - [ ] content-adapters
    - [x] custom-shortcodes
[...]

PR #3005, opened 2025-03-17
content: add documentation for Timer.Stop

irkode:
implements: #3004 

- Add `debug.Timer` Method page
- Add Documentation of  `Stop` method
- Add keyword "timer" to data/keywords.yaml
- Add "timer" keyword to debug.Timer and timer.Stop

### DIFF
--- a/content/en/functions/debug/Timer.md
+++ b/content/en/functions/debug/Timer.md
@@ -2,12 +2,12 @@
 title: debug.Timer
 description: Creates a named timer that reports elapsed time to the console.
 categories: []
-keywords: []
+keywords: ["timer"]
 params:
   functions_and_methods:
     aliases: []
     returnType: debug.Timer
-    signatures: [debug.Timer NAME] 
+    signatures: [debug.Timer NAME]
 ---
 
 {{< new-in 0.120.0 />}}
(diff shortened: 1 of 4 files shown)


content/en/functions/debug/Timer.md
```diff
@@ -35,3 +35,47 @@ The results are displayed in the console at the end of the build. You can have a
...
+
+will produce the following result.
+
+````text { hl_Lines=1 }
+INFO  timer:  name Inner count 2000 duration 3.1275511s average 1.563775ms median 0s
+INFO  timer:  name Outer count 1 duration 3.9826673s average 3.9826673s median 0s```
+````
+
+> [!Note]
+> You don't have to explicitly stop a timer. It is automatically stopped when a new timer is created with the same name.
```
bep, line 63: "It is automatically stopped when a new timer is created with the same name."

The above is not correct:

* You can have many running timers with the same name (that's one of the key points)
* A timer starts when you do `$t := debug.Timer "foo"`
* A timer stops on either `$t.Stop` or at the end of the build.

content/en/functions/debug/Timer.md
```diff
@@ -35,3 +35,47 @@ The results are displayed in the console at the end of the build. You can have a
...
+
+## Example - Unique names
+
+To have each timer reporting it's execution time use unique names for example with math.Counter
+
+```go-html-template { hl_Lines=2}
+   {{ range seq 3 }}
+      {{ $t1 := debug.Timer (printf "Run %d" math.Counter) }}
```
bep, line 71: I don't think this example adds much value, so I suggest we remove it. The primary use case for timers are aggregated timers, e.g. is the partial of name `mypartial`  slow?

bep: Also, I'm not totall convinced about `timer` needing to be its own section; it has one method, Stop, and there will  most likely never be more, but I'll leave that to @jmooring

jmooring: > I'm not totall convinced about timer needing to be its own section

Agreed.

content/en/functions/debug/Timer.md
irkode: Rechecked my first results and with your comments I guess managed to get it. confused of the "Just one line" output

So in the first case the T1 just measures the first partial, but without stop it continues processing til the end.

So the timers are still running but summed up to the one with the same name...

```go-html-template
   {{ $t1 := debug.Timer "ONE" }}
   {{ partial "sleep" 500 }}
   {{ $t1.Stop }}
   {{ $t2 := debug.Timer "ONE" }}
   {{ partial "sleep" 500 }}

INFO  timer:  name ONE count 2 duration 3.9068468s average 1.9534234s median 1.5616702s

   {{ $t1 := debug.Timer "ONE" }}
   {{ partial "sleep" 500 }}

   {{ $t2 := debug.Timer "ONE" }}
   {{ partial "sleep" 500 }}

INFO  timer:  name ONE count 2 duration 6.2918664s average 3.1459332s median 0s
```

content/en/functions/debug/Timer.md
irkode: Indeed one method for one object seems overkill. 

but that's how the docs are organized. Two sections:
* functions
  - for function calls
* methods 
  - method invocaction
  - field access

reasoning with the documentation layout might not be the best... The whole documentation works like that. To find the docs of a Thing returned by a function -> I check the Method section. (except I suppose it's a simple type... 

Documenting a method/function just by a remark on another page doesn't seem right to me.

maybe an option  would be to convert it to a `debug.StopTimer TIMER` function. Will solve the docs section problem and maybe make the interface even clearer to have thet in one section.

all about how users expect to use the documentation after they learned how it's organized ...

p.s. this issue will still be there when addressing #2996 for the reference part

all that said - Not feeling really strong about that, it's just a small flaw

irkode:
mmh, could't set that back to draft.

 closing for now - it at least would need an update regarding stop and name...

PR #2992, opened 2025-03-11
content: Miscellaneous updates related to v0.146.0

jmooring:
- [x] https://github.com/gohugoio/hugo/pull/13464
- [x] https://github.com/gohugoio/hugo/pull/13532
- [x] https://github.com/gohugoio/hugo/pull/13540
- [x] https://github.com/gohugoio/hugo/pull/13554
- [x] https://github.com/gohugoio/hugo/pull/13566

See:

- <https://deploy-preview-2992--gohugoio.netlify.app/shortcodes/vimeo/#loading>
- <https://deploy-preview-2992--gohugoio.netlify.app/functions/css/sass/#silencedependencydeprecations>
- <https://deploy-preview-2992--gohugoio.netlify.app/functions/time/in/>
- <https://deploy-preview-2992--gohugoio.netlify.app/functions/templates/current/>

PR #3020, opened 2025-04-07
content: Added correct links to source code of embedded templates

gurunandan-bhat:
In the page "Docs->Templates->Embedded Templates" many links to the sources of embedded templates have been corrected


jmooring:
This was already addressed in https://github.com/gohugoio/hugoDocs/pull/2992.

PR #3022, opened 2025-04-08
deps: TailwindCSS 4.1.3

bep:


bep:
Never mind, seems l misunderstood the ^version:

```
^version “Compatible with version”, will automatically update you to all future minor/patch versions that are backwards-compatible, without incrementing the major version. ^1.2.3 will use releases from 1.2.3 to < 2.0.0.
```

PR #3031, opened 2025-04-15
Make "Hugo" link static

ellnix:
First time using/contributing to Hugo, so I'm sorry if the etiquette or the code is off somehow. Willing to make changes.

---

Noticed that in the themes page clicking the top left Hugo link takes you to themes instead of the main docs homepage. It didn't quite make sense to me.

### DIFF
--- a/hugo.toml
+++ b/hugo.toml
@@ -142,6 +142,10 @@ category = 'categories'
 
 ######## GLOBAL ITEMS TO BE SHARED WITH THE HUGO SITES ########
 [menus]
+    [[menus.hugo]]
+        identifier = 'Hugo'
+        name       = 'HUGO'
+        pageRef    = '/'
     [[menus.global]]
         identifier = 'news'
         name       = 'News'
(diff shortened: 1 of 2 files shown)


bep:
@jmooring I'm not sure what the intention of PR creator was for this, but after I upgraded the themes site (https://themes.gohugo.io/) to the new theme, the Hugo link isn't correct on both site -- which I guess this PR could help fix (assuming we also update the menu config in the theme repo).

ellnix:
> All you need to do is change this:
>  ...

Judging by the documentation, that I'm seeing [here](https://gohugo.io/methods/site/home/#article), that change would only make a difference to whether or not the link is absolute or relative. Permalink is based on the `baseURL` which on the themes repo would continue to point to https://themes.gohugo.io/, whereas with this change we can make the themes page point to https://gohugo.io

I'm sorry in advance if I'm misunderstanding documentation.

> Your commit message should be something like:
> 
> ```
> theme: Use permalink to link to home page
> ```

Thanks, will keep in mind.

> @jmooring I'm not sure what the intention of PR creator was for this, 

I'll be clearer:
image

My expectation was that clicking "Hugo" while on the themes page would take me back to the documentation page (the home page essentially), but instead it takes me back to the root of the themes page.

Since the two pages are two separate projects, and the link is relative to the baseUrl, I thought that to get a link pointing to the same place in both I should add a new menu item which can be set to different values in different repos.

Apologies for the confusion, I should have been clearer from the beginning.

> (assuming we also update the menu config in the theme repo).

That was my intention, but I felt like making multiple PRs without getting some feedback would essentially be spam.

PR #3059, opened 2025-05-08
Add a one pager about the new template system

bep:
https://deploy-preview-3059--gohugoio.netlify.app/templates/new-templatesystem-overview/


jmooring:
This example doesn't make sense to me:
<term.mylayout.section.en.rss.xml>

There are other breaking changes that should be included, but it would be easier to handle those as a separate PR after merging this one.

bep:
><term.mylayout.section.en.rss.xml>

You're right about that ... I'm fixing that and merging this.

PR #3068, opened 2025-05-12
Fix Relref examples

MakoPhil:
Examples were incorrectly showing usage of `ref` instead of `relref`.

https://gohugo.io/shortcodes/relref/


jmooring:
Thank you.

PR #3071, opened 2025-05-19
Added GitCMS information in hugo docs

Waishnav:


bep:
I have tested this with a few Hugo repos, and pushing the "init button" does this:

image

But no other popup/activity.

Have anyone had any success with this vs Hugo?

Waishnav:
@bep I don't know what GitCMS config editor is not shown in your browser. But it should have. Maybe try refreshing current tab.

bep:
@Waishnav I have tried variations of what you describe + disable all other extensions. I see no errors in any console, but no popup when I hit the init button.

Waishnav:
> We've removed this from our documentation for now.

Yeah, No worries. Would love to solve that bug and get it working on your devices. let me know if you guys are available. maybe we can schedule the call

bep:
>We've removed this from our documentation for now.

Note that I think it's a very interesting extension (on paper, that is). But my thinking is that when I couldn't get it to work (I like to think I'm a good with computers and stuff), then it's a good chance others also will struggle.

Waishnav:
> Note that I think it's a very interesting extension (on paper, that is). But my thinking is that when I couldn't get it to work (I like to think I'm a good with computers and stuff), then it's a good chance others also will struggle.

Sorry for that experience. I'm on a Linux machine and have only tested it on Chrome so far. I'll test it across all Chromium-based browsers tonight and fix the on-boarding-related issues. I'll raise a new PR once everything's resolved.

bep:
>Sorry for that experience. I'm on a Linux machine and have only tested it on Chrome so far

I'm on a MacBook in Chrome Version 136.0.7103.114 (Official Build) (arm64).

Waishnav:
@bep Interesting — when I built this extension, I was using Chrome version 135. But now, testing on version 136.0.71 on Linux, the content script doesn't run on the initial render, which is causing this bug. What's even weirder is that if you resize the Chrome window, the GitCMS config editor re-renders.

PR #3076, opened 2025-05-21
Content: Add guide for hosting Hugo websites on Azion

egermano:
Introducing a new option for deploy Hugo projects on Azion Web Platform.


jmooring:
Thanks for this, but we've decided to pass on including it in our documentation. There are many ways to host and deploy a Hugo site, and we document those that are most commonly used by our community. 

You might consider sharing this information in the [forum](https://discourse.gohugo.io/) as a "tips & tricks" topic.

egermano:
Hi, thanks for the quick answer and also thanks for the suggestion, I'll do it.

PR #3078, opened 2025-05-22
Update the linux-amd64 link in index.md

chriseborowski:
Updated the linux-amd64 link in index.md. The current link, upon pasting in the browser, seems inactive and generates runtime failures.


jmooring:
No.

PR #3079, opened 2025-05-22
Add hosting support for ENS & IPFS

rrbin666:
As agreed with @bep


bep:
@jmooring OK, I will think about this a little more.

jmooring:
Closing for now.

PR #3092, opened 2025-05-31
content: Fix incorrect link to Shortcodes tutorial

Skimlk:
Change the Shortcodes tutorial link from

https://www.giraffeacademy.com/static-site-generators/hugo/archetypes/

to the correct URL:

https://www.giraffeacademy.com/static-site-generators/hugo/shortcodes/


jmooring:
Thank you.

PR #3094, opened 2025-06-01
content: Add instructions about deploying to root of user's Codeberg Pages

rimas-kudelis:
This article expands the chapter about automated deployment  to Codeberg Pages which I added in #3083 by adding another section which details how one can build sources from one repo and commit the result to another.

It's likely that the two sections can be merged, since the differences between them are quite minimal, but I don't want to withhold that information just because it isn't perfectly organized yet.


jmooring:
This is quite thorough. I will assume it's all correct, and will merge this, but I must admit that my eyes glazed over a bit while reading it. Compared to some of the more popular/common hosting options this seems like a lot of work.

PR #3095, opened 2025-06-02
theme: Remove the print only QR code header

bep:
I had a head scratching moment wondering where the qr codes directly below public/ came from.

I understand the motivation, but this produces a little too many files for this ... narrow feature.


jmooring:
Would you consider a compromise where we use the existing (cached) QR codes instead?

jmooring:
Yeah, I should have re-used the images when I implemented this. I'll take care of it.

PR #3109, opened 2025-06-17
search: Use front matter parameter to control search indexing

### LINKED ISSUE #3108: search: Glossary page does not appear in search results
This happens because the glossary has page kind "section" and our search config only looks for kind "page":

```text
selectorsToMatch: [".kind-page"],
```

We could hardcode something in the search config, but I'd rather do something extensible like using a front matter field:

site config:

```toml
[[cascade]]
  [cascade.params]
    searchable = true
  [cascade.target]
    kind = 'page'
[[cascade]]
  [cascade.params]
    searchable = false
  [cascade.target]
    kind = '{home,section,taxonomy,term}'
```

If `.Page.Params.searchable` is `true` we include `searchable` in the body element's `class` attribute.

And this way we can override the default behavior in front matter.

jmooring:
Closes #3108

### DIFF
--- a/content/en/contribute/documentation.md
+++ b/content/en/contribute/documentation.md
@@ -152,26 +152,27 @@ This example demonstrates the minimum required front matter fields.
 
 If quotation marks are required, prefer single quotes to double quotes when possible.
 
-Seq|Field|Description|Required
---:|:--|:--|:--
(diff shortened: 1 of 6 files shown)


djmartberry66:
Djsmartberry

PR #3140, opened 2025-07-18
added instructions on custom domain for Codeberg Pages and Forgejo Actions

devkamiki:
The document's sample `hugo.yaml` is only effective if the user set baseURL to `username.codeberg.page`. It's worth noticing that this config clears `.domains` file of the target branch, therefore, it would cause "misdirected request" if you use automated deployment following this Forgejo Actions config file. I added a small modification to make sure that `.domains` file will be present in the target branch when automated deploy is on following the instructions.
Thanks!


devkamiki:
Sorry, this actually doesn't fix the problem of .domains file. It's rooted from https://github.com/actions/upload-artifact/issues/602 and forgejo actions v3 being a mirror of github actions v3. I will initiate a new PR with correct statement and fixation by explicitly indicating dot files should be included

PR #3162, opened 2025-08-16
misc: Update spellcheck configs

jmooring:


content/en/about/dimensions.ai
```diff
@@ -0,0 +1,589 @@
+%PDF-1.5%����
```
bep, line 1: What is this?

PR #3184, opened 2025-08-25
content: Document the collections.D function

jmooring:
<https://deploy-preview-3184--gohugoio.netlify.app/functions/collections/d/>

Merge after release of v0.149.0.

### DIFF
--- a/content/en/functions/collections/D.md
+++ b/content/en/functions/collections/D.md
@@ -1,0 +1,53 @@
+---
+title: collections.D
+description: Returns a slice of sequentially ordered random integers.
+categories: []
+keywords: [random]
+params:
+  functions_and_methods:
+    returnType: '[]int'
+    signatures: [collections.D SEED N HIGH]
+---
+
+{{< new-in 0.149.0 />}}
+
+The `collections.D` function returns a slice of `N` sequentially ordered unique random integers in the half-open [interval](g) [0, `HIGH`) using the provided `SEED` value. This function implements J. S. Vitter's Method&nbsp;D[^1] for sequential random sampling, a fast and efficient algorithm for this task.
+
+See [this article][] for a detailed explanation.
+
+```go-html-template
+{{ collections.D 6 7 42 }} → [4, 9, 10, 20, 22, 24, 41]
+```
+
+The example above generates the _same_ random numbers each time it is called. To generate a _different_ set of 7 random numbers in the same range, change the seed value.
+
+```go-html-template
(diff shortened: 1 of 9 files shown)


content/en/functions/collections/D.md
```diff
@@ -0,0 +1,49 @@
+---
+title: collections.D
+description: Returns a slice of sequentially ordered random integers.
+categories: []
+keywords: [random]
+params:
+  functions_and_methods:
+    returnType: '[]int'
+    signatures: [collections.D SEED N MAX]
```
bep, line 9: I have renamed max => hi in the implementation; hi is commonly used as the upper boundary in Go slice notation, which should give a slightly better indication of the half-open interval just by looking at the name.

content/en/functions/collections/D.md
```diff
@@ -0,0 +1,49 @@
...
+
+> [!note]
+> All arguments are cast to integers, so setting the seed to `3.14` is the same as setting it to `3`.
+
+A common use case is the selection of random pages from a page collection. For example, to render an unordered list of 5 random pages using the current time as the seed value:
+
+```go-html-template
+<ul>
+  {{ $p := site.RegularPages }}
+  {{ range collections.D now.Unix 5 ($p | len) }}
```
bep, line 36: I don't think `now.Unix` is a great `seed` value used in an example, maybe `now.YearDay` ? In most cases you would want a stabler output.

content/en/functions/collections/D.md
```diff
@@ -0,0 +1,49 @@
...
+The example above generates the _same_ random numbers each time it is called. To generate a _different_ set of 7 random numbers in the same range, change the seed value.
+
+```go-html-template
+{{ collections.D 2 7 42 }} → [3, 11, 19, 25, 32, 33, 38]
+```
+
+> [!note]
+> All arguments are cast to integers, so setting the seed to `3.14` is the same as setting it to `3`.
+
+A common use case is the selection of random pages from a page collection. For example, to render an unordered list of 5 random pages using the current time as the seed value:
```
bep, line 31: It's an ordered list (which is a difference from shuffle, but a difference I suspect most people prefer).

content/en/functions/collections/D.md
jmooring: The signature is now `collections.D SEED N HIGH`.

content/en/functions/collections/D.md
jmooring: Changed seed to `now.YearDay`.

PR #3192, opened 2025-08-30
Fix templating introduction

kryzar:
Replace `partial` by `template`, which should have been used in the first place.


bep:
Using the `partial` func is the new recommended way even in this case. There's a longer story, but that needs to wait.

kryzar:
Alright, but then the explanation surrounding the event seems very confusing... Maybe one sentence stating your comment would be useful.

PR #3229, opened 2025-10-12
Update shortcode.md

jamescridland:
Adding clearer documentation on how to change shortcodes for RSS feed generation


jmooring:
In https://github.com/gohugoio/hugoDocs/pull/3257 we changed the shortcode lookup examples to include the `rss` output format. We intentionally omitted the admonition because this is already covered by the introductory sentence:

> Hugo selects shortcode templates based on the shortcode name, the current output format, and the current language.

PR #3252, opened 2025-10-31
Enhance mathematics.md with partial template guidance

xxraincandyxx:
Added instructions for using partial templates with external themes and the math parameter. Specially for the case where the user use external themes and doesn't have the file `baseof.html` under directory `layouts`, tested works for me.


jmooring:
Thanks for this, but `layout/_partials/extend_head.html` is a theme-specific template (i.e., some themes might have it, some may have something similar with a different path, some may not have this feature at all). 

An in-progress documentation update includes instructions for overriding theme templates:

image

PR #3302, opened 2025-12-10
Added Note from Discourse #23605

cmahnke:


jmooring:
The existing admonition is sufficient.

I've long advocated for the deprecation of `markdownify` in favor of `.Page.RenderString`, but have yet to garner much support.

jmooring:
> wouldn't it be easier to just make markdownify an alias for .Page.RenderString

Sure, but it needs to be in the context of the current... Page.

In v0.93.0 (specifically [#9544](https://github.com/gohugoio/hugo/pull/9544)) we solved some related problems by  aliasing `markdownify` to `site.Home.RenderString`:




https://github.com/gohugoio/hugo/blob/364d237b511e5d36b71796e2966549b34034915b/tpl/transform/transform.go#L184-L199

But this approach (setting context to the home page) causes problems that can be difficult to troubleshoot for most users. Some of those problems are described in the comments here:

<https://github.com/gohugoio/hugo/issues/9692>

PR #3341, opened 2026-01-11
Fix description of disableDefaultLanguageRedirect

chalin:
It originally read:

> Whether to disable generation of the alias redirect to the default language when `DefaultContentLanguageInSubdir` is `true`.

But AFAIK, no redirect is generated when `DefaultContentLanguageInSubdir` is `true`. Changed it to "... when `DefaultContentLanguageInSubdir` is `false`".

### DIFF
--- a/content/en/configuration/languages.md
+++ b/content/en/configuration/languages.md
@@ -25,7 +25,7 @@ defaultContentLanguageInSubdir
 
 disableDefaultLanguageRedirect
 : {{< new-in 0.140.0 />}}
-: (`bool`) Whether to disable generation of the alias redirect to the default language. Default is `false`.
+: (`bool`) Whether to disable the generation of the alias redirect for the default language. When `defaultContentLanguageInSubdir` is `true`, this setting prevents the root directory from redirecting to the language subdirectory. Conversely, when `defaultContentLanguageInSubdir` is `false`, this setting prevents the language subdirectory from redirecting to the root directory. Default is `false`.
 
 disableLanguages
 : (`[]string]`) A slice of language keys representing the languages to disable during the build process. Although this is functional, consider using the [`disabled`](#disabled) key under each language instead.


jmooring:
> But AFAIK, no redirect is generated when DefaultContentLanguageInSubdir is true. 

I just tested again with v0.154.4, and a redirect _is_ created and the `disableDefaultLanguageRedirect` works as expected. Can you provide a failing test case?

jmooring:
> I've noticed that when `defaultContentLanguageInSubdir` is `false`, `disableDefaultLanguageRedirect` seems to be ignored. Is that the desired behavior?

If you're not creating a directory for the default language, a redirect to a non-existent directory is nonsensical.

jmooring:
@chalin Regarding v0.152.2 and earlier, I think I understand your comments, and the need to update the docs (but differently than what you suggest).

Consider this content structure where the `defaultContentLanguage` is `en`...

```text
content/
├── _index.de.md
├── _index.en.md
├── p1.de.md
└── p1.en.md
```

### Case 1

With `defaultContentLanguageInSubdir` = `true`, we create a redirect from the root directory to the directory containing the default content language:

```text
public/
├── de/
│   ├── p1/
│   │   └── index.html
│   └── index.html
├── en/
│   ├── p1/
│   │   └── index.html
│   └── index.html
└── index.html  <-- redirect to /en/
```

If you set `disableDefaultLanguageRedirect` to `true`, a redirect from the root directory to the directory containing the default content language is not created:

```text
public/
├── de/
│   ├── p1/
│   │   └── index.html
│   └── index.html
└── en/
    ├── p1/
    │   └── index.html
    └── index.html
```

### Case 2

With `defaultContentLanguageInSubdir` = `false`, we create a redirect from the directory corresponding to the default content language to the root directory:

```text
public/
├── de/
│   ├── p1/
│   │   └── index.html
│   └── index.html
├── en/
│   └── index.html   <-- redirect to /
├── p1/
│   └── index.html
└── index.html
```

If you set `disableDefaultLanguageRedirect` to `true`, a redirect from the directory corresponding to the default content language to the root directory is not created:

```text
public/
├── de/
│   ├── p1/
│   │   └── index.html
│   └── index.html
├── p1/
│   └── index.html
└── index.html
```

jmooring:
Assuming the behavior doesn't change in v0.154.5, the docs should read something like this:

> disableDefaultLanguageRedirect
>
> (`bool`) Whether to disable the generation of the alias redirect for the default language. When `defaultContentLanguageInSubdir` is `true`, this setting prevents the root directory from redirecting to the language subdirectory. Conversely, when `defaultContentLanguageInSubdir` is `false`, this setting prevents the language subdirectory from redirecting to the root directory. Default is `false`.

chalin:
> Assuming the behavior doesn't change in v0.154.5, the docs should read something like this:

Done (in `configuration/languages.md`). Should `all.md` be updated too or is that auto-updated?

If you'd rather do the edits, feel free to close this PR. Otherwise, I'll do the necessary adjustments. Thanks.

PR #3342, opened 2026-01-11
Deployment docs: bump dependencies to their latest versions

deining:
This PR bumps dependencies in deployment docs to their latest versions.


jmooring:
@deining Thanks for this. I periodically update the Hugo, Dart Sass, Node, and Go version references throughout the documentation, and that includes more than just the deployment docs. I just updated the other references in https://github.com/gohugoio/hugoDocs/pull/3343... just in case you do this again at some point. :smiley:

PR #3345, opened 2026-01-12
BSD: drop sudo, use #; add FreeBSD ports install

ykla:
BSD does not use sudo; use `#` instead. Also add a ports-based installation method for FreeBSD.

BSD* do not include the sudo package or command in base system by default.

This PR is sponsored by the Chinese FreeBSD Community.


jmooring:
> This is probably the most common scenario.

Perhaps, but it is not the lowest common denominator. And, as I mentioned before, if a system is configured to run only as root, the root user will (or should) know that they don't have run to sudo or doas... they're the root user

This documentation has been in place for years, and this is the first time someone has taken issue  with it. We're leaving this as-is for now.

PR #3358, opened 2026-01-29
Update `ToMath` example to include warning about the `hasMath`

DanilaFe:
Following up on the discussion in https://github.com/gohugoio/hugo/issues/14455, update the documentation to include a warning about the issue I ran into. Specifically, the recommended approach of using `hasMath` + render-hook-time tagging of pages will fall through if other pages can display summaries (which would include the LaTeX but not correctly set `hasMath`).

### DIFF
--- a/content/en/functions/transform/ToMath.md
+++ b/content/en/functions/transform/ToMath.md
@@ -150,6 +150,9 @@ Step 3
 
   In the above, note the use of a [noop](g) statement to force content rendering before we check the value of `hasMath` with the `Store.Get` method.
 
+  > [!note]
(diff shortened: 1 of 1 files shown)


jmooring:
@DanilaFe 

You can also do something like this:

```text
  {{ if or (.Page.Store.Get "hasMath") .IsNode }}
    <link
      rel="stylesheet"
      href="https://cdn.jsdelivr.net/npm/katex@0.16.25/dist/katex.min.css"
      integrity="sha384-WcoG4HRXMzYzfCgiyfrySxx90XSl2rxY5mnVY5TwtWE6KLrArNKn0T/mOgNL0Mmi"
      crossorigin="anonymous"
    >
  {{ end }}
```

That means you'd include the CSS is there's math on the page, or when the page kind is one of `section`, `home`, `taxonomy`, or `term`.

PR #3362, opened 2026-01-31
theme: Fix Alpine broken by Turbo v8.0.21

### LINKED ISSUE #3361: Menu on the sidebar not working
After clicking on one a menu on the sidebar menu (**On this page* section to the right), we cannot click again a second time.

The cursor won't change to click mode.

### Reproducing the issue
- Just visit one page on gohub.io - for example: `https://gohugo.io/configuration/introduction/`
- Click on one menu from the sidebar on the right (**On this page** section)
- Click on one menu again. This time it should fail.

jmooring:
Closes #3361

AI Disclosure...

I used Copilot to track this down, which was painful due to several self-inflicted wounds, and consequently a great learning exercise. My final question for Copilot: "OK, so this whole thing was a one line change, and we spent two days on it. How could I have guided you better?"

> You could have been more explicit earlier about:
> 
> 1. "Test in browser console first, before making ANY file changes." - I kept wanting to guess-and-commit. You'd say "test with JS first" but I should have heard that as a hard rule, not a suggestion.
> 1. "Check git/npm history for recent changes." - You mentioned "I'm convinced it's recent" and "it worked after the theme launched," but I didn't dig into actual version history until much later. A direct "look at dependency changes in the last month" would have pointed straight to Turbo.
> 1. "We're looking at Turbo event lifecycle, not Alpine initialization." - Once we understood the issue was same-page anchor clicks, that should have immediately suggested "what Turbo events fire on anchors?" Instead we kept going down rabbit holes about Alpine setup.
> 1. "The fix should be tiny. If you're touching multiple files, you're wrong." - We added pointer-events defensively. A reminder that the actual problem was probably one-line would have kept us focused.
> 1. "Don't commit unless you've verified it works AND can explain why." - This would have prevented 30 exploratory commits.
> 
> Honestly though, the real issue was my approach: I kept favoring action over understanding. You were trying to get me to understand the problem first, then fix it. I kept trying to fix it without understanding. A more patient, debugging-first mindset would have solved this in hours, not days.

### DIFF
--- a/assets/js/helpers/bridgeTurboAndAlpine.js
+++ b/assets/js/helpers/bridgeTurboAndAlpine.js
@@ -1,4 +1,10 @@
 export function bridgeTurboAndAlpine(Alpine) {
+	let isNavigating = false;
+
+	document.addEventListener('turbo:before-visit', () => {
+		isNavigating = true;
+	});
+
 	document.addEventListener('turbo:before-render', (event) => {
 		event.detail.newBody.querySelectorAll('[data-alpine-generated]').forEach((el) => {
 			if (el.hasAttribute('data-alpine-generated')) {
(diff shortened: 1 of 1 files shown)


jmooring:
> You should ask Copilot to also check for any JS errors on the console as a result of this "fix". 

Thanks. I should have examined the console during testing. This PR now "fixes" the original problem (https://github.com/gohugoio/hugoDocs/issues/3361) and there are no console errors.

PR #3429, opened 2026-03-10
content: Document css.Build function

jmooring:
> [!warning]
> Please do not squash this commit.

<https://deploy-preview-3429--gohugoio.netlify.app/functions/css/build/>

TODO:

- [x] ~~See [#14623](https://github.com/gohugoio/hugo/issues/14623).~~
- [x] Change the `target` options to a comma-separated list if needed.
    **Not needed. It's a slice.**
- [x] Determine if we should remove the `assets/jsconfig.json` admonition. I'm not sure if we should be generating that file for `css.Build`.
  **Yes, remove this. We no longer generate this file.**
- [x] Is this correct? "For extensions other than `.css` the `file` loader is used by default." See [this comment](https://github.com/gohugoio/hugo/pull/14610#issuecomment-4032894019).
  **Yes, that is correct.**
- [x] Is this correct? "If unset, no transformations or prefixing are performed." See [this comment](https://github.com/gohugoio/hugo/pull/14610#issuecomment-4032947408).
    **Yes, that is correct.**
- [x] ~~Finish~~ Remove the "Import resolution" section.
- [x] Create a "how do import X" section (keep it simple). Consider placing this section ("how do I do") before the "path resolution" section ("how does it do").
- [x] Finalize the `mainFields` description.
- [x] Remove `both` from the allowable values for the `sourcemap` option
- [x] Note location requirement for referenced resources

For examples, base `target`  on the [browserlist][] "baseline widely available" profile.

```text
{{ $target := slice "chrome115" "edge115" "firefox116" "ios16.4" "opera101" "safari16.4" }}
```

Use this example to test vendor prefixing:

```css
.example-card {
  display: grid;
  user-select: none;
  background: linear-gradient(to bottom, white, black);
  appearance: none;
}
```

[browserlist]: https://browsersl.ist

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -7,7 +7,7 @@
 skip = *.ai,chroma.css,chroma_dark.css,.cspell.json,./data/docs.yaml
 
 # Comma separated list of words to be ignored. Words must be lowercased.
-ignore-words-list = abl,edn,januar,te,trys,ue,womens
+ignore-words-list = abl,edn,ist,januar,te,trys,ue,womens
 
 # Check file names as well.
 check-filenames = true
--- a/.cspell.json
+++ b/.cspell.json
@@ -26,24 +26,24 @@
     "**/tools/*"
   ],
   "ignoreRegExpList": [
-    "# cspell: ignore fenced code blocks",
+    // cspell: ignore fenced code blocks
     "^(\\s*`{3,}).*[\\s\\S]*?^\\1$",
-    "# cspell: ignore words joined with dot",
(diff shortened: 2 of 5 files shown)


jmooring:
OK, this `css.Build` documentation feels solid. I've replaced the "path resolution" section (how it works)  with a  "common patterns" section (how to use it).

While the resolution logic might be useful for troubleshooting, I don't think it belongs on this page. For example, the fact that a package import can resolve to a `node_modules` directory above the project directory is a "how Node works" detail, not a Hugo detail.

PR #3472, opened 2026-04-11
Update actions/upload-pages-artifact version

ulgens:
https://github.com/actions/upload-pages-artifact/releases/tag/v5.0.0

Resolves the remaining

> Node.js 20 actions are deprecated.

warning.

Relates to #3469


jmooring:
Please test things like this before submitting a PR. It's broken. See <https://github.com/actions/upload-pages-artifact/issues/143>.

ulgens:
I see, I tested it on an existing setup and it worked fine but I was using hash pinning (which I believe the example setup needs to do too) Still, I think we could just convert the PR into draft and and merge it when https://github.com/actions/upload-pages-artifact/issues/143 is resolved.

PR #3474, opened 2026-04-15
content: Fix Querify example

sargentalec:
Add second example with htmlEscape function to escape output of querify.


jmooring:
This is wrong.

sargentalec:
Checked again, and this does not happen in production builds but it happens while serving. The Hugo version it was tested is 0.160.0

jmooring:
I suggest you open a topic in the [forum](https://discourse.gohugo.io/).

PR #3539, opened 2026-06-01
theme: Replace Turbo with Speculation Rules

bep:
There's a long living upstream bug that doesn't seem to get resolve anytime soon. Also, Turbo is a little bit of an overkill for our use of it.

I have tested this OK on my (modern) browsers on my MacBook: Chrome, Brave, Safari and Firefox.

This is possibly a UX degradation in older browsers, but in my head much worth it to greatly reduce complexity. If we should ever need any of the more advanced features of Turbo, we should look at https://alpine-ajax.js.org/ -- which should work much more seamless with AlpineJS.


jmooring:
I wasn't aware of this browser API, which isn't surprising. While trying to understand its nuances, I ran across this which may be of interest:

```text
{
  "prefetch": [{
    "source": "document",
    "where": { "href_matches": "/*" },
    "eagerness": "moderate" 
  }],
  "prerender": [{
    "source": "document",
    "where": { "href_matches": "/*" },
    "eagerness": "conservative" 
  }]
}
```

I'll test some more, but so far I like it. A lot.

jmooring (approved):

bep:
>I wasn't aware of this browser API, which isn't surprising. 

I wasn't aware of it until today, when I was looking for a simpler alternative to Turbo and read this:

https://alpine-ajax.js.org/reference/#navigation

bep:
>I ran across this which may be of interest:

I'll merge this now, we can adjust the details later.

PR #3556, opened 2026-06-09
content: Improve segmentation documentation

jmooring:
Closes #3421

<https://deploy-preview-3556--gohugoio.netlify.app/configuration/segments/>


bep:
A detail that may be worth mentioning: Having some `excludes` filter for either `sites` and/or `output` allow us to skip entire "blocks" (we don't have to check every page), which could help with performance in bigger setups. So, if you're rendering JSON only `excludes = ! json` is currently faster than `includes = json`...

jmooring:
> A detail that may be worth mentioning:

Done. See <https://deploy-preview-3556--gohugoio.netlify.app/configuration/segments/#performance-optimization>.

bep:
>output = '!json'

You need a space after `!`.

jmooring:
> You need a space after

Thanks; didn't see that. I'm going to merge this now.

PR #3573, opened 2026-06-18
Update Homebrew's Sass installation instruction to `brew install dart-sass`

Leedehai:
## Problem

`brew install sass/sass/sass` requires an additional step of `brew tap dart-lang/dart`, creating unnecessary frictions for users:

Screenshot 2026-06-17 at 22 00 19

## Cause

This is because the official repo of Sass requires building Dart Sass from source: see the Formula file
https://github.com/sass/homebrew-sass/blob/1db6e8372c08136fff895e0fd34977e420801411/Formula/sass.rb#L23
which calls Dart to compile at installation time:
https://github.com/sass/homebrew-sass/blob/1db6e8372c08136fff895e0fd34977e420801411/Formula/sass.rb#L64

## Solution

Advise users to install the prebuilt Dart Sass binary with `brew install dart-sass` ([brew.sh info page](https://formulae.brew.sh/formula/dart-sass#default)).

### DIFF
--- a/content/en/functions/css/Sass.md
+++ b/content/en/functions/css/Sass.md
@@ -292,9 +292,9 @@ When you install Dart Sass somewhere in your PATH, Hugo will find it.
 
 OS      | Package manager | Site               | Installation
 :-------|:----------------|:-------------------|:-----------------------------
-Linux   | Homebrew        | [brew.sh][]        | `brew install sass/sass/sass`
+Linux   | Homebrew        | [brew.sh][]        | `brew install dart-sass`
 Linux   | Snap            | [snapcraft.io][]   | `sudo snap install dart-sass`
-macOS   | Homebrew        | [brew.sh][]        | `brew install sass/sass/sass`
+macOS   | Homebrew        | [brew.sh][]        | `brew install dart-sass`
 Windows | Chocolatey      | [chocolatey.org][] | `choco install sass`
 Windows | Scoop           | [scoop.sh][]       | `scoop install sass`
 


jmooring:
See:

- <https://sass-lang.com/install/#install-brew>
- <https://github.com/sass/dart-sass#from-homebrew-macos-or-linux>

If they change their documentation we'll change ours.

Leedehai:
Is it possible the Hugo team - instead of me - could reach out to Sass to inform that getting Sass requires installing Dart and compiling from source? If they could make `brew install sass/sass/sass` install a prebuilt binary, or provide an official way from their third-party tap to download the prebuilt binary, that's way better.

I think the current command breaks Hugo philosophy of being a single binary. I mean, since the [deprecation](https://github.com/gohugoio/hugo/issues/14261) of the baked-in `libsass`, requiring a separate peer dependency of Dart Sass is already a compromise, but requiring Dart the compiler to make it work is one step too far TBH.

jmooring:
If this is something you feel strongly about, please create an issue in [their queue](https://github.com/sass/dart-sass/issues).

Leedehai:
I respectfully disagree.

If the Sass team does decide to offer a way to download a prebuilt Sass binary (instead of having users building from source with Dart at installation time), that would be a courtesy they made for the benefits of the Hugo users. Therefore, it's better to be requested by the Hugo team, as it carries weight.

I'm personally okay with the status quo, and don't intend to move forward.

PR #3602, opened 2026-07-14
content: Add front matter example to the taxonomies page

### LINKED ISSUE #3251: expand on taxonomies
Please expand on taxonomies and if they can use the full depth of the data format specification (YAML, JSON, TOML) or if they're limited to a simple key:value / term: value format. Thinking here of how a taxonomie built using YAML should be abl to contain multi-values like collections.

https://yaml.org/spec/1.2.2/#chapter-2-language-overview

The example on Actors is also not consistent due to lack of a example data for YAML/TOML/JSON

https://gohugo.io/content-management/taxonomies/#example-taxonomy-movie-website

chinesepowered:
Adds a front matter example to the movie taxonomy section. The section illustrates the relationships between taxonomies, terms, and values with plain-text diagrams, but never shows the corresponding front matter, so it was not obvious how terms are actually assigned or what they look like in each configuration format.

The new example reuses the data already shown in the diagrams (the actors and director of *Unbreakable*) and renders in TOML, YAML, and JSON via the existing `code-toggle` shortcode. A short follow-up sentence answers the question raised in the issue by noting that a taxonomy is a flat list of string terms rather than a nested data structure, and links to the existing Metadata section for associating richer data with a term.

Closes #3251.

### DIFF
--- a/content/en/content-management/taxonomies.md
+++ b/content/en/content-management/taxonomies.md
@@ -69,6 +69,18 @@ Moonrise Kingdom            <- Value
     ...
 ```
 
+### Assign terms in front matter
+
+Continuing with the example above, assign the terms for each movie in its front matter. Use the plural name of each taxonomy as the field name, and assign the terms as an array, even when a movie has only one term for a given taxonomy:
+
+{{< code-toggle file=content/movies/unbreakable.md fm=true >}}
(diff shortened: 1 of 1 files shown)


jmooring:
Thanks.

PR #3603, opened 2026-07-15
Fix link to 'try' statement in GetRemote.md

willfaught:
I guessed what the URL should be.

### DIFF
--- a/content/en/functions/resources/GetRemote.md
+++ b/content/en/functions/resources/GetRemote.md
@@ -13,7 +13,7 @@ params:
 {{< new-in 0.141.0 >}}
 The `Err` method on the returned resource was removed in v0.141.0.
 
-Use the [`try`][] statement instead, as shown in the [error handling](#error-handling) example below.
+Use the [`try`](/functions/go-template/try/) statement instead, as shown in the [error handling](#error-handling) example below.
 {{< /new-in >}}
 
 ```go-html-template


content/en/functions/resources/GetRemote.md
jmooring, line 237:
```suggestion
```

content/en/functions/resources/GetRemote.md
```diff
@@ -13,7 +13,7 @@ params:
 {{< new-in 0.141.0 >}}
 The `Err` method on the returned resource was removed in v0.141.0.
 
-Use the [`try`][] statement instead, as shown in the [error handling](#error-handling) example below.
+Use the [`try`](functions/go-template/try/) statement instead, as shown in the [error handling](#error-handling) example below.
```
jmooring:
```suggestion
Use the [`try`](/functions/go-template/try/) statement instead, as shown in the [error handling](#error-handling) example below.
```

jmooring (changes requested):

content/en/functions/resources/GetRemote.md
willfaught: This link is used on line 155. I think the issue with the use that was fixed by this PR is that the Markdown is in a shortcode that gets rendered outside the document, so the link isn't available there. It should remain here for the other use.

willfaught:
Force pushed a commit that adds the absolute URL change, but leaves the link definition on line 237 (see https://github.com/gohugoio/hugoDocs/pull/3603#discussion_r3591620050).

jmooring:
Thanks.

PR #3604, opened 2026-07-21
content: Update GetRemote.md with resource save location

willfaught:


content/en/functions/resources/GetRemote.md
```diff
@@ -29,6 +29,8 @@ Use the [`try`](/functions/go-template/try/) statement instead, as shown in the
 {{ end }}
 ```
 
+Resources are saved to the site root.
```
bep: This line doesn't make sense to me.

jmooring: Perhaps this instead:

```md
When you publish a remote resource with the [`Permalink`][], [`RelPermalink`][], or [`Publish`][] method as shown above, Hugo places the resulting file in the root of the [`publishDir`][] rather than mirroring the remote URL path structure. To ensure unique cache keys, Hugo appends a hash to the original base name.

[`Permalink`]: /methods/resource/permalink/
[`RelPermalink`]: /methods/resource/relpermalink/
[`Publish`]: /methods/resource/publish/
[`publishDir`]: /configuration/all/#publishdir
```

willfaught: Fixed! Thanks.

>To ensure unique cache keys, Hugo appends a hash to the original base name.

But isn't that to dedup files with matching HTTP requests? My impression was that the key was for Hugo to decide whether to run the request again, but it's not getting that key from the file's basename.

jmooring: I don't understand your question.

willfaught: https://gohugo.io/functions/resources/getremote/#caching:

>By default, Hugo derives the cache key from the arguments passed to the function. Override the cache key by setting a key in the options map. Use this approach to have more control over how often Hugo fetches a remote resource.

This seems to mean that the cache key isn't taken from the filename to determine whether to repeat the request or reuse the data in the cache, so putting a hash digest in the filename won't stop duplicate requests. Is that right?

bep: >putting a hash digest in the filename won't stop duplicate requests

The hash digest is, by default, built from the URL + options passed in (e.g. headers). That will effectively stop duplicate requests to the same URL using the same options.

PR #3610, opened 2026-07-28
content: Document the css.ChromaStyles function

jmooring:
With the exception of the function name, we use "syntax highlighter" instead of "Chroma" to be consistent with the rest of the documentation. "Syntax highlighter" is more descriptive for a general audience.

See <https://deploy-preview-3610--gohugoio.netlify.app/functions/css/chromastyles/>.


jmooring:
The  examples are verbose (multi-line opts definitions, multi-line concat operation in the third example) to make them as understandable as possible without code comments and to make sure all of the code is visible when viewing the site at max width.

jmooring:
Superseded by #3615.

PR #3619, opened 2026-08-02
theme: Use css.ChromaStyles to generate chroma CSS

bep:
See https://github.com/gohugoio/hugo/pull/15156


bep:
This is mostly a "eat your own dog food" PR -- as I think the current highlighting styles in Hugo Docs looks fine. Note that there may be better GitHub dark/light pairings coming in the next Chroma (e.g. with some fixes in that would make the github styles usable).

jmooring (approved):
This looks great.

Issue #3643, opened 2026-08-30
theme: Look for opportunities to cache partial calls


jmooring:
This is mainly a dogfooding exercise.

```text
hugo build --templateMetrics --templateMetricsHints
```

```text
Template Metrics:

       cumulative       average       maximum      cache  percent  cached  total  
         duration      duration      duration  potential   cached   count  count  template
       ----------      --------      --------  ---------  -------  ------  -----  --------
         16.63  s      22.84 ms     286.92 ms          0        0       0    728  single.html
          3.61  s       4.56 ms      38.31 ms         99        0       0    793  _partials/layouts/header/header.html
          3.21  s       4.05 ms      26.22 ms         10        0       0    793  _partials/layouts/head/head.html
          1.99  s       2.51 ms      33.02 ms        100        0       0    793  _partials/layouts/head/head-js.html
          1.63  s       2.06 ms      22.26 ms         91        0       0    793  _partials/layouts/header/qr.html
          1.61  s     243.27 µs      15.97 ms          0        0       0   6632  _markup/render-link.html
          1.02  s     384.21 µs      21.21 ms          0        0       0   2655  _markup/render-codeblock.html
        914.39 ms      14.51 ms      42.97 ms          0        0       0     63  list.html
        779.27 ms     982.69 µs      18.65 ms          4        0       0    793  _partials/opengraph/opengraph.html
        683.94 ms     862.47 µs      23.20 ms        100        0       0    793  _partials/layouts/footer.html
        508.27 ms       1.42 ms      25.25 ms          0        0       0    359  _shortcodes/code-toggle.html
        475.19 ms     299.62 µs       9.66 ms         58        0       0   1586  _partials/_inline/qr
        413.47 ms     521.39 µs      25.09 ms        100        0       0    793  _partials/layouts/search/input.html
        348.91 ms     439.99 µs      21.14 ms          4        0       0    793  _partials/schema.html
        310.65 ms     392.73 µs      14.31 ms         38        0       0    791  _partials/layouts/docsheader.html
        303.74 ms     382.54 µs      13.08 ms         99        0       0    794  _partials/layouts/home/sponsors.html
        287.33 ms     287.33 ms     287.33 ms          0        0       0      1  __hdeferred/165d8abe79ffade9
        286.62 ms     286.62 ms     286.62 ms        100        0       0      1  _partials/helpers/linkcss.html
[...]

PR #3648, opened 2026-09-04
theme: Remove hyphen variant arg from partialCached calls

jmooring:
See:

- https://github.com/gohugoio/hugoDocs/issues/3643#issuecomment-5538730235
- https://github.com/gohugoio/hugoDocs/issues/3643#issuecomment-5539881706


jjrauch32-lab (commented):
K

